PR #5, opened 2020-04-06
Broadcast messages via telegram

roschaefer:
Broadcast questions via Telegram

* Setup rspec, create issue and feedback model
* Install and setup rubocop
* Implement feedback via Telegram

* feat: refactor docker, broadcast via Telegram
  This is a refactoring of our docker setup which allows us to debug
  ruby code interactively with `binding.pry`:

  ```bash
  docker-compose up
  ```

  In another terminal run

  ```bash
  docker ps # check the container id
  docker attach <CONTAINER_ID>
  ```

  Second, this implements broadcasting via Telegram. Once you've send an
  initial message to our bot, you will be registered as a user. If you
  submit the form in the news frontend, the message gets broadcasted to
  all users.


app/components/textarea/textarea.rb
```diff
@@ -16,7 +18,7 @@ def call
         class: 'c-textarea',
         data: {
           controller: 'textarea',
-          action: 'input->textarea#resize',
+          action: 'input->textarea#resize'
```
tillprochaska, line 21: I like to have a trailing comma in multiline hash or array literals, because it makes it way more convenient to swap lines if you have them. (Without them, I always forget to add and remove commas… 🙃). Plus, you don’t get extra diff lines when adding lines. But I’ll get used to it… :)

app/controllers/telegram/webhook_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class Telegram::WebhookController < Telegram::Bot::UpdatesController
+  def message(message)
+    chat_id = message['chat']['id']
+    user = User.find_by(chat_id: chat_id)
+    unless user
```
tillprochaska, line 7: > Once you've send an initial message to our bot, you will be registered as a user.

I’m not sure that’s the behavior we want, the audience for the tool should be closed, i.e. users shouldn’t be able to signup theirselves. If I understood correctly, it would be more like this:

1. An editor or administrator imports a list of users. Eventually, I guess there should be a UI for it, but for now (and for this PR), I think it would be enough if we manually set them up in the DB.
2. If users would like to communicate via Telegram, they need to either
  a) use the [login with Telegram](https://telegram.org/blog/login) button or
  b) send an initial message to the bot (as it is implemented right now).
  (As 100eyes should be transparent to the end user – at least for now – I think I’d prefer b).

cc @FrauCsu

app/models/feedback.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class Feedback < ApplicationRecord
```
tillprochaska, line 3: As I commented in #3: 

> That was just the first thing that came to my mind, but I like issues. For replies, I think I’d prefer something more generic like reply. I directly associated feedback with a certain type of reply, but maybe that’s just me messing with the meaning of English words… ;)

What do you think?

app/controllers/telegram/webhook_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+      user = User.new(chat_id: chat_id)
+      user.save!
+    end
+    user.respond_feedback(answer: message['text'])
```
tillprochaska, line 11: To be honest, I think the `respond_feedback` is a bit opaque (at least, I had to look it up to understand what exactly it does.) I’d prefer something like this:

```ruby
Issue.last.add_reply(user: user, reply: message['text']
```

db/migrate/20200406192204_create_users.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class CreateUsers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :users do |t|
+      t.string :email
+      t.integer :chat_id
```
tillprochaska, line 7: Maybe use `telegram_username`  or `telegram_id` instead of `chat_id`? It’s more explicit and future-proof in case we add support for other messaging services?

public/icons.svg
```diff
@@ -0,0 +1 @@
...5-11-11-11S11,5.935,11,12v2C11,20.012,15.805,27,22,27z"/><path data-color="color-2" d="M46.707,45.293l-5.109-5.108C42.474,39.016,43,37.57,43,36c0-3.86-3.141-7-7-7s-7,3.14-7,7 s3.141,7,7,7c1.57,0,3.015-0.526,4.184-1.401l5.109,5.109C45.488,46.902,45.744,47,46,47s0.512-0.098,0.707-0.293 C47.098,46.316,47.098,45.684,46.707,45.293z M31,36c0-2.757,2.243-5,5-5s5,2.243,5,5s-2.243,5-5,5S31,38.757,31,36z"/><path fill="currentColor" d="M27,36c0-2.097,0.727-4.024,1.934-5.556C26.824,30.174,24.5,30,22,30c-6.553,0-11.925,1.188-15.278,2.185 C3.3,33.203,1,36.286,1,39.856V44c0,0.552,0.447,1,1,1h34C31.037,45,27,40.962,27,36z"/></g></symbol><symbol id="nc-icon-face-man-glyph-48" viewBox="0 0 48 48"><title>face man</title><g transform="translate(0, 0)"><path fill="currentColor" d="M24,1C11.318,1,1,11.318,1,24s10.318,23,23,23s23-10.318,23-23S36.682,1,24,1z M24,45C12.42,45,3,35.58,3,24 c0-0.595,0.041-1.181,0.089-1.764c2.658-3.33,6.237-3.336,10.744-3.334c5.119,0.012,11.45,0.007,17.959-5.399 c4.957,7.216,10.906,8.954,13.15,9.364C44.963,23.245,45,23.618,45,24C45,35.58,35.58,45,24,45z"/><circle data-color="color-2" cx="15" cy="29" r="3"/><circle data-color="color-2" cx="33" cy="29" r="3"/></g></symbol></svg>
```
tillprochaska, line 1: I’m not sure about this, but it might be possible to import the SVG file in one of the components (or rather in a separate *icon*  component. That way, we’d get hashed file names for this too. That would require that webpacker has an SVG loader setup by default, too, or we set that up ourselves. (But we don’t need to sort that out right now.)

tillprochaska:
Thanks for refactoring the Docker stuff and all the other things I missed… :D

app/controllers/telegram/webhook_controller.rb
tillprochaska: You can get information about the user like this:

```ruby
message.user.username
message.user.first_name
message.user.last_name
```

So in case we want to manually add users using their Telegram username, we could initially use the username to match an incoming message to a user in a database, then save `message.chat_id` and use it for all subsequent incoming messages (to account for a changed username).

tillprochaska:
Hi @roschaefer: I’ve fixed the problem with assets not being compiled correctly. As the dev server was running in a separate container, rails/webpacker wasn't aware. Setting `WEBPACKER_DEV_SERVER_HOST=webpack` was the solution! :)

app/models/feedback.rb
roschaefer: Issues + replies? I'm OK with that

roschaefer: Requests + Replies now

roschaefer:
@tillprochaska the `telegram_chat_id` is required to send a message to a chat. The `telegram_id` is a unique id for the user. You cannot send a message to a `telegram_id` only.

PR #19, opened 2020-04-09
Setup GitHub actions

tillprochaska:
Merge #3 first.

GitHub Actions seem to be a pretty good fit for our requirements, although there are some caveats.

GitHub currently [does not provide Ruby 2.7.1](https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md) on hosted action runners. I set the ruby version to 2.7.0 in the docker setup and Gemfile to prevent any version warnings.

With GitHub Actions, you create workflow YAML files, and you can only specify when to run them (PR, push to master, …) on a per-workflow basis. As a result, there are two workflows, `tests.yml` and `deploy.yml`. That alone isn’t a problem, but the fact that you can’t share steps between workflows (i.e. building the app) is a bit unfortunate. This results in some duplicated code in both workflow files.

 Still, I think we should use it – it works good enough for now and it’s nice not to have to maintain a separate service.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,4 +44,7 @@ jobs:
         env:
           RAILS_ENV: test
           DATABASE_URL: postgres://postgres:postgres@localhost:5432
-        run: bundle exec rails test
+        run: |
+          bundle exec rake db:create
+          bundle exec rake db:migrate
+          bundle exec rake test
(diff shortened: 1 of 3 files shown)


tillprochaska:
@roschaefer I left this PR in draft state intentionally… 😉 The deployment config is actually still missing, but I’ll add that in a separate PR.

PR #23, opened 2020-04-22
Add deployment

tillprochaska:
* Uses `rsync` (via `scripts/deploy`) to deploy changes incrementally and restart services on push to master. Only files/directories specified in `scripts/deploy-include` will be deployed. I prefer whitelisting files for deployment over deploying everything by default, but it might come unexpected if you’re not aware of this limitation
* Caches dependencies (gems and npm packages) between CI runs.
* Adds a bash script (`scripts/uberspace-setup-postgres`) to set up Postgres (download, compile, create root user)
* Adds a bash script (`scripts/uberspace-setup-app`) to set up application related stuff (setup db user, setup rails daemon, …)
* Unfortunately, there’s quite some duplication in the workflow file (`.github/workflows/ci.yml`) as there’s [no easy way](https://github.community/t5/GitHub-Actions/Support-for-YAML-anchors/td-p/30336) to reuse build steps with GitHub actions.
* Currently, neither inbound nor outbound email is setup.

# TODO
- [x] Fix permissions of `scripts/uberspace-setup-*` (add x)


config/database.yml
```diff
@@ -84,4 +84,4 @@ production:
   <<: *default
   database: app_production
   username: app
-  password: <%= ENV['APP_DATABASE_PASSWORD'] %>
+  password: <%= Rails.application.credentials.dig(:production, :postgres_password) %>
```
roschaefer, line 87: Good to know!

scripts/deploy
```diff
@@ -0,0 +1,65 @@
...
+# Write master.key
+echo $RAILS_MASTER_KEY > config/master.key
+
+# Add user's private key
+mkdir $HOME/.ssh
+echo $DEPLOY_USER_PRIVATE_KEY | base64 --decode > $HOME/.ssh/deploy_rsa
+chmod 400 $HOME/.ssh/deploy_rsa
+
+# Add target's key
+echo $DEPLOY_TARGET_PUBLIC_KEY | base64 --decode >> $HOME/.ssh/known_hosts
```
roschaefer, line 32: Why is that necessary? Anyways, I'm sure it does not hurt.

scripts/deploy
```diff
@@ -0,0 +1,65 @@
...
+  --checksum \
+  --delete \
+  --recursive \
+  --links \
+  --compress \
+  --human-readable \
+  --info=name1,stats2 \
+  --rsh="ssh -i $HOME/.ssh/deploy_rsa" \
+  --exclude-from=$deploy_exclude \
+  --include-from=$deploy_include \
```
roschaefer, line 56: Fancy

scripts/uberspace-setup-app
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
roschaefer, line 1: This script is only run initially, right?

scripts/uberspace-setup-postgres
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
```
roschaefer, line 1: Have you written it? Or is this copy+pasted from some uberspace docs?

roschaefer:
What's the URL of our deployed app? @tillprochaska

scripts/deploy
tillprochaska: You could disable strict host key checking (otherwise you’d be prompted to confirm the connection). But I can’t imagine a reason to disable this security feature.

scripts/uberspace-setup-app
tillprochaska: Yes, both `uberspace-setup-postgres` and `uberspace-setup-app` are meant to be run manually, a single time on a fresh Uberspace account. They are responsible for compiling Postgres, setting up db users and daemons.

scripts/uberspace-setup-postgres
tillprochaska: It’s basically this [Uberspace guide](https://lab.uberspace.de/guide_postgresql.html) on how to run Postgres, but in the form of a bash script as I was annoyed by having to set up everything over and over when creating new Uberspace accounts. I’m pretty sure there are a few cases it does not handle – but as long as you run it on a fresh Uberspace account, it should work just fine.

tillprochaska:
> What's the URL of our deployed app? @tillprochaska

https://hundred.uber.space

PR #25, opened 2020-04-27
Show conversations for each user

roschaefer:
~please merge #22 first~ done


app/components/notice/notice.css
```diff
@@ -0,0 +1,4 @@
+#notice {
```
tillprochaska, line 1:
```suggestion
.Notice {
```

app/components/notice/notice.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if notice %>
+  <p id="notice"><%= notice %></p>
```
tillprochaska, line 2: I usually don’t use IDs for styling purposes at all as they usually only introduce specificity issues.

```suggestion
  <div class="Notice"><%= notice %></div>
```

frontend/global/base.css
```diff
@@ -6,7 +6,7 @@
 
 html, body {
     font-size: var(--font-size);
-    line-height: var(--spacing-unit);
+    line-height: 2*var(--spacing-unit);
```
tillprochaska, line 9: What’s the purpose of this?

This isn’t valid CSS. If you actually wanted to multiply by two you’d have to do `calc(2 * var(--spacing-unit))` or, even better, use `var(--spacing-unit-xl)` (which resolves to `calc(2 * var(--spacing-unit))`.

But I don’t think you actually wanted to increase global line height like this?

frontend/global/base.css
```diff
@@ -16,7 +16,7 @@ html, body {
 
 input, button, textarea {
     font-size: var(--font-size);
-    line-height: var(--spacing-unit);
+    line-height: 2*var(--spacing-unit);
```
tillprochaska, line 19: Same as above.

frontend/global/base.css
roschaefer: To increase the space between the lines in the `<h1>`. It *does* have an effect.

Screenshot - 2020-04-30T150548 424


Screenshot - 2020-04-30T150635 133


Anyways, I'm using `c 'page_header'` now.

tillprochaska: Ah, that’s true and interesting. I think what’s happening is that the value is invalid and the rule does not have an effect and in consequence, it falls back to the line height defined in the default browser style sheet (`line-height: normal`) which is usually specified relative to the elements font size.

PR #27, opened 2020-04-29
feat: Run telegram bot poller in production

roschaefer:
@tillprochaska I fixed this obscure $(ENV_HOME)s/ line in the setup
script. Also, the command `bin/rails` was actually `bin/bundle exec
rails` on our VPS server, so I changed that in the code, too.

### DIFF
--- a/scripts/uberspace-setup-app
+++ b/scripts/uberspace-setup-app
@@ -15,8 +15,8 @@ printf "\n✅  Done!\n"
 printf "\n\n🔷  Setup service...\n"
 cat <<EOT >> ~/etc/services.d/100eyes.ini
 [program:100eyes]
-directory=$(ENV_HOME)s/100eyes
-command=bin/rails server -b 0.0.0.0
+directory=$HOME/100eyes
+command=bin/bundle exec rails server -b 0.0.0.0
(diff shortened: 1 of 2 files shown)


scripts/uberspace-setup-app
```diff
@@ -15,8 +15,8 @@ printf "\n✅  Done!\n"
 printf "\n\n🔷  Setup service...\n"
 cat <<EOT >> ~/etc/services.d/100eyes.ini
 [program:100eyes]
-directory=$(ENV_HOME)s/100eyes
-command=bin/rails server -b 0.0.0.0
+directory=$HOME/100eyes
```
tillprochaska, line 18: `$(ENV_VAR)/s` actually is the syntax to use with supervisord.

http://supervisord.org/configuration.html#environment-variables

scripts/uberspace-setup-telegram-bot-poller
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+db=app_production
+user=app
```
tillprochaska, line 4:
```suggestion
```

scripts/uberspace-setup-app
roschaefer: I thought you would run the script via bash and then at least for me, the `ENV_HOME` would get substituted with empty string.

PR #30, opened 2020-04-30
Don’t select fallback avatars randomly

tillprochaska:
* Selects a fallback image from `public/avatars` if no URL is provided.
* The fallback image is based on the user’s ID (or any other key).


app/components/avatar/avatar.rb
```diff
@@ -1,15 +1,33 @@
...
+      @key = key
     end
 
     private
 
-    attr_reader :url
+    attr_reader :key
+
+    def url
+      return fallback_url unless @url
```
roschaefer, line 24:
```suggestion
      @url || fallback_url
```

app/components/avatar/avatar.rb
```diff
@@ -1,15 +1,33 @@
...
-    include ComponentHelper
+  class Avatar < ApplicationComponent
+    FALLBACK_BASE_URL = '/avatars'
+    FALLBACK_IMAGES = [
+      'fallback-cat.jpg',
+      'fallback-dog.jpg',
+      'fallback-otter.jpg',
+      'fallback-seal.jpg',
+      'fallback-squirrel.jpg',
+    ]
```
roschaefer, line 13:
```suggestion
    FALLBACK_IMAGES = {
      cat: 'fallback-cat.jpg',
      dog: 'fallback-dog.jpg',
      otter: 'fallback-otter.jpg',
      seal: 'fallback-seal.jpg',
      squirrel: 'fallback-squirrel.jpg',
    }


```

app/components/avatar/avatar.rb
```diff
@@ -1,15 +1,33 @@
...
 
-    def initialize(url:)
+    def initialize(url: nil, key: 0)
```
roschaefer, line 14:
```suggestion
    def initialize(url: nil, animal: nil)
```

app/components/avatar/avatar.rb
```diff
@@ -1,15 +1,33 @@
...
-    attr_reader :url
+    attr_reader :key
+
+    def url
+      return fallback_url unless @url
+      @url
+    end
+
+    def fallback_url
+      file = FALLBACK_IMAGES[key % FALLBACK_IMAGES.length]
```
roschaefer, line 29:
```suggestion
      file = FALLBACK_IMAGES[animal] || FALLBACK_IMAGES.values.sample
```

app/components/avatar/avatar.html.erb
```diff
@@ -1,3 +1,3 @@
 <div class="Avatar">
-    <img src="<%= url || 'https://source.unsplash.com/random?cat' %>" />
+    <img src="<%= url %>" alt="" />
```
roschaefer, line 2: Does it help in any way to have an empty `alt`?

app/components/avatar/avatar.rb
drjakob: Otter!

app/components/avatar/avatar.html.erb
tillprochaska: Yes, an image tag without an `alt` attribute is invalid HTML. An empty `alt` attribute indicates that the image is optional. This way, many screen readers will simply skip the element. If you completely omit the attribute, screen readers might announce there’s an image element, but aren’t able to provide any details what is shown – which is confusing and unnecessary.

app/components/avatar/avatar.rb
tillprochaska: Why would you want to specify a concrete fallback image?

PR #31, opened 2020-05-04
feat: Configure telegram bot with BOT=robert

roschaefer:
@tillprochaska for testing, the gem will complain if you start your
containers with `env BOT=something docker-compose up` and run
`docker-compose exec app rails spec`. It will complain about missing
default paths, see below :point_down:.

```
Failures:

  1) Requests POST /requests given a user with a telegram_chat_id is expected to send telegram message "How do you do?"
     Failure/Error: it { should respond_with_message 'How do you do?' }
       expected to make exactly 1 #<Telegram::Bot::ClientStub#13600(1245394754:AAFVzLe5HUubaz1IsEQ8RaQAFUQRFEoGqIQ)>.sendMessage requests, with [#<RSpec::Mocks::ArgumentMatchers::HashIncludingMatcher:0x000055fc2f367cd0 @expected={:chat_id=>4711, :text=>"How do you do?"}>], but made 0, and 0 with
     # ./spec/requests/requests_spec.rb:34:in `block (4 levels) in <main>'

  2) Telegram::WebhookController#start! is expected to send telegram message "Hello Joe!"
     Failure/Error: subject { -> { dispatch_command :start, { from: { username: 'Joe' } } } }

     NoMethodError:
       undefined method `default_telegram_webhook_path' for #<Module:0x000055fc2f40f598>
     # /usr/local/bundle/gems/telegram-bot-0.14.3/lib/telegram/bot/rspec/integration/rails.rb:8:in `public_send'
     # /usr/local/bundle/gems/telegram-bot-0.14.3/lib/telegram/bot/rspec/integration/rails.rb:8:in `block (2 levels) in <main>'
     # /usr/local/bundle/gems/telegram-bot-0.14.3/lib/telegram/bot/rspec/integration/rails.rb:19:in `dispatch'
     # /usr/local/bundle/gems/telegram-bot-0.14.3/lib/telegram/bot/rspec/message_helpers.rb:11:in `dispatch_message'
     # /usr/local/bundle/gems/telegram-bot-0.14.3/lib/telegram/bot/rspec/message_helpers.rb:18:in `dispatch_command'
     # ./spec/requests/telegram/webhook_spec.rb:8:in `block (4 levels) in <main>'
     # /usr/local/bundle/gems/telegram-bot-0.14.3/lib/telegram/bot/rspec/client_matchers.rb:74:in `matches?'
     # ./spec/requests/telegram/webhook_spec.rb:9:in `block (3 levels) in <main>'
```

PR #32, opened 2020-05-07
Full text search

roschaefer:
Change the target branch of this PR to `master` as soon as #31 is merged.


app/components/search_result_component/search_result_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<li>
+  <% case result.class.to_s %>
+  <% when 'User' %>
+    <%= c 'button', label: result.name , styles: [:link], link: user_path(result) %>
+  <% when 'Reply' %>
+    <%= c 'button', label: result.text , styles: [:link], link: user_request_path(result.user, result.request) %>
+  <% end %>
```
tillprochaska, line 7: I think I’d prefer to have a lean template and put that logic into the component class:

```erb
<%= c 'button', label: title , styles: [:link], link: link %>
```

… and then methods in `SearchResult`.

app/components/search_result_component/search_result_component.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module SearchResultComponent
+  class SearchResultComponent < ApplicationComponent
```
tillprochaska, line 4: I know the component generator is currently appending `component` to everything – to be consistent with how the other components are named, could you rename the files/class? (I’ll push an updated generator!)

`search_result_component/search_result_component.rb` --> `search_result/search_result.rb`

```suggestion
module SearchResult
  class SearchResult< ApplicationComponent
```

PR #37, opened 2020-05-13
Send emails in production

roschaefer:
For some strange reasons, I will only receive an email to my own verified sender email address.


config/environments/production.rb
```diff
@@ -106,4 +106,16 @@
 
   # CUSTOM
   config.bot_id = (ENV['BOT'] || :default).to_sym
+  config.action_mailer.smtp_settings = {
+    user_name: Rails.application.credentials.sendgrid[:user_name],
+    password: Rails.application.credentials.sendgrid[:password],
+    domain: Rails.application.credentials.sendgrid[:domain],
+    address: 'smtp.sendgrid.net',
+    port: 587,
+    authentication: :plain
```
tillprochaska, line 115:
```suggestion
    authentication: :plain
    tls: true
```

app/controllers/requests_controller.rb
```diff
@@ -9,10 +9,12 @@ def new; end
   def create
     question = params[:question]
     Request.create!(text: question)
-    QuestionMailer
-      .with(question: question)
-      .new_question_email
-      .deliver_later
+    User.where.not(email: nil).find_each do |user|
+      QuestionMailer
```
tillprochaska, line 13: I know this wasn’t introduced in this PR – but I was slightly confused why it’s called `QuestionMailer` as we don’t have questions any more, but requests. Probably something, we should think about when we have clear terms for everything.

roschaefer: :+1:

PR #38, opened 2020-05-14
feat: Edit user email address through web UI

roschaefer:


app/components/user_profile/user_profile.html.erb
```diff
@@ -7,6 +7,12 @@
         id: 'user[name]',
         styles: ['clickToEdit', 'center', 'heading']
     %>
+    <%= c 'input',
+        placeholder: 'E-Mail Adresse',
```
tillprochaska, line 11:
```suggestion
        placeholder: 'E-Mail-Adresse',
```

PR #39, opened 2020-05-14
Message component

roschaefer:
@tillprochaska this is not yet finished - it could be merged though. 

Screenshot - 2020-05-15T161413 339

This PR fixes the bug with different avatars for the same user on different pages.

### DIFF
--- a/app/components/avatar/avatar.css
+++ b/app/components/avatar/avatar.css
@@ -1,22 +1,14 @@
 .Avatar {
-    display: inline-block;
-    box-sizing: content-box;
-    border-radius: 50%;
-    overflow: hidden;
-}
-
-.Avatar-small {
-  width: calc(2 * var(--spacing-unit));
-  height: calc(2 * var(--spacing-unit));
-}
-
-.Avatar-large {
+  display: inline-block;
+  box-sizing: content-box;
+  border-radius: 50%;
+  overflow: hidden;
   width: calc(3 * var(--spacing-unit));
   height: calc(3 * var(--spacing-unit));
 }
 
 .Avatar > img {
-    width: 100%;
-    height: 100%;
-    object-fit: cover;
+  width: 100%;
+  height: 100%;
+  object-fit: cover;
 }
--- a/app/components/avatar/avatar.css
+++ b/app/components/avatar/avatar.css
@@ -1,10 +1,18 @@
 .Avatar {
     display: inline-block;
     box-sizing: content-box;
-    width: calc(3 * var(--spacing-unit));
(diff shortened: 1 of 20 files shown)


app/components/avatar/avatar.css
```diff
@@ -1,12 +1,20 @@
 .Avatar {
     display: inline-block;
     box-sizing: content-box;
-    width: calc(3 * var(--spacing-unit));
```
tillprochaska, line 4: Why don’t we keep one of them as the default variant and only add a single modifier? I think having a default of 2 * spacing-unit, and then a large variant would be fine.

app/components/avatar/avatar.rb
```diff
@@ -11,22 +11,30 @@ class Avatar < ApplicationComponent
...
-      @url || fallback_url
+      user&.avatar_url || fallback_url
     end
 
     def fallback_url
       file = FALLBACK_IMAGES[key % FALLBACK_IMAGES.length]
       "#{FALLBACK_BASE_URL}/#{file}"
     end
+
+    def css_class
```
tillprochaska, line 36: Once #42 is merged, you can simply do this:

```ruby
class Avatar < ApplicationComponent
  def initialize(user: nil, **)
    super

    ...
  end
end
```

And in the template:

```erb
<div class="<% class_names.join(' ') %>">
  ...
</div>
```

Or alternatively, by defining a `call` method on the component:

```ruby
class Avatar < ApplicationComponent
  ...
  
  def call
    content_tag(:div, ..., class: class_names)
  end
end
```

app/components/message/message.css
```diff
@@ -0,0 +1,14 @@
+.Message {
+  display: flex;
+  align-items: center;
+  margin: auto;
+  justify-content: flex-end;
+}
+
+.Message-left {
```
tillprochaska, line 8: Use `.Message--left` for modifiers and `.Message-child` for subelements.

app/components/message/message.rb
```diff
@@ -6,10 +6,17 @@ class Message < ViewComponent::Base
...
     private
 
-    attr_reader :message
+    attr_reader :message, :user
+
+    def avatar_url
+      return user.avatar_url if user
+
+      nil
+    end
```
tillprochaska, line 20: Is that still required?

tillprochaska:
Hi @roschaefer! I changed a few minor things, hope that’s fine! In most cases, that was to use a generic component method instead of a component-specific implementation (e.g. `class_names`) or a generic component instead of custom styles (e.g. using a `Stack` instead of manually adding margin).

PR #40, opened 2020-05-15
Component generator

tillprochaska:
```
Usage:
  rails generate view_component NAME [options]

Options:
  [--skip-namespace], [--no-skip-namespace]  # Skip namespace (affects only isolated applications)
  [--css], [--no-css]                        # Indicates when to generate css
                                             # Default: true
  [--js], [--no-js]                          # Indicates when to generate js
                                             # Default: true
  [--template], [--no-template]              # Indicates when to generate template
                                             # Default: true
  [--spec], [--no-spec]                      # Indicates when to generate spec
                                             # Default: true

Description:
    Creates a new component and spec file.
    Pass the component name either CamelCase or under_scored.

Example:
    rails generate component MyButton

    This will create:
        Component:           app/components/my_button/my_button.rb
        Template:            app/components/my_button/my_button.html.erb
        CSS:                 app/components/my_button/my_button.css
        Stimulus Controller: app/components/my_button/my_button.js
        Spec:                spec/components/my_button_spec.rb
```

### DIFF
--- a/lib/generators/view_component/USAGE
+++ b/lib/generators/view_component/USAGE
@@ -1,0 +1,13 @@
+Description:
+    Creates a new component and spec file.
+    Pass the component name either CamelCase or under_scored.
+
+Example:
+    rails generate component MyButton
+
+    This will create:
+        Component:           app/components/my_button/my_button.rb
+        Template:            app/components/my_button/my_button.html.erb
+        CSS:                 app/components/my_button/my_button.css
+        Stimulus Controller: app/components/my_button/my_button.js
+        Spec:                spec/components/my_button_spec.rb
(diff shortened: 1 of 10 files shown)


lib/generators/view_component/USAGE
```diff
@@ -0,0 +1,13 @@
...
+
+Example:
+    rails generate component MyButton
+
+    This will create:
+        Component:           app/components/my_button/my_button.rb
+        Template:            app/components/my_button/my_button.html.erb
+        CSS:                 app/components/my_button/my_button.css
+        Stimulus Controller: app/components/my_button/my_button.js
+        Spec:                spec/components/my_button_spec.rb
```
roschaefer, line 13: rspec-rails should put the spec files into the same directory :unamused:

PR #41, opened 2020-05-15
Nav bar

tillprochaska:
Merge #40 first.

Adds a basic top bar to every view, allowing editors to switch between the two main functionalities.

This is probably a good place to add the app-wide search to as well.

Bildschirmfoto 2020-05-15 um 11 15 15


drjakob:
Müsste Lokalexpert*innen heißen, nicht Teilnehmerinnen.

PR #42, opened 2020-05-15
Add improved request creation form

tillprochaska:
For now this only adds components to render the improved requests form – it is not yet used for the actual view to create a new request. There are a number of smaller tasks that need to be finished before, but I’d like to merge this already as it adds a number of generic-use components.

Bildschirmfoto 2020-05-15 um 11 19 41

**To do:**

- [ ] Add request title
- [ ] Add notes and deadline to request model
- [ ] Add method to render full request text based on notes and deadline
- [ ] Double-check wording/copy
- [ ] Improve deadline selection


app/components/request_form/request_form.js
```diff
@@ -0,0 +1,58 @@
...
+    medicalInfo: 'Textbaustein für medizinische Informationen',
+    confidential: 'Textbaustein für vertrauliche Informationen',
+};
+
+const template = ({name, message, notes, deadline }) => {
+    name = sanitize(name) || '<span class="Placeholder">Name</span>';
+    message = sanitize(message).replace(/\n/g, "<br>") || '<span class="Placeholder">Nachricht</span>';
+    const noteTexts = [];
+
+    for(const [key, isActive] of Object.entries(notes)) {
```
roschaefer: Use [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce).

app/components/stack/stack.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Stack
+  class Stack < ApplicationComponent
+    def initialize(size: nil)
```
roschaefer, line 5:
```suggestion
    def initialize(space: nil)
```

PR #52, opened 2020-05-17
User component

roschaefer:
Screenshot - 2020-05-17T215019 641

Screenshot - 2020-05-17T215150 615


public/email.svg
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
```
roschaefer, line 1: @tillprochaska please tell me where you got this icon set from and how to add more icons.

tillprochaska:
@roschaefer: Here’s what I changed:

* The component doesn’t show the full email address/user name any more, instead, it shows only icons indicating which communication channels are active for a given user. I moved the logic to determine which channels are active to the User model and to the component class.
* I updated the icon component so you can easily use icons inline.

Screenshot 2020-05-19 20 27 03

PR #53, opened 2020-05-18
Auto-generate request message based on text and hints/text blocks

tillprochaska:
**UX:**
* Adds a field for an internal title.
* Adds checkboxes for a number of hints ("Hinweise/Textbausteine").
* Adds a live-preview of the message that will be sent to contributors, including hints.
* Currently, submitting the form won’t redirect you anywhere – I think we should eventually redirect to the request overview ("Übersicht für eine Recherchefrage") displaying the request and (once they come in) all replies for that request.

**Technical Details:**
* Adds a `hints` column to `requests`. This column contains an array hints ("Textbausteine") the editor selected and that should be included in the message sent to users via email or messenger.
* Adds a `plaintext` method to `Request` that generates a plaintext representation of the request using (currently) hard-coded salutation, greetings, the question text, and all hints ("Textbausteine").
* Adds a `Page` component that provides a card-like page in front of a light-gray background.
* Adds a `PageTitle` that provides a simple headline with a subtle box shadow to separate it from other page contents. (This is pretty similar to the `PageHeader` component we already use and might eventually replace it. However, I didn’t want to break existing views so I decided to add this as a separate component for now.)
* Some static strings/UI copy is duplicated in separated places. I’d like to eventually extract them to a locale file, but didn’t want to address this in this PR as it’s the case in many places across the app.

### DIFF
--- a/spec/models/request_spec.rb
+++ b/spec/models/request_spec.rb
@@ -27,7 +27,7 @@ RSpec.describe Request, type: :model do
   describe '.plaintext' do
     subject { request.plaintext }
 
-    it 'renders correct plaintext message' do
+    it 'returns correct plaintext message' do
       expected  = "Hallo, die Redaktion hat eine neue Frage an dich:\n\n"
       expected += "What is the answer to life, the universe, and everything?\n\n"
       expected += "Textbaustein für Foto\n\n"
@@ -38,7 +38,15 @@ RSpec.describe Request, type: :model do
     end
 
     describe 'without hints' do
-      subject { Request.new(title: 'Example', text: 'Hello World!').plaintext }
+      subject { Request.new(title: 'Example', text: 'Hello World!', hints: []).plaintext }
+
+      it 'returns correct plaintext message' do
+        expected  = "Hallo, die Redaktion hat eine neue Frage an dich:\n\n"
+        expected += "Hello World!\n\n"
+        expected += "Vielen Dank für deine Hilfe bei unserer Recherche!"
+
+        expect(subject).to eql(expected)
+      end
     end
   end
 
--- a/spec/models/request_spec.rb
+++ b/spec/models/request_spec.rb
@@ -19,7 +19,12 @@ RSpec.describe Request, type: :model do
     expect(subject.attributes.keys).to include('title', 'text', 'hints')
   end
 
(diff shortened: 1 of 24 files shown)


spec/models/request_spec.rb
```diff
@@ -4,16 +4,59 @@
...
+
+  describe '.plaintext' do
+    subject { request.plaintext }
+
+    it 'returns correct plaintext message' do
+      expected  = "Hallo, die Redaktion hat eine neue Frage an dich:\n\n"
+      expected += "What is the answer to life, the universe, and everything?\n\n"
+      expected += "Textbaustein für Foto\n\n"
+      expected += "Textbaustein für vertrauliche Informationen\n\n"
+      expected += "Vielen Dank für deine Hilfe bei unserer Recherche!"
```
roschaefer, line 35:
```suggestion
    expected = <<~EXP
      Hallo, die Redaktion hat eine neue Frage an dich:

      What is the answer to life, the universe, and everything?

      Textbaustein für Foto

      Textbaustein für vertrauliche Informationen

      Vielen Dank für deine Hilfe bei unserer Recherche!
    EXP
```

spec/models/request_spec.rb
```diff
@@ -4,16 +4,59 @@
...
+
+      expect(subject).to eql(expected)
+    end
+
+    describe 'without hints' do
+      subject { Request.new(title: 'Example', text: 'Hello World!', hints: []).plaintext }
+
+      it 'returns correct plaintext message' do
+        expected  = "Hallo, die Redaktion hat eine neue Frage an dich:\n\n"
```
roschaefer, line 44: Same :point_up:

spec/requests/requests_spec.rb
```diff
@@ -5,10 +5,18 @@
...
+    it { should change { Request.count }.from(0).to(1) }
+
+    describe 'without hints param' do
+      let(:params) { { title: 'Example Question', text: 'How do you do?' } }
+      it { should_not raise_error }
+    end
+
     describe 'without users' do
       it { should_not raise_error }
-      it { should change { Request.count }.from(0).to(1) }
```
roschaefer, line 11: Why that?

PR #58, opened 2020-05-18
Set input font size relative to parent

tillprochaska:
This sets the font size of input elements depending on the parent element, resulting in a more balanced apeareance of the navigation bar.

Screenshot 2020-05-19 01 23 04


drjakob:
Good idea

PR #63, opened 2020-05-19
Stick request preview to top if form is too long

tillprochaska:

### DIFF
--- a/app/components/request_form/request_form.css
+++ b/app/components/request_form/request_form.css
@@ -8,7 +8,16 @@
 }
 
 .RequestForm-column + .RequestForm-column {
+    display: flex;
+    flex-direction: column;
     margin-left: var(--spacing-unit);
+}
+
+.RequestForm-column > * {
+    /* Ensure direct children take the full column height,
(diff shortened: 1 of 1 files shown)


app/components/request_form/request_form.css
```diff
@@ -8,9 +8,18 @@
 }
 
 .RequestForm-column + .RequestForm-column {
+    display: flex;
+    flex-direction: column;
```
roschaefer, line 12: Hm, why only every item except the first? :thinking:

tillprochaska: Argh, good catch, that’s by mistake.

tillprochaska:
> I think I can see the difference in the gif but I can't understand how you manage to do it with CSS.

Most of it is just [`position:sticky`](https://developer.mozilla.org/en-US/docs/Web/CSS/position#sticky). But in order for it to work, you need to ensure it’s parent container is actually large enough. See this demo: https://developer.mozilla.org/en-US/docs/Web/CSS/position#sticky

PR #77, opened 2020-05-20
Various refactorings

roschaefer:


app/views/search/index.html.erb
```diff
@@ -1,9 +1,15 @@
...
-
-  <% @results.each do |result| %>
-    <%= c 'search_result', result: result %>
+<%= c 'page' do %>
+  <%= c 'page_title' do %>
+    Suchergebnisse
+  <% end %>
+  <%= c 'box' do %>
+    <%= c 'stack', space: :xlarge do %>
+      <%= c 'stack' do %>
```
tillprochaska, line 7: Why do you use nested stacks here?

PR #83, opened 2020-05-21
Fix code formatting

tillprochaska:


app/components/input/input.css
```diff
@@ -41,7 +41,7 @@
 
 .Input--inverted {
   border: none;
-  background: rgba(255, 255, 255, .2);
+  background: rgba(255, 255, 255, 0.2);
```
roschaefer, line 44: Ok, so omitting the leading `0` is considered bad practice?

PR #102, opened 2020-05-21
fix: Workaround broken PAGER in docker image

roschaefer:
thanks: https://github.com/pry/pry/issues/1248#issuecomment-366056015


Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN apk --update add \
     tzdata \
     libxslt-dev \
     libxml2-dev
```
roschaefer, line 14: missing `\` @tillprochaska

PR #108, opened 2020-05-21
Edit user data

### LINKED ISSUE #90: BUTTON edit user profiles
As editor
i want to edit user information in VERLAUF by the simple click on a button
in order to keep regulary notes

tillprochaska:
Closes #90 

**Details:**
* This primarily adds a new `Field` variant `horizontal` that displays the input next to its label.
* This new variant is used in the `UserForm` component (previously `UserProfile`). I renamed that component, as it’s really only a form.
* Additionally, I adjusted some of the button/input styles or added new variants here.


drjakob:
Personally I think the save button is at the wrong place. should be (ss in github) green button botrtom right, no?

tillprochaska:
> Personally I think the save button is at the wrong place. should be (ss in github) green button botrtom right, no?

Agree!

>  Can we give a hint if you hover Telegram-Handle why you can't change it?

Sure!

Screenshot 2020-05-22 15 36 20

drjakob:
Great

PR #109, opened 2020-05-21
Telegram download photos

roschaefer:
Screenshot - 2020-05-22T233254 704


app/models/photo.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class Photo < ApplicationRecord
```
tillprochaska, line 3: Maybe it would be a good idea to extract all the telegram logic into a separate class at some point? Right now, it’s scattered across controllers and models.

```ruby
class TelegramMessage
  def initialize(message)
    @message = message
  end

  def create_reply!
    # create a Reply recors
  end

  def download_photos!
    # download photos from Telegram API and attach them to reply
  end

  private

  def user
    # find user by chat id
  end
end
```

app/models/photo.rb
```diff
@@ -0,0 +1,19 @@
...
+  belongs_to :reply
+  has_many_attached :images
+
+  def telegram_message=(message)
+    (message['photo'] || []).each do |telegram_file|
```
tillprochaska, line 8:
```suggestion
    return unless message.any?
    
    message['photo'].each do |telegram_file|
```

app/models/photo.rb
tillprochaska: This might even provide a common interface for incoming messages of all channels:

```ruby
class EmailMessage
  def initialize(email)
    @email = email
  end

  def create_reply!
    # ...
  end

  def create_photos!
    # ...
  end

  private

  def user
    # find user by email address
  end
end
```

app/models/photo.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class Photo < ApplicationRecord
+  belongs_to :reply
+  has_many_attached :images
```
tillprochaska, line 5: Wait, a photo record has many images? 🤯 Does Telegram provide different resolutions or why’s that?

Otherwise, if that’s just all the photos a user sent with a single message, why not name the record either `ReplyPhotos` or alternatively, attach the images directly to a `Reply`?

roschaefer: Each photo comes with a somehow unpredictable set of resolutions. With my test photos yesterday, it was usually between 1-3. The last of them seems to be the original or highest resolution. Then you can send multiple photos as an album. The telegram API will send each photo with its image files in one message. It took me a while to get through that.

So the connection is Reply <1:m> Photo <1:m> attached_images.

drjakob: There are many use cases for a photo collection e.g. a document with more than one page.

config/environments/test.rb
```diff
@@ -55,7 +55,7 @@
   #
   config.active_job.queue_adapter = :test
 
-  config.bot_id = :default
+  config.bot_id = (ENV['BOT'] || :default).to_sym
```
roschaefer, line 58: For re-recording VCR cassettes

roschaefer:
I'm finished refactoring - finally! I hope it was worth it.

app/components/message/message.css
```diff
@@ -11,3 +11,19 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: space-around;
+  padding-bottom: var(--spacing-unit);
+}
+
+.Message-photos > img {
+  border: 3px solid rgba(0, 0, 0, 0.05);
+  max-width: 100%;
+  height: auto;
```
tillprochaska, line 25: `height: auto` should be the initial value of any `img` element, so I think this might not be necessary.

```suggestion
```

app/components/message/message.css
```diff
@@ -11,3 +11,19 @@
...
+
+.Message-photos {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-around;
+  padding-bottom: var(--spacing-unit);
+}
+
+.Message-photos > img {
+  border: 3px solid rgba(0, 0, 0, 0.05);
```
tillprochaska, line 23: Nice! 🎉

app/components/message/message.css
```diff
@@ -11,3 +11,19 @@
   color: var(--color-text-light);
   font-size: var(--font-size-s);
 }
+
+.Message-photos {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-around;
```
tillprochaska, line 18: Why do you need a flexbox if all you do is displaying several images below each other?

app/components/message/message.css
```diff
@@ -11,3 +11,19 @@
...
+}
+
+.Message-photos > img {
+  border: 3px solid rgba(0, 0, 0, 0.05);
+  max-width: 100%;
+  height: auto;
+}
+.Message-photos > img + img {
+  margin-top: var(--spacing-unit);
+}
```
tillprochaska, line 29: You *could* just wrap the images in a `stack` – no need for an additional rule set then.

spec/decorators/telegram_message_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+            tm = TelegramMessage.new message_with_media_group_id
+            reply = tm.reply
+            reply.request = request
+            reply.save
+            reply.photos << tm.photos
+          }
+        end
+
+        it { expect { 3.times { save_reply_and_photo.call } }.to(change { Reply.count }.from(0).to(1)) }
+        it { expect { 3.times { save_reply_and_photo.call } }.to(change { Photo.count }.from(0).to(3)) }
```
tillprochaska, line 67: Hmmm, intuitively, I’d have expected it not to save a photo a second time if it already exists.

Edit: I guess we’d have to save the `file_id` for every photo to ensure that, so I’m fine with that.

tillprochaska:
> I'm finished refactoring - finally! I hope it was worth it.

Totally, this is way cleaner, I think! 😍

tillprochaska:
There’s just one thing I’ve been wondering.

What happens, if I send a message containing multiple photos with a different caption for each photo?

IMG_7B9B1A735C74-1

If I understood correctly, only the caption of the (last?) photo would be saved? And maybe it’d be totally fine to have a separate message per photo?

drjakob:
Best kitten.

PR #137, opened 2020-05-27
Implement custom date time helper

### LINKED ISSUE #114: Show exact time (and date) of requests and replies
As editor I need exact timing of questions and answers so I can put them into context with time-sensitive issues.

roschaefer:
close #114

Screenshot - 2020-05-27T214915 361
Screenshot - 2020-05-27T214735 111

### DIFF
--- a/app/helpers/date_time_helper.rb
+++ b/app/helpers/date_time_helper.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module DateTimeHelper
+  def date_time(date)
+    return date.strftime('%H:%M') if (Time.zone.now - 1.day) < date
+
+    date.strftime('%F %H:%M')
+  end
+end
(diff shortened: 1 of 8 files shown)


drjakob:
Please show date in German format: dd.mm.yyyy

app/helpers/date_time_helper.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module DateTimeHelper
+  def date_time(date)
+    return date.strftime('%H:%M') if (Time.zone.now - 1.day) < date
+
+    date.strftime('%F %H:%M')
```
tillprochaska, line 7: There already are time formats in the [German locale](
https://github.com/tactilenews/100eyes/blob/master/config/locales/de.yml#L280). All you need to do is add a new date formats for the short and long version, then you can use the i18n helpers:

```ruby
I18n.l date, format: :time_only # no idea what would be an appropriate name...
I18n.l date, fromat: :short_date_time
```ruby

PR #139, opened 2020-05-28
Show request feed in user profile

### LINKED ISSUE #125: Improve replies/request activity in user profile
Currently, the listing of requests a user has replied to is confusing – it mixes the title of the request with the replies of the user.

Screenshot 2020-05-22 11 37 29

As there already is a possibility to show all replies of a user for a specific, we don’t necessarily need to show all of the replies in the user profile. Instead we could just show their activity, i.e. the time a user has replied to a certain request. Clicking on this activity item would then show the complete conversation.

IMG_4366

tillprochaska:
Screenshot 2020-05-28 15 32 54

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -4,7 +4,7 @@ class User < ApplicationRecord
   include PgSearch::Model
   multisearchable against: %i[first_name last_name username note]
   has_many :replies, dependent: :destroy
-  has_many :requests, through: :replies
+  has_many :requests, -> { reorder(created_at: :desc).distinct }, through: :replies
   default_scope { order(:first_name, :last_name) }
   validates :email, presence: false, 'valid_email_2/email': true
 
--- a/app/components/request_row/request_row.html.erb
+++ b/app/components/request_row/request_row.html.erb
@@ -6,7 +6,7 @@
       <% end %>
     </div>
     <span class="RequestRow-date">
-      <%= time_ago_in_words(request.created_at) %>
+      <%= date_time(request.created_at) %>
     </span>
   </div>
     <div class="RequestRow-stats">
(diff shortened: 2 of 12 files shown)


app/models/user.rb
```diff
@@ -4,7 +4,7 @@ class User < ApplicationRecord
   include PgSearch::Model
   multisearchable against: %i[first_name last_name username note]
   has_many :replies, dependent: :destroy
-  has_many :requests, through: :replies
+  has_many :requests, -> { reorder(created_at: :desc).distinct }, through: :replies
```
tillprochaska, line 7: @roschaefer: The extra `reorder` is necessary, as otherwise, ActiveRecord creates the following SQL query:

```
"SELECT DISTINCT "requests".* FROM "requests" INNER JOIN "replies" ON "requests"."id" = "replies"."request_id" WHERE "replies"."user_id" = 1 ORDER BY "requests"."created_at" DESC, "replies"."created_at" DESC"
```

That’ll cause an SQL error (as `replies.created_at` is not part of the selected columns).

What we actually want is:

```
SELECT DISTINCT "requests".* FROM "requests" INNER JOIN "replies" ON "requests"."id" = "replies"."request_id" WHERE "replies"."user_id" = $1 ORDER BY "requests"."created_at" DESC
```

Don’t think that’s the best solution – and I think not having a default_scope but rather applying even simple ordering explicitly might help…? Thoughts?

roschaefer: efff, yes, also why the `distinct`? Is this necessary?

tillprochaska: If there are two replies to the same request by the same user, without `distinct`, `User.requests` would contain the request two times.

PR #167, opened 2020-05-30
feat: Send messages to one user via telegram+mail

roschaefer:
Screenshot - 2020-05-31T155047 301

@tillprochaska when we merge this, let's have a look if the search index gets updated during the migration. Perhaps we need to manually re-index the search index in `production`.


Gemfile
```diff
@@ -70,3 +70,5 @@ gem 'telegram-bot'
 gem 'pg_search', '~> 2.3'
 
 gem 'valid_email2', '~> 3.2'
+
+gem 'counter_culture', '~> 2.5'
```
tillprochaska, line 74: We need that to have a couter cache for photos through messages/replies on the request table, right?

app/components/chat_form/chat_form.css
```diff
@@ -0,0 +1,9 @@
+.ChatForm {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-end;
+}
+
+.ChatForm > button {
```
tillprochaska, line 7: I’d use a stack to add spacing between the `textarea` and the `button` – but for such simple cases, I’m also fine with a simple descendent selector.

app/models/message.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  scope :replies, -> { where.not(sender_id: nil) }
+
+  after_create do
+    send_email
+    send_telegram_message
+  end
+
+  def reply?
+    !!sender_id
```
tillprochaska, line 25:
```suggestion
    sender_id.present?
```

app/models/message.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def send_telegram_message
+    return unless recipient&.telegram_chat_id
+
+    Telegram.bots[Rails.configuration.bot_id].send_message(
+      chat_id: recipient.telegram_chat_id,
+      text: text
+    )
+  end
+end
```
tillprochaska, line 67: I think it would be nice to have this encapsulated in classes similar to `TelegramMessage` as well – just a note though, as this is not the scope of the PR.

config/locales/de.yml
```diff
@@ -1,4 +1,14 @@
 de:
+  application_name: 100eyes
+  components:
+    chat_form:
+      placeholder: Gib hier deine Nachfrage ein..
```
tillprochaska, line 5:
```suggestion
      placeholder: Gib hier deine Nachfrage ein…
```

drjakob:
Is it ensured that demands on individuals do not conflict with new crowd questions? An answer to a direct question should NOT appear as an answer to a crowd question to avoid misunderstandings.

config/locales/de.yml
roschaefer: Cool, didn't know there is an extra character for ellipsis. Is it good practice to use it? Like more semantic and still available in all fonts?

roschaefer:
@drjakob there is no concept of a direct question without a preceding request to all users. As a journalist, you have to first create a request which will be broadcasted. From then on, you can have a chat with a user. This chat will be attached to the request. As soon as another request is send (broadcasted) all conversations will be attached to the new request.

config/locales/de.yml
tillprochaska: I’m pretty sure that if a font doesn’t include a separate ellipsis glyph, the character will be [decomposed](https://unicode.org/reports/tr15/pdtr15.html) into three dots. That’s some first-class unicode magic… 😉

In general, all high-quality/professionally-made fonts will include that glyph – at least I’ve never come across one that didn’t.

PR #181, opened 2020-06-07
feat: Display date of answer on user profile

roschaefer:
@tillprochaska I was not able to write a test which would ensure that
requests (the keys) are ordered in descending order. Trying it out
locally seemed OK though..


app/views/users/show.html.erb
```diff
@@ -13,8 +13,8 @@
         <% end %>
 
         <%= c 'requests_feed' do %>
-          <% @user.replied_to_requests.each do |request| %>
-            <%= c 'requests_feed_item', user: @user, request: request %>
+          <% @user.replies_by_request.each do |request, replies| %>
+            <%= c 'requests_feed_item', user: @user, request: request, first_answer: replies.first %>
```
tillprochaska, line 17: Why not pass the date to `RequestsFeedItem` directly?

app/models/user.rb
```diff
@@ -63,4 +63,8 @@ def telegram?
   def email?
     email.present?
   end
+
+  def replies_by_request
+    replies.eager_load(:request).reorder('requests.created_at DESC, messages.created_at ASC').group_by(&:request)
```
tillprochaska, line 68: I think the requests feed should be ordered by the date/time a user has replied/interacted, right? Also, intuitively, I though it would make sense to display the date the users replied the last time (not the first time).

Example:
1. An editor creates a new request on Monday, 8am.
2. A user replies on Monday, 9am.
3. The same user replies a second time on Tuesday, 10am.

Then in that user’s activity feed, I should see:

> has replied to {Request} on Tuesday, 10am.

instead of

> has replies to {Request} on Monday, 9am.

PR #188, opened 2020-06-11
84 login

### LINKED ISSUE #84: simple login
As editor
I need a Login by password
in order to protect my user data.

tillprochaska:
* Enforces HTTPS for all connections and sets longer HSTS duration
* Uses HTTP Basic Auth for a simple login
* Throttles requests per IP address (max. 5 req/2 secs), we might need to tweak that

close #84


spec/auth_helper.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+RSpec.configure do |config|
+  config.before(:each) do
+    login = { user: 'user', password: 'password' }
+    allow(Rails.application.credentials).to receive(:login).and_return(login)
```
roschaefer, line 6: Maybe there is a cleaner way to do that without method stubs. We could e.g. put public secrets for testing into `secrets.yml`.

roschaefer: In case you need to change the tested code to make that happen, leave it as it is.

tillprochaska: Yes, tried that as well, but it doesn’t work as secrets and credentials cannot be used interchangeably. We could have [separate credentials for different environments](https://github.com/rails/rails/pull/33521), though.

PR #196, opened 2020-06-12
184 unknown content

roschaefer:
close #184 

Screenshot - 2020-06-13T003248 642

PR #200, opened 2020-06-13
fix: Blocked Telegram messages are marked but saved

### LINKED ISSUE #199: No request is created when some Telegram users disabled the bot
**Steps to reproduce:**
* Sign up to the Telegram bot (`/start`)
* Stop or disable the Telegram bot
* Send a request
* HTTP 500 error and no request is created

**Expected behaviour**
No 500 error and Telegram users who disabled the bot are ignored.

roschaefer:
fix: Blocked Telegram messages are marked but saved

This will avoid HTTP 500 error and unexpected behaviour

fix #199

Screenshot - 2020-06-13T142237 328

@tillprochaska please change target branch of this PR when you merge #196

### DIFF
--- a/app/components/chat_message/chat_message.css
+++ b/app/components/chat_message/chat_message.css
@@ -37,6 +37,8 @@
   display: block;
 }
 
-.ChatMessage-unknownContent {
+.ChatMessage-warnings {
+  display: flex;
+  flex-direction: column;
   color: var(--color-red);
 }
--- a/app/components/chat_message/chat_message.html.erb
+++ b/app/components/chat_message/chat_message.html.erb
@@ -25,10 +25,14 @@
     <div class="ChatMessage-meta">
       <a href="#<%= id %>"><%= date_time(message.created_at) %></a>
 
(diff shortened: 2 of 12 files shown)


tillprochaska:
I’m not sure wether flagging a message is enough – the way it’s implemented right now, an editor might not see the warning, because that message isn’t displayed anywhere. Maybe we should also indicate non-deliverable messages in users listings/profiels?

Also, instead of "blocked", maybe we should go for something more generic, e.g. "undeliverable"? Exceptions might occur not only because a user has blocked, but also for other reasons.

roschaefer:
@tillprochaska I think it's true that editors won't see the warning, but do they have to? We can check the database for analytical purposes. When you get in direct contact with a user it's helpful to see it, no?

The Exception is called 'Forbidden' so I do think it's that specific (blocked, not only undeliverable).

tillprochaska:
The Telegram gem raises a Forbidden  Exception for any 4xx response code, e.g. when the API token is invalid, a user isn’t found/has deletes their account, …

PR #218, opened 2020-06-17
fix: Use the most recent request as fallback

### LINKED ISSUE #215: Answers by Users will be dropped if the user has never received a request
Right now, we're onboarding new users. If a user has never received an initial question, then any messages to the Telegram Bot and any Email to Sendgrid will be silently dropped.

**Expected Behaviour**:
No message should be dropped. 
Either:
* The user gets a feedback: "Wait for us ...."
* Or a message will be saved to the most recent request, even if the user actually has never seen the request.

roschaefer:
As requested by @drjakob
https://github.com/tactilenews/100eyes/issues/215#issuecomment-645019491

fix #215

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -46,7 +46,7 @@ class User < ApplicationRecord
   end
 
   def active_request
-    received_requests.reorder(created_at: :desc).first
+    received_requests.reorder(created_at: :desc).first || Request.reorder(created_at: :desc).first
   end
 
   def telegram?
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -211,6 +211,14 @@ RSpec.describe User, type: :model do
       it { should eq(the_request) }
     end
 
+    describe 'if a request was created' do
+      before(:each) { the_request }
+      describe 'and afterwards a user joins' do
+        before(:each) { user }
+        it { should eq(the_request) }
+      end
+    end
+
     describe 'when many requests are sent to the user' do
       before(:each) do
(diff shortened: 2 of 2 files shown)


app/models/user.rb
```diff
@@ -46,7 +46,7 @@ def channels
   end
 
   def active_request
-    received_requests.reorder(created_at: :desc).first
+    received_requests.reorder(created_at: :desc).first || Request.reorder(created_at: :desc).first
```
tillprochaska, line 49: I think that’s just equivalent to the following:

```suggestion
    Request.reorder(created_at: :desc).first
```

Two cases:

* A user has never received a request, then it’ll be the most recent request as a fallback
* A user has received a request, then it’ll be the most recent request, as all user’s receive the same requests.

Do we still need an active request per user? I think from a UX perspective, right now, editors already expect the question to be the same for all users at the same time. So why don't we use `Request.active` as we did previously?

tillprochaska:
With this fix, messages will still be dropped if no request has ever been sent. I think we should handle that case as well: Either bounce the message, or alternatively, save them without an associated request (also then, we wouldn’t have a UI for it). Will open a separate issue.

app/models/user.rb
roschaefer: @drjakob @FrauCsu I talked with @tillprochaska about this problem. I don't really like this solution in this PR nor do I like the solution proposed here by Till.

For the future: Would be a reasonable solution to automatically send the user on registration (via Mail or Telegram) the most recent request as a message? So the user always knows, what he/she is responding to?

I'm going to merge this as it is for now.

PR #230, opened 2020-06-19
Case-insensitive emails

### LINKED ISSUE #224: Emails should be compared case-insensitively
Some emails are bounced, even though the email address matches. In that particular case, the `From` header included the email address *and* a display name:

```
From: First Last <first.last@provider.test>
```

tillprochaska:
Close #224

#### To do:

- [x] Do not override `find_by_email`
- [x] Add mailbox test case
- [x] Add `EmailMessage` decorator test case


spec/models/user_spec.rb
```diff
@@ -31,7 +41,8 @@
   describe '#email' do
     it 'must be unique' do
       create(:user, email: 'user@example.org')
-      expect { build(:user, email: 'user@example.org').save!(validate: false) }.to raise_error(ActiveRecord::RecordNotUnique)
```
tillprochaska, line 34: @roschaefer What was the reason you chose to test for the database exception, not the validation exception?

app/decorators/email_message.rb
```diff
@@ -24,7 +24,7 @@ def initialize_text(mail)
   end
 
   def initialize_user(mail)
-    User.find_by(email: mail.from)
+    User.find_by_email(mail.from.first)
```
roschaefer, line 27: Are you sure? `mail.from` would be an array and that's intented, as far as I can tell.

spec/models/user_spec.rb
roschaefer: To ensure the unique constraint is in the database

app/models/user.rb
```diff
@@ -8,12 +8,16 @@ class User < ApplicationRecord
...
   default_scope { order(:first_name, :last_name) }
-  validates :email, uniqueness: true, allow_nil: true, 'valid_email_2/email': true
+  validates :email, uniqueness: { case_sensitive: false }, allow_nil: true, 'valid_email_2/email': true
 
   before_validation do
     self.email = nil if email.blank?
   end
 
+  def self.find_by_email(email)
+    find_by('lower(email) = ?', email.downcase)
```
roschaefer, line 18:
```suggestion
    find_by(email: Array(email).map(&:downcase))
```
Maybe? To allow both types, string and array?

app/models/user.rb
```diff
@@ -8,12 +8,16 @@ class User < ApplicationRecord
   has_many :replied_to_requests, -> { reorder(created_at: :desc).distinct }, source: :request, through: :replies
   has_many :received_requests, -> { reorder(created_at: :desc).distinct }, source: :request, through: :received_messages
   default_scope { order(:first_name, :last_name) }
-  validates :email, uniqueness: true, allow_nil: true, 'valid_email_2/email': true
+  validates :email, uniqueness: { case_sensitive: false }, allow_nil: true, 'valid_email_2/email': true
 
   before_validation do
     self.email = nil if email.blank?
```
roschaefer, line 14:
```ruby
self.email = self.email.downcase
```

PR #267, opened 2020-06-29
fix: Avoid `name` setter to type in more than 3 words for the name

### LINKED ISSUE #266: BUG Truncate Names
The name of community members is shortened to two parts. For example: "Erika von Mustermann" becomes "Erika von". This is not a desired behavior, since names can consist of more than two components (double names, titles of nobility, non-German names). For me its a bug, not a feature.

roschaefer:
close #266


tillprochaska:
@roschaefer Can we also completely remove the `name` setter?

https://github.com/tactilenews/100eyes/blob/154232ba773bf227b76465efa69d3b5f8845045f/app/models/user.rb#L35-L39

PR #269, opened 2020-06-29
feat: Add default sender username to translations

### LINKED ISSUE #223: Set sender name to "50survivors Redaktion"
Currently, emails sent via 100eyes will show up in user’s inboxes as "mail" or "mail@..." which might be confusing. Next to the email address, we should also set a dislplay name:

```
50survivors Redaktion <mail@...>
```

Related issue #204

roschaefer:
close #223


app/mailers/message_mailer.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class MessageMailer < ApplicationMailer
+  default username: I18n.t('message_mailer.username')
```
tillprochaska, line 4: How does that work? I can’t find documentation on the `username` option anywhere…

config/locales/de.yml
```diff
@@ -15,15 +15,16 @@ de:
...
     new_message_email:
       subject: Die Redaktion hat eine neue Frage an Sie
 
   save: Speichern
   discard: Änderungen verwerfen
 
   dashboard:
-    title: Willkommen bei 100eyes. Nutzen Sie die Intelligenz der Community.
-    subtitle: Mit 100eyes kannst du 50 Menschen aus deiner Region Fragen stellen.
+    title: Willkommen bei 50survivors. Nutze die Intelligenz der Community.
```
tillprochaska, line 26: I’m not 💯 sure, but if I recall correctly @FrauCsu and @drjakob preferred the web app to be branded 100eyes, while outside communication should be customized absed on the project/newsroom etc.

@drjakob @FrauCsu, can you clarify?

Should this be changed to "Willkommen bei 50survivors." and "Mit 50survivors kannst du…" or stay as it is right now?
Bildschirmfoto 2020-06-29 um 19 30 21

Same question, should we update this to "50 survivors"?
Bildschirmfoto 2020-06-29 um 19 30 44

drjakob: Its correct.- Webapp should be 100eyes. 50survivors is the name of one research (Recherche). Recherchen are the parents of questions.

tillprochaska:
@roschaefer Works perfectly! :)

PR #313, opened 2020-08-18
254 copy to clipboard

### LINKED ISSUE #254: Copy reply to Clipboard
As an editor, I would like to copy a reply to the clipboard with one click in order to process the content in other applications

tillprochaska:
Merge #318 first. Close #254 

***


app/components/chat_message/chat_message.js
```diff
@@ -35,4 +35,24 @@ export default class extends Controller {
...
+
+    navigator.clipboard.writeText(clipboardText).then(() => this.onCopy());
+  }
+
+  onCopy() {
+    this.element.classList.add('ChatMessage--copied');
+
+    setTimeout(() => {
+      this.element.classList.remove('ChatMessage--copied');
+    }, 2000);
```
tillprochaska, line 56: Move this into a constant…

```suggestion
    setTimeout(() => {
      this.element.classList.remove('ChatMessage--copied');
    }, SUCCESS_MESSAGE_DURATION);
```

drjakob (approved):

FrauCsu:
Approved

app/components/chat_message/chat_message.js
```diff
@@ -35,4 +37,24 @@ export default class extends Controller {
...
+  copy() {
+    const text = this.textTarget.innerText;
+    const sender = this.data.get('sender-name');
+    let clipboardText = `${sender}: ${text}`;
+
+    if (!sender) {
+      clipboardText = text;
+    }
+
+    navigator.clipboard.writeText(clipboardText).then(() => this.onCopy());
```
roschaefer, line 50: I learned something new today!

app/components/chat_message/chat_message.js
```diff
@@ -35,4 +37,24 @@ export default class extends Controller {
...
+  }
+
+  onCopy() {
+    this.element.classList.add('ChatMessage--copied');
+
+    setTimeout(() => {
```
roschaefer, line 56: Can't we implement that with CSS?

roschaefer (approved):
clipboard

@tillprochaska feel free to refactor the effect with CSS if possible.

app/components/chat_message/chat_message.js
tillprochaska: Yes, that should be possible, e.g. using different transition delays – however I think that’d make it more opaque/more difficult to understand…

PR #314, opened 2020-08-18
94 star replies

tillprochaska:
Close #94. Merge #318 and #313 first!

***

@roschaefer, backend implementation isn’t finished yet – I opened this PR so @FrauCsu and @drjakob can already give feedback on the UI interaction, wording etc..

***

PR #315, opened 2020-08-18
Highlight targeted message

tillprochaska:
Merge #318, #314, and #313 first.

***

We’ve had the feature that allows you to copy a direct link to a reply for some time now. That’s great for collaboration in newsrooms, as one editor can easily share a specific reply with another editor.

This PR adds a subtle outline to the respective message, making it easier to spot it.

***

### DIFF
--- a/app/components/chat_message/chat_message.css
+++ b/app/components/chat_message/chat_message.css
@@ -2,6 +2,12 @@
   background-color: var(--color-gray-light);
   padding: var(--spacing-unit-xs) var(--spacing-unit-s);
   border-radius: var(--border-radius-l);
+  border: 1px solid transparent;
+}
+
+.ChatMessage:target {
+  border-color: var(--color-focus);
+  box-shadow: var(--input-focus-shadow);
 }
 
 .ChatMessage-meta {


drjakob (commented):
Don‘t understand how to copy the link for a specific reply

tillprochaska:
> Don‘t understand how to copy the link for a specific reply

Right-click on the time stamp/date – just as you’d do in apps like Slack or here on GitHub.

drjakob:
Thanks. I only use the software occasionally. Probably useful for powerusers.

FrauCsu:
Don't understand, why to klick on time stamp either. Does not make sense to me. But: HIGHLIGHT is great!

tillprochaska:
> Don't understand, why to klick on time stamp either. Does not make sense to me.

That’s how most apps do it – if you need a permalink to a specific item/entry, click on the time stamp and I think it makes sense to go with a convention like this. But as Jakob pointed out, this probably has a very specific use case and we shouldn’t spent too much time thinking about this rn, even though there might be better, more discoverable places for this.

**Slack**
Bildschirmfoto 2020-08-19 um 22 42 13

**Twitter**
Bildschirmfoto 2020-08-19 um 22 40 21

**GitHub**
Bildschirmfoto 2020-08-19 um 22 41 15

roschaefer (approved):
Sure!

PR #323, opened 2020-08-20
Inter webfont

tillprochaska:


drjakob:
Typography rules

public/fonts/subsets.txt
```diff
@@ -0,0 +1,10 @@
+# This file includes Unicode ranges used to subset
+# the Inter webfont for use with this application.
+# You may use a tool like `pyftsubset` to create
+# a subset from source files.
```
roschaefer, line 4: Huh? :thinking:

roschaefer (approved):
Ok, just some imports, looks good enough for me :+1: 

I just don't understand one thing :point_down:

tillprochaska:
> Ok, just some imports, looks good enough for me 👍

Yes, exactly – I did already use that font family for some time, but you would only see it if it was installed locally on your computer. The default system font on macOS ist pretty similar, both in terms of its design characteristic and the number of weights etc. But on other platforms like Windows and Linux desktops, it might have been quite different.

public/fonts/subsets.txt
tillprochaska: The [Inter](https://rsms.me/inter/) font family contains glyphs for many different alphabets (Latin, Cyrillic, Greek) as well as many alternate characters etc. That’s really cool and it’s a really high-quality font family. However, in most situations, we won’t need all those glyphs.

Subsetting is the process of keeping only those glyphs you actually need, in our case basic Latin characters plus some basic punctuation etc, thus reducing the file size. In our case, subsetting the font reduced the file size from ~100 kB per weight (we use four different weights) to around 16-18kB.

In Unicode, similar characters have similar codes, i.e. all basic Latin characters have hex codes between 0020 and 007F which are listed in this file for future reference (i.e. in case we want to update the font family).

PR #330, opened 2020-08-23
feat: Strip HTML from emails

### LINKED ISSUE #229: Strip HTML from emails
Some messages contain HTML markup. Most email clients will send both, an HTML and a plaintext variant (even if you compose the email as HTML). I’d guess (but haven’t verified it) that in these cases, the email didn’t include a plaintext variant. If that’s the case, all we can do is to strip HTML markup.

A few different edge cases: 
* /requests/3#message-193
* /requests/2#message-90
* /requests/5#message-375
* /requests/5#message-379
* /requests/5#message-420

roschaefer:
close #229 

As discussed, this PR does not alter existing data in the database.


tillprochaska:
@roschaefer: I added a few more test cases for real-world edge cases. I think we should handle line breaks (at least those introduced by `<br>` and `<p>` tags, possibly by other block elements as well) – otherwise without line breaks long-form emails will become difficult to read.

roschaefer:
> @roschaefer: I added a few more test cases for real-world edge cases. I think we should handle line breaks (at least those introduced by `<br>` and `<p>` tags, possibly by other block elements as well) – otherwise without line breaks long-form emails will become difficult to read.

done

tillprochaska (approved):
Nice! :)

PR #338, opened 2020-08-28
Remove grayscale effect on footer logos

### LINKED ISSUE #335: change PF and BMBF Logos - colored
URGENT: The above named institutions insist on displaying their logos correctly, that means

BMBF colored
PF colored

Please change today: Friday, 28th before 6pm

tillprochaska:
Close #335


roschaefer (approved):
:art: yep

PR #369, opened 2020-09-21
build: Remove error-prone bundle images

roschaefer:


mattwr18 (changes requested):
Can you give some context to this `PR` @roschaefer ??
I don't think anyone this week ran into any of this errors, did they?

roschaefer:
> Can you give some context to this `PR` @roschaefer ??
> I don't think anyone this week ran into any of this errors, did they?

Yes, we had difficulties during setup. Here's what can happen:

1. You `docker-compose up --build` the images, this will also create volumes
2. You `git pull` the latest changes, including an update to `Gemfile.lock`
3. You `docker-compose up --build` again, to rebuild the images with the new gem versions
4. Even though the images are rebuilt, when the containers start, the volumes of step 1 will overwrite the bundles directory resulting in `bundler` complaining about missing gem versions

mattwr18 (approved):
As discussed, I just run `docker-compose exec app bundle` every time I pull down and don't run into this issue. My reservation is that it takes longer to get the app up and running without these volumes, but if you think it makes it easier for new developers who might not be used to running into this issue, why not?

tillprochaska (approved):

PR #371, opened 2020-09-21
160 feat pull new replies

### LINKED ISSUE #160: Push Notification for Incoming Answers
```
As an editor
I would like to receive push notifications on new responses. (A refresh hint is missing!)
So that I don't need to refresh the page to see new responses.
```

mattwr18:
close #160

### DIFF
--- a/app/views/requests/show.html.erb
+++ b/app/views/requests/show.html.erb
@@ -1,3 +1,7 @@
+<% content_for :notifications do %>
+  <%= c 'request_notification', request: @request %>
+<% end %>
+
 <%= c 'box', styles: [:large, :gray] do %>
   <%= c 'wrapper' do %>
     <%= c 'stack', space: :small do %>
--- a/app/components/button/button.css
+++ b/app/components/button/button.css
@@ -60,3 +60,7 @@
 .Button--positive {
   --button-color: var(--color-green);
 }
+
+.Button--underline {
+  text-decoration: underline;
+}
--- a/app/components/nav_bar/nav_bar.css
+++ b/app/components/nav_bar/nav_bar.css
@@ -1,8 +1,4 @@
 .NavBar {
-  position: sticky;
-  top: 0;
-  z-index: 99;
-
   display: flex;
   align-items: center;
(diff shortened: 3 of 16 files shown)


smalbroc:
@mattwr18 Could I get a screenshot showcasing the behavior?

app/views/requests/show.html.erb
```diff
@@ -1,3 +1,7 @@
+<% content_for :notifications do %>
+  <%= c 'request_notification', request: @request %>
+<% end %>
+
```
smalbroc, line 4: @mattwr18 what about the request overview page? The notifications are not shown there.

smalbroc: speaking of the request overview page, @mattwr18 : What about notifications at the chat page? `user_request GET    /users/:user_id/requests/:id(.:format)`
Screenshot - 2020-09-23T112759 687

smalbroc (changes requested):

app/views/requests/show.html.erb
mattwr18: > @mattwr18 what about the request overview page? The notifications are not shown there.

We talked about this, and agree that it would be cool to have, but we see that as another feature that would take more thought and time to develop. At the moment, we just have a link on the request show page that let's the editor know that there is a new message(reply) to the request and the link refreshes the page. If the user is on the index page, what should the link do? If we are polling for messages from all requests, we would need to render more info about the messages and to which request they belong. There are other cases to consider, we will create a ticket to revisit this later.

mattwr18: > speaking of the request overview page, @mattwr18 : What about notifications at the chat page? `user_request GET /users/:user_id/requests/:id(.:format)`
> 

Good that you brought this up. At the moment, we are considering how helpful this page is in the first place since it sends the editor away from the request show page simply to be able to send a reply to the community member, but we could easily render an input field a the bottom of the chat conversation between an editor and that community member on the request show page itself. This would probably be better UX since we don't need to switch pages.

smalbroc (approved):

PR #377, opened 2020-09-22
Assign tags to users, send requests only to users with certain tags

### LINKED ISSUE #6: Ask subgroups
```gherkin
Feature: Ask subgroups
  As editor
  I want to send a research question to a subgroup e.g. entrepreneurs, parents, dog owners 
  in order to avoid asking questions which are irrelevant for certain contributors
  e.g. "How much is the rent?" for property owners who don't pay any rent
```

Screenshot - 2020-09-21T153446 552

mattwr18:
Close #6

This PR adds a simps UI to assign tags to users and then use tags to send requests only to users with certain tags.

* Colors are randomly assigned to tags, but are persistent, i.e. the same tag always has the same color, but you cannot select which color is assigned to a tag.
* Tags are combined using boolean AND, i.e. only users are selected that have all the tags.
* There’s a small note/help tags displayed informing the user about the number of users with the selected tags. (I’m actually not super happy with the placement of help text – and I think that might have been come up during the user test with Anna as well – as it can easily be overlooked. However, placing it below the tags input is no option as well, as it will be invisible when the autocomplete dropdown is opened.

Actually, I have some ideas how to solve that if we slightly reworked the whole form (#388), but I’d rather not do that right now.

Bildschirmfoto 2020-09-24 um 20 03 05

### DIFF
--- a/app/components/request_form/request_form.js
+++ b/app/components/request_form/request_form.js
@@ -44,7 +44,7 @@ export default class extends Controller {
           return;
         }
 
-        const message = messageTemplates.other.replace('{count}', count);
+        const message = messageTemplates.other.replace('%{count}', count);
         this.membersCountTarget.innerHTML = message;
       },
     });
--- a/app/components/request_form/request_form.js
+++ b/app/components/request_form/request_form.js
@@ -1,4 +1,5 @@
 import { Controller } from 'stimulus';
+import Rails from '@rails/ujs';
 import sanitize from '../../../frontend/helpers/sanitize.js';
 
 const template = ({ message, notes }) => {
@@ -7,10 +8,11 @@ const template = ({ message, notes }) => {
 };
 
 export default class extends Controller {
-  static targets = ['preview', 'message'];
+  static targets = ['preview', 'message', 'membersCount'];
 
   connect() {
     this.updatePreview();
+    this.updateMembersCount();
   }
 
   updatePreview() {
@@ -19,5 +21,32 @@ export default class extends Controller {
     };
 
     this.previewTarget.innerHTML = template(data);
+  }
+
+  updateMembersCount(event) {
+    if (!event || event.detail.tags.length <= 0) {
+      this.membersCountTarget.hidden = true;
+      return;
+    }
+
+    const messageTemplates = JSON.parse(this.data.get('members-count-message'));
+    const tags = event.detail.tags;
+
+    Rails.ajax({
+      url: '/users/count',
+      type: 'GET',
+      data: new URLSearchParams({ tag_list: tags }).toString(),
+      success: ({ count }) => {
+        this.membersCountTarget.hidden = false;
+
+        if (count == 1) {
+          this.membersCountTarget.innerHTML = messageTemplates.one;
+          return;
+        }
+
+        const message = messageTemplates.other.replace('{count}', count);
+        this.membersCountTarget.innerHTML = message;
+      },
+    });
   }
 }
--- a/app/components/tags_input/tags_input.css
+++ b/app/components/tags_input/tags_input.css
@@ -33,6 +33,20 @@
   justify-content: space-between;
 }
 
-.TagsInput-count {
+.tagify__dropdown__item--active,
+.tagify__dropdown__item:hover {
+  background-color: var(--color-gray-light);
+}
+
+.TagsInput-name {
+  padding: 0 var(--spacing-unit-xs);
+  border-radius: var(--border-radius);
+  background-color: var(--tag-bg) !important;
+  color: #fff;
+}
+
+.TagsInput-count,
+.tagify__dropdown__item--active .TagsInput-count {
   font-size: var(--font-size-xs);
+  color: var(--color-text-light);
 }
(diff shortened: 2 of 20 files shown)


drjakob:
Textvorschlag **Empfänger einschränken**: "Wähle Tags, um den Empfängerkreis deiner Frage einzuschränken. Die Tags summieren sich: "Lehrer" und "Biologie" verschickt die Frage an alle Biologie-Lehrer. Und nur an sie."

tillprochaska:
@roschaefer @mattwr18: Sorry, didn’t rebase this yet – will do that after your reviews! :)

smalbroc (approved):

app/components/request_form/request_form.js
```diff
@@ -20,4 +22,31 @@ export default class extends Controller {
 
     this.previewTarget.innerHTML = template(data);
   }
+
+  updateMembersCount(event) {
+    if (!event || event.detail.tags.length <= 0) {
```
mattwr18, line 27:
```suggestion
    if (!event || event.detail.tags.length === 0) {
```
or `< 1`... it could never be less than one right?
or `!!event.detail.tags.length`?

app/components/tags_input/tags_input.css
```diff
@@ -28,3 +26,27 @@
...
+
+.tagify__dropdown__item--active,
+.tagify__dropdown__item:hover {
+  background-color: var(--color-gray-light);
+}
+
+.TagsInput-name {
+  padding: 0 var(--spacing-unit-xs);
+  border-radius: var(--border-radius);
+  background-color: var(--tag-bg) !important;
```
mattwr18, line 44:
```suggestion
  background-color: var(--tag-bg);
```
this isn't necessary, right?
I tested it without and it seems fine...

app/components/tags_input/tags_input.js
```diff
@@ -1,18 +1,92 @@
...
-    new Tagify(this.inputTarget, {
-      originalInputValueFormat: tags => tags.map(tag => tag.value).join(','),
-      whitelist: JSON.parse(this.data.get('options')),
+    this.tagify = new Tagify(this.inputTarget, {
+      originalInputValueFormat: tags => {
+        return tags.map(tag => tag.value).join(',');
+      },
+
+      whitelist: JSON.parse(this.data.get('available-tags')),
+      enforceWhitelist: !JSON.parse(this.data.get('allow-new')),
```
mattwr18, line 58: I have found this [issue in tagify](https://github.com/yairEO/tagify/issues/465) reporting the weird behavior for `request_form` adding them invalidating and removing the tag... Haven't had time to go through it...

app/components/tags_input/tags_input.rb
```diff
@@ -2,24 +2,24 @@
 
 module TagsInput
   class TagsInput < ApplicationComponent
-    def initialize(value: [], options: [], **kwargs)
+    def initialize(value: [], available_tags: [], allow_new: true, **kwargs)
```
mattwr18, line 5:
```suggestion
    def initialize(value: [], allow_new: true, **kwargs)
```
can be removed now, right?

mattwr18 (commented):
Have a look at my comments :point_down:  and please review the fix to the bug where the taggings_count can't be use to identify users with a specific tag...

app/components/tags_input/tags_input.js
tillprochaska: Thanks, I added that as a separate issue: https://github.com/tactilenews/100eyes/issues/398

app/components/request_form/request_form.js
tillprochaska: @mattwr18: I like to test for `<= 0` instead of `== 0` so that all values that an integer could possibly have are handled. There are many situations (see below) in JavaScript where this comes in handy. Even though it’s rather unlikely in this specific example, I like to make it a habit.

I guess you’d have concerns about this making the code less clear?

```
> null <= 0
< true
```

```
> null == 0
< false 
```

```
> 'lorem ipsum'.indexOf('hello world')
< -1
```

app/components/tags_input/tags_input.rb
tillprochaska: That’s actually the only thing I didn't like about your fix. I think the TagsInput view component should not rely on business logics/concrete models

tillprochaska:
@mattwr18 Thanks again for fixing that weird bug! :)

tillprochaska:
@roschaefer: I had a look at Matt’s comments and would be ready for your review now! :)

app/components/request_form/request_form.js
```diff
@@ -20,4 +22,31 @@ export default class extends Controller {
...
+      data: new URLSearchParams({ tag_list: tags }).toString(),
+      success: ({ count }) => {
+        this.membersCountTarget.hidden = false;
+
+        if (count == 1) {
+          this.membersCountTarget.innerHTML = messageTemplates.one;
+          return;
+        }
+
+        const message = messageTemplates.other.replace('{count}', count);
```
roschaefer, line 47:
```suggestion
        const message = messageTemplates.other.replace('%{count}', count);
```

roschaefer (approved):
render javascript

This looks good, although I find the render method a great use-case for client-side rendering.

PR #381, opened 2020-09-23
feat: Add voice model

roschaefer:


app/decorators/email_message.rb
```diff
@@ -1,18 +1,22 @@
...
   def initialize(mail)
+    @voice = nil
     @text = initialize_text(mail)
     @sender = initialize_user(mail)
     @message = initialize_message(mail)
     @photos, @unknown_content = initialize_photos_and_unknown_content(mail)
     @message.unknown_content = unknown_content
+    @photos.each do |photo|
+      @message.association(:photos).add_to_target(photo)
+    end
```
mattwr18, line 19: how is this change related to this `PR`? what issue does this solve?

app/decorators/telegram_message.rb
```diff
@@ -82,4 +87,20 @@ def initialize_photos(telegram_message)
...
+    bot_token = "bot#{Telegram.bots[Rails.configuration.bot_id].token}"
+    telegram_file = telegram_message[:voice]
+    file_id = telegram_file[:file_id]
+    uri = URI("https://api.telegram.org/#{bot_token}/getFile")
+    uri.query = URI.encode_www_form({ file_id: file_id })
+    response = JSON.parse(URI.open(uri).read)
+    file_path = response.dig('result', 'file_path')
+    remote_file_location = URI("https://api.telegram.org/file/#{bot_token}/#{file_path}")
+    voice.attachment.attach(io: URI.open(remote_file_location), filename: File.basename(remote_file_location.path))
+    voice
```
mattwr18, line 104:
```suggestion
    retrieve_message_type_and_attach(message_type)
```

I think this code is so similar to what we do for photos that we could DRY up the code a bit by extracting it to its own method...

app/models/user.rb
```diff
@@ -28,7 +28,6 @@ def reply(message_decorator)
       message = message_decorator.message
       message.request = request
       message.save!
-      message.photos << message_decorator.photos
```
mattwr18, line 31: I guess this is related to the other change you made :point_up_2: 
You are solving an issue that is not related to adding voice support, but you've noticed while working on `voice`... I would argue this is `code creep` :open_mouth: 
It would be nice to create a clear ticket with the issue, so that our future selves/other developers can get a quick overview of why these changes were introduced.

mattwr18 (changes requested):

PR #387, opened 2020-09-24
Remove street info from user as it is not needed

phoeinx:
Street info is not wanted according to Astrid


mattwr18 (approved):
Looks good! I rebased it since there were merge conflicts... so that's why you might notice I force-pushed.

PR #408, opened 2020-10-01
build: Add .dockerignore for security

roschaefer:
@tillprochaska while working on traefik and docker I almost exposed our
security keys `config/credentials/{production,staging,development}.key`
because we didn't had a `.dockerignore`.

For this reason I find it bad practice to have password_files in
application directory. Isn't there a better way?


tillprochaska:
@roschaefer Is this still relevant? I saw this change is included in #409…

roschaefer:
> @roschaefer Is this still relevant? I saw this change is included in #409…

@tillprochaska correct!

PR #409, opened 2020-10-02
Create repeatable automatic setup with ansible+traefik+docker

roschaefer:
Follow the instructions in [the README](./ansible/README.md).


tillprochaska (commented):
Could you provide some insight in how this works (i.e. not how to set up a new instance, but what’s happening behind the scenes)? As I’ve never used Docker in production, I can’t really review this PR, but instead I’ve got some newbie questions… 😄

* What is the host OS at Netcup?
* Does Netcup provide automated backups/snapshots?
* How does this affect local development setups?
* Which of the `docker-compose.yml` files is suitable for which use cases?
* What about default system hardening?
* How would you add new users to the host system?
* Where does traefik store TLS certificate files?

Apart from that, it would be nice to mention how to deploy new releases to production. I guess it should be enough to run the `deployment` role? Maybe add a separate playbook for that, too?

roschaefer:
> Could you provide some insight in how this works (i.e. not how to set up a new instance, but what’s happening behind the scenes)? As I’ve never used Docker in production, I can’t really review this PR, but instead I’ve got some newbie questions… 
> 
> * What is the host OS at Netcup?

Screenshot - 2020-10-03T192143 621

> * Does Netcup provide automated backups/snapshots?

Screenshot - 2020-10-03T192315 843

> * How does this affect local development setups?

The goal is: Not at all. I would update the `README.md` to explain how to create your personal `.env` file from a `.env.template`.

> * Which of the `docker-compose.yml` files is suitable for which use cases?

* `docker-compose config` reads in `docker-compose.yml` and `docker-compose.override.yml` by default (for development)
* `docker-compose -f docker-compose.yml -f docker-compose.prod.yml` is used to build and publish docker images (on CI/CD)
* `docker-compose -f docker-compose.yml -f docker-compose.prod.yml -f docker-compose.traefik.yml` is used in production and automagically refreshes TLS certificates among other things



> * What about default system hardening?

I followed [this tutorial](https://www.cyberciti.biz/faq/how-to-disable-ssh-password-login-on-linux/) and this also disables `PAM`. Furthermore, every incoming traffic except SSH, HTTP and HTTPs is denied. We would have to whitelist everything else.

> * How would you add new users to the host system?

Currently not implemented. But I thought about collecting our public keys (jakob's, yours, mine) and putting them into location where they can be uploaded. I don't know what's the best workflow here.

> * Where does traefik store TLS certificate files?

In a shared volume `/home/ansible/traefik/:/traefik/` and there it's called `acme.json`. You can have a look with e.g. `cat .traefik/acme.json` after login with your ansible user.


> * What happens if deployment happens (and the DB container is rebuilt) while Postgres is running a transaction?

This configuration is installation only and does not do automatic deployments, yet. Restarting containers does wait for transactions to finish, as far as I know.


> * What happens when one of the containers terminates for some reason?

Traefik offers a `/ping` route for health checks. I didn't look into this, but I think you can configure it to call e.g. the application on a route that hits the database, for example. I guess there is also a configuration to automatically restart containers.

> 
> I might be incorrect, but I think Rails logs are only saved inside of the `app` container, i.e. they will get deleted with the container (e.g. on deployments).

I used `docker logs <CONTAINER_ID>`. They are available when the container has stopped.

> 
> Apart from that, it would be nice to mention how to deploy new releases to production. I guess it should be enough to run the `deployment` role? Maybe add a separate playbook for that, too?

I didn't get there yet. But deployment would roughly look like this:
* New docker images get build and published by CI
* CI runs ansible which:
  * bumps image version tag in docker-compose.yml
  * pulls new images
  * restarts containers
  * runs database migrations

tillprochaska:
> Screenshot - 2020-10-03T192315 843

But does it support creating those automatically, e.g. create a snapshot every day and keep them for seven days? If not, how/where would we backup app data, certificates, …

> I followed this tutorial and this also disables PAM. Furthermore, every incoming traffic except SSH, HTTP and HTTPs is denied. We would have to whitelist everything else.

That tutorial unfortunately does not explain why disabling PAM is necessary…?

> I used docker logs <CONTAINER_ID>. They are available when the container has stopped.

Additionally, maybe we should mount the `log` path as a volume? Otherwise, log not printed to stdout/stderr might get lost. Also, deploying a new release means the old container is deleted and a new one is created, i.e. logs are lost.

ansible/roles/installation/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+    -  /home/ansible/traefik/
+- name: Add traefik admin user with htpasswd file
+  htpasswd:
+    path: /home/ansible/traefik/passwords
+    name: "{{ traefik.user }}"
+    password: "{{ traefik.password }}"
+- name: Copy rails credentials file
+  copy:
+    src: ../config/credentials/ansible.yml.enc
+    dest: /home/ansible/config/credentials/production.yml.enc
```
tillprochaska, line 17: What’s happening here? Why can we simply use the Ansible vault file as a Rails credentials file?

tillprochaska:
@roschaefer You might want to have a look at https://github.com/dev-sec/ansible-os-hardening

drjakob (approved):

tillprochaska (approved):
As I said previously, my understanding of Docker in production and Traefik is very limited, so I can’t really give a review for this PR, but I’m going to approve so you can merge this.

roschaefer:
@tillprochaska you could run `ansible-playbook site.yml -i inventories/100eyes --ask-vault-pass` in the `ansible` directory which would run the installation on the Schweine-System.

tillprochaska:
> @tillprochaska you could run ansible-playbook site.yml -i inventories/100eyes --ask-vault-pass in the ansible directory which would run the installation on the Schweine-System.

What to do in order to setup a new server or to deploy a new version of the app is pretty clear from the documentation you wrote, I was talking about reviewing the underlying implementation (i.e. the host server setup, Traefik config etc).

ansible/README.md
```diff
@@ -0,0 +1,73 @@
+# Local setup
+
+Install ansible roles:
+```
+ansible-galaxy install geerlingguy.docker
+ansible-galaxy install geerlingguy.pip
+```
```
tillprochaska, line 7:
````suggestion
Our setup uses a number of community-maintained Ansibles roles. Install them using the `ansible-galaxy` command:

```
ansible-galaxy collection install community.general
ansible-galaxy install geerlingguy.docker
ansible-galaxy install geerlingguy.pip
```
````

PR #423, opened 2020-10-10
Add health endpoint

### LINKED ISSUE #364: Add /health endpoint

tillprochaska:
Closes #364.

This adds a simple `/health` route for use with our monitoring provider (and for that purpose skips authentication). This route simply returns a 200 status code if everything is good – in case there are database related errors (a bad connection, pending migreations), Rails will always send an error response. In cases where the app isn’t running at all, obviously, the request will be unsuccessful as well.

### DIFF
--- a/app/controllers/health_controller.rb
+++ b/app/controllers/health_controller.rb
@@ -4,6 +4,11 @@ class HealthController < ApplicationController
   skip_before_action :authenticate
 
   def index
+    ActiveRecord::Base.connection.execute('select 1')
+    ActiveRecord::Migration.check_pending!
+  rescue StandardError
+    render status: :service_unavailable, json: { status: 'Service Unavailable' }
+  else
     render json: { status: 'OK' }
   end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/health_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class HealthController < ApplicationController
+  skip_before_action :authenticate
+
+  def index
+    render json: { status: 'OK' }
```
roschaefer, line 7: In `production` I can render the index route even if there are pending migrations.

roschaefer (changes requested):
Looks good, except for a double-check if the database is doing OK.

app/controllers/health_controller.rb
tillprochaska: That’s good to know – didn’t know that behaviour is different in prod and development. I added an explicit check for a healthy database connection and pending migrations.

roschaefer (approved):
excellent

PR #430, opened 2020-10-15
Upgrade pg_search after issue closed by new release

### LINKED ISSUE #385: BUG: Search is broken with update to 2.3.3
Executing a search returns `no implicit conversion of nil to Hash`.
This comes from `gems/pg_search-2.3.3/lib/pg_search/document.rb:22`
```
{ query: args.first }.merge(PgSearch.multisearch_options)
```
PgSearch.multisearch_options is nil
This is configured in our `config/initializers/pg_search.rb` like so:
```
PgSearch.multisearch_options = {
  using: { tsearch: { dictionary: 'german' }, trigram: {} }
}
```

Downgrading to `2.3.2` fixes the bug. There is a similar, though slightly different ticket here https://github.com/Casecommons/pg_search/issues/446

Write a test, which fails would be great!

mattwr18:
Just saw that they have a fix for the issue upgrading pg_search @phoeinx @tillprochaska 

Fixes #385

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -69,7 +69,7 @@ gem 'telegram-bot'
 
 # FIXME: version 2.3.3 introduced a bug, found https://github.com/Casecommons/pg_search/issues/446.
 # Add back pessimistic operator after resolved.
-gem 'pg_search', '2.3.2'
+gem 'pg_search'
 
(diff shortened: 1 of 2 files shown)


tillprochaska (approved):
Just checked out the branch and it works fine for me!

PR #432, opened 2020-10-17
Always show request stats

### LINKED ISSUE #406: Show request stats even without replies
We now also display a total number of recipients of a request. This metric is relevant no matter the number of users that have replied, i.e. one would like to know how many users have received a request even if nobody has replied yet. Therefore, it would be an improvement to display the stats above the empty state notice (as it is already the case for requests with replies).

Bildschirmfoto 2020-09-28 um 21 08 59

tillprochaska:
This allows editors to see the total number of recipients of a requests directly after they’ve sent it. (Close #406)

**Before:**
Screen Shot 2020-10-17 at 12 55 40

**After:**
Screen Shot 2020-10-17 at 12 55 03

### DIFF
--- a/app/views/requests/show.html.erb
+++ b/app/views/requests/show.html.erb
@@ -22,16 +22,16 @@
 <%= c 'box', style: :large do %>
   <%= c 'wrapper' do %>
 
-    <% if @message_groups.blank? %>
+    <%= c 'stack', space: :xlarge do %>
+      <%= c 'request_metrics', request: @request, style: :cards %>
 
-      <%= c 'empty_state' do %>
-        <%= I18n.t 'request.no_replies' %>
-      <% end %>
+      <% if @message_groups.blank? %>
 
-    <% else %>
+        <%= c 'empty_state' do %>
+          <%= I18n.t 'request.no_replies' %>
+        <% end %>
 
-      <%= c 'stack', space: :xlarge do %>
-        <%= c 'request_metrics', request: @request, style: :cards %>
+      <% else %>
 
         <%= c 'stack' do %>
           <%= c 'heading', tag: :h2 do %>
@@ -42,6 +42,7 @@
             <%= c 'chat_messages_group', user: user, messages: messages %>
           <% end %>
         <% end %>
+
       <% end %>
 
     <% end %>


mattwr18 (commented):
On the surface, this should be a simple `PR` review, but it took me a while to feel confident giving one. First, I needed to familiarize myself with this part of the site. Then, when I attempted to send a message to my bot, I realized that the setup has changed and now we use a `.env` file. 

All good, but there is quite a few components at play here, so I wanted to understand what they are for. 

My overall impression is that I was quite surprised to see that we are not using semantic HTML :shrug: I know there is nothing in this `PR` that adds new html, so I guess it's not a concern of this `PR`, but I would like to bring up this topic before we continue building new components.

The first issue is that `div`s don't have any semantic meaning and according to the [MDN docs on %div](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div):

> should be used only when no other semantic element (such as %article or %nav) is appropriate.

Instead of writing more about semantic HTML, I would prefer to get your response, and maybe we can create a discussion ticket where we can have this conversation and keep it for our future selves. 

The other issue I could see is one of `divitis`, which has some concerns associated with it. See http://www.apaddedcell.com/div-itis-what-it-and-how-avoid-it ... while this is a little old, and I don't agree with everything, like using ids to target elements for styling with CSS, I think it brings up some good points. There is also this well-written, very thorough blog that touches on the subject http://designingforperformance.com/optimizing-markup-and-styles/ ... Going through the DOM tree, there are some places we have 9 nested divs!!  Look at this screen shot :point_down: 
Screenshot from 2020-10-19 11-00-39

While wrapping everything in a `div` instead of first attempting to use a semantic tag, and/or seeing if we achieve the same results with CSS and less HTML, can seem to be a faster way to build a page, there are medium to long term downsides that should be considered.

tillprochaska:
Hi @mattwr18, thanks for the review. I agree with most of your comments. I’ve definitely been a bit lazy with semantic tags, and I agree we use semantic tags if they exist.

I’m not so sure about `divitis` in general. I try to keep components reusable, independent of context and make them solve a single problem. (One part of this is to have certain cosmetic-free layout tasks abstracted into separate components – here’s a [more detailed explanation](https://every-layout.dev/rudiments/composition/) what this is about.)

That’s why I prefer something like this …

```html
<div class="Stack">
    <div class="Card">Card 1</div>
    <div class="Card">Card 2</div>
</div>
```

… over something where e.g. spacing is applied to the `.Card` elements directly. After all, the wrapping `div` in the case does not hurt – it doesn’t add any semantic context, but it’s not like the HTML would have been more semantic if we applied the spacing to the subelements directly. All it costs is a few bytes of HTML markup (and HTML is relatively cheap to parse).

I should probably note that there are *some* cases where we (in most cases probably I) wrapped elements in a div because I’ve been lazy or haven’t really thought about it, although personally I think for all of the elements in the screenshot above, I would only eliminate the `.CardMetrics` element.

So to summarize:

* Yes, there are probably a few places where we use a generic tag over a semantic alternative.
* I don’t think we should always avoid additional elements if that increases composability.

tillprochaska:
@mattwr18 Oh, and if you’d like to, maybe let’s have a call to further discuss at some point. I also created an issue for this (#439).

mattwr18 (approved):
Ok, I'm happy to continue the conversation in https://github.com/tactilenews/100eyes/issues/439
As I said, this `PR` is not directly affecting my concerns

roschaefer (approved):
divitis

PR #435, opened 2020-10-18
build: remove webpack service

roschaefer:
@tillprochaska in development environment, the assets are rebuild when I
run `bin/rails webpacker:clobber` no need for an extra service,
apparently. What was the service for originally?

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -19,13 +19,8 @@ services:
 
   app:
     <<: *x-defaults
-    depends_on: [ webpack, db, mailserver ]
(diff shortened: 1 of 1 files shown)


tillprochaska:
@roschaefer: The `webpack` service runs the webpack dev server, which enables us to make use of incremental (i.e. faster) builders and automatic reloading when CSS or JS changes.

I noticed yesterday the setup is broken (not sure when that happened) and just created a PR that fixes this (#438).

roschaefer:
Well well, if we need to optimize the compilation speed with the extra webpack service, I'm going to close this.

PR #437, opened 2020-10-19
426 implement single use invite urls

mattwr18:


app/controllers/onboarding_controller.rb
```diff
@@ -3,11 +3,12 @@
 class OnboardingController < ApplicationController
   skip_before_action :authenticate
   before_action :verify_token
+  skip_before_action :verify_token, only: %i[create_invite_url]
```
tillprochaska, line 6:
```suggestion
  before_action :verify_token, except: :create_invite_url
```

app/controllers/onboarding_controller.rb
```diff
@@ -17,28 +18,43 @@ def create
...
 
     render :index
   end
 
   def success; end
 
+  def create_invite_url
+    payload = SecureRandom.base64(16)
+    jwt = JsonWebToken.encode(payload)
+    render json: { url: URI::HTTPS.build(path: '/onboarding', query: "jwt=#{jwt}") }
```
tillprochaska, line 34: When I run this locally, I get an URL like `https:/onboarding?jwt=abc`. I think I’d make use of Rails’ url helpers here which should also make sure that we use the correct protocol scheme and host name.

Something like this should work:

```ruby
onboarding_path(jwt: jwt)
```

config/routes.rb
```diff
@@ -9,6 +9,7 @@
   get '/onboarding', to: 'onboarding#index'
   post '/onboarding', to: 'onboarding#create'
   get '/onboarding/success', to: 'onboarding#success'
+  post '/onboarding/invite-url', to: 'onboarding#create_invite_url'
```
tillprochaska, line 12: I think it’s a Rails convention to use underscores instead of dashes. (I personally dislike that, but for the sake of consistency, I’d go with underscores… 😄)

app/controllers/onboarding_controller.rb
mattwr18: oh, yeah, I was pretty sure there was an `except`, but didn't find it in the docs...

app/controllers/onboarding_controller.rb
mattwr18: way simpler hehe

config/routes.rb
mattwr18: rails convention to use underscores in url's? that seems odd... I'd have to look better into that, but I believe underscores would be the anti-pattern. See https://www.woorank.com/en/blog/underscores-in-urls-why-are-they-not-recommended

tillprochaska: Yeah, I don’t like that either, but I think if you had a multi-word resource (e.g. `OnboardingInvite`) the url helpers and everything would create a path like `/onboarding_invite` if you don’t explicitly pass in the path with dashes.

I’m not 100% sure if that’s still the case though – will try it out.

tillprochaska: Also I’m not sure if Google is still unable to interpret underscores in URLs in 2020. Even if they weren’t, I’d try to stick with the Rails way as SEO isn’t a priority for our app.

But let me check if what I said about underscores being a Rails convention is true in the first place… 😄

mattwr18: Yeah, I would generally stick to the Rails way, but in this case I think the world-wide web is much bigger than Rails and it seems to be an anti-pattern to use underscores in urls, so I would prefer to shorten it to `/onboarding/invite` than to follow aimlessly the Rails way.

Curious what others think, and what you find out. Maybe, I am overstating the underscores in urls being an anti-pattern.

app/controllers/onboarding_controller.rb
```diff
@@ -46,10 +44,8 @@ def verify_token
     raise ActionController::BadRequest if invalidated_jti.exists?
 
     JsonWebToken.decode(jwt_param)
-  rescue ActionController::BadRequest
-    redirect_to onboarding_unauthorized_path
-  rescue JWT::DecodeError
-    redirect_to onboarding_unauthorized_path
+  rescue ActionController::BadRequest, JWT::DecodeError
+    render :unauthorized, status: :unauthorized
```
mattwr18, line 48: Nice! I like this! But I got confused cause the `onboarding#create_invite_url` is also a `post` request, so wouldn't your comment :point_up: be relevant?

mattwr18: Maybe it doesn't need to be a `post` request?

tillprochaska: You’re referring to my comment about always redirecting from a post to a get route to prevent resubmitting the post request? If so, I don’t think it applies to the `create_invite_url` action, as that’s not a user-facing route, but an API endpoint.

mattwr18: yeah, true... wasn't thinking about that.

config/routes.rb
tillprochaska: > so I would prefer to shorten it to /onboarding/invite than to follow aimlessly the Rails way.

Yes, that sounds like a good solution (maybe even better than `create-invite` in the first place…)

Sounds like a good solution!

> Maybe, I am overstating the underscores in urls being an anti-pattern.

Even though I’m not sure how important it is from a technical and/or UX perspective, I think it just looks ugly! 😄

mattwr18: Yeah, I don't think it's important from that perspective. And it's a post endpoint that isn't exposed for SEO, so even from that aspect it's not important, but I agree that it raises my eyebrows, in a similar way that you can use underscores in class names as well, but it's widely considered best practice to use hyphens.

mattwr18:
@tillprochaska I noticed that there is still a todo about making it easier to pass the stimulus controller to avoid wrapping a button in a div for the sole purpose of adding the stimulus controller. At least that is what I understood anyways. I really really like the idea of removing this wrapping div cause I feel like it is not necessary and inappropriate semantically, I also worry about `a11y` when wrapping everything in `div`s as well.

Unfortunately, I am dealing with issues getting my screenreader to work in the browser, so it's difficult for me to give a meaningful audit, but it is generally think that avoiding unnecessary tags is a good practice to get into. 

I want to update the ticket you opened https://github.com/tactilenews/100eyes/issues/439 cause I feel like it would be good for us to be on the same page in regards to these issues.

That being said, I would be interested in helping out with this, my thought was to allow any component to accept a `stimulus_controller`, maybe in `application_component.rb`, so that it could be applied directly to the root element of the component receiving it, but maybe you've thought of a way you prefer to handle this case?

tillprochaska:
Yes, that’s exactly what I started working on!

mattwr18:
> Yes, that’s exactly what I started working on! (Actually I worked on adding any data attribute to a component’s root element.)

awesome! I guess I'll leave you to it since you have already started and know the infrastructure much better than I do, but just let me know if you want some help with anything

tillprochaska:
> awesome! I guess I'll leave you to it since you have already started and know the infrastructure much better than I do, but just let me know if you want some help with anything

I’m planning to finish that today or tomorrow (would like to add some tests and do some clean up). Would definitely be cool if you could have a look at it then. Maybe there are use cases for smth I didn’t consider etc.

tillprochaska:
> I want to update the ticket you opened #439 cause I feel like it would be good for us to be on the same page in regards to these issues.

Yes, sure, that’d be great! I just quickly opened it to make sure we don’t forget about the issue.

tillprochaska:
Hi @mattwr18, I [pushed](https://github.com/tactilenews/100eyes/pull/437/commits/8d2c7fe35e19765d8f0ebedb2375fa7e14b6f709) a few changes that make it easier to add data attributes to component root elements.

This is how you’d use them:

When rendering the component from the component class:

```rb
content_tag(:button, 'Button label', data: data)
```

When rendering the component using a template:

```html
<button <%= data_string %>>Button label</button>
```

In both cases, the component can then be used like this:

```rb
c 'button', data: { key: 'value' }
```

… which would render the following:

```html
<button data-key="value">Button label</button>
```

Here’s a few things I was wondering:

* Should I add this as a separate PR? We could then also refactor other existing components that might profit from these changes. (That would also mean we’d wait for that PR to be merged or merge this without the refactoring.)
* Is `data_string` a good name for the helper method? I think having `string` in the name is nice to make it clear this is already stingified and not a hash or something similar. On the other hand, something like `data_attrs` might make it more clear that this is about HTML data attributes. Also, for context, we already have a `class_attr` helper (which returns only one value though).
* With Stimulus, elements can have multiple controllers/actions/targets, for example `data-action="input->doSomething blur->doSomethingElse` registers event handlers for the `input` and `blur` events. With the current implementation, you’d have to set all actions at once. This might be relevant in the following use case:

   The `FancyButton` component template:
   ```html
   <button class="FancyButton" data-action="click->fancy-button#showConfetti" <%= data_string %>>Button label</button>
   ```

   The `LikeButton` component template:
   ```html
   <%= c 'fancy_button', data: { action: 'click->like-button#likeTweet } %>
   ```

   The `LikeButton` component would render the following HTML:
   ```html
   <button class="FancyButton" data-action="click->fancy-button#showConfetti" data-action="click->like-button#likeTweet>Button label</button>
   ```

   This obviously doesn’t work. I’m not sure if we should handle this right now though, as I don’t think we every had that use case.

tillprochaska:
@mattwr18: Okay, reading my comment above again, maybe we should really have that as a separate PR. Also to make sure the other devs can have a look at it.

app/components/create_invite_button/create_invite_button.html.erb
```diff
@@ -0,0 +1,22 @@
+<div
+  class="CreateInviteButton"
+  data-controller="create-invite-button"
+  data-action="click->create-invite-button#copyLink"
+>
```
tillprochaska, line 5: Forgot to remove that tag…

mattwr18:
I like the idea of having a separate `PR` to refactor as well, but it would be nice not to block this from being merged in. Maybe, it could be reviewed on its own with the wrapping div, then it could be refactored as well in the other `PR`?

I like the idea of naming it `data_attrs`, and I also like your idea of extending it as well. I think there will be times we could pass in `aria_attrs` as well for `a11y` concerns.

tillprochaska:
> Maybe, it could be reviewed on its own with the wrapping div, then it could be refactored as well in the other PR?

Yes, we should do that!

tillprochaska:
@mattwr18: I removed my last commit and created #451 and a draft PR #452 where I’m also collecting components that should be refactored.

tillprochaska:
@roschaefer @phoeinx Do you think one of you could review this today or tomorrow?

(No pressure, if you don’t have time, that’s no problem. I just wanted to check with you so we can plan the #Schweinesystem onboarding accordingly. Just let me know!)

app/controllers/onboarding_controller.rb
```diff
@@ -17,28 +17,42 @@ def create
...
   private
 
   def redirect_to_success
-    redirect_to onboarding_success_path(token: token_param)
+    redirect_to onboarding_success_path(jwt: jwt_param)
   end
 
   def verify_token
-    raise ActionController::BadRequest unless token_param == ENV['ONBOARDING_TOKEN']
+    invalidated_jti = JsonWebToken.where(invalidated_jti: jwt_param)
```
tillprochaska, line 43: @mattwr18 Why is the column called `invalidated_jti`? From what I can see, it stores the complete token.

app/controllers/onboarding_controller.rb
```diff
@@ -17,28 +17,42 @@ def create
     return redirect_to_success if User.email_taken?(user_params[:email])
```
tillprochaska, line 17: We need to invalidate the token in this case as well. Otherwise the following scenario would be possible:

* User A has already registered for 100eyes using email.
* User B hasn’t yet registered and has a valid invite link and knows the email address of user A.
* User B enters user A’s email address into the onboarding form.
* User B is redirected to the success page, but the token is still valid.
* User B knows that user A is a 100eyes contributor.

This is a potential privacy flaw.

app/components/create_invite_button/create_invite_button.html.erb
tillprochaska, line 5: We’re going to remove the wrapping `div` in #452.

app/controllers/onboarding_controller.rb
```diff
@@ -17,28 +17,42 @@ def create
...
+    payload = SecureRandom.base64(16)
+    jwt = JsonWebToken.encode(payload)
+    render json: { url: onboarding_url(jwt: jwt) }
+  end
+
   private
 
   def redirect_to_success
-    redirect_to onboarding_success_path(token: token_param)
+    redirect_to onboarding_success_path(jwt: jwt_param)
```
tillprochaska, line 39: This doesn’t actually work. When the user is redirected to the success page, the token is already invalidated. I’d suggest to simply allow visiting the `/onboarding/success` page without a token as it doesn’t contain any sensitive information.

app/controllers/onboarding_controller.rb
mattwr18: yeah, that's true, I think it doesn't make a big difference either way, but it's nice to keep our semantics more accurate.

app/controllers/onboarding_controller.rb
mattwr18: I agree... maybe I should have asked, but I wasn't sure why we had the token there in the first place.

app/components/onboarding_email_form/onboarding_email_form.html.erb
```diff
@@ -18,7 +18,7 @@
       <%= c 'input', field.input_defaults.merge(placeholder: '') %>
     <% end %>
 
-    <input type="hidden" id="token" name="token" value="<%= token %>" />
+    <input type="hidden" id="token" name="token" value="<%= jwt %>" />
```
tillprochaska, line 21: I think the `id` here would have to be `jwt` (instead of `token`) as well, right? Because you expect the token to be passed via the `jwt` param…

But actually, maybe removing the hidden input and just adding the JWT to the form action URL as a query param would be way easier. In the `form` tag helper, we could just update the URL to `onboarding_path(jwt: jwt)`.

app/controllers/onboarding_controller.rb
mattwr18: yeah, that's a good point

app/components/onboarding_email_form/onboarding_email_form.html.erb
mattwr18: yep, good call.

tillprochaska (approved):
@mattwr18: I’m happy with it right now – fell free to request additional reviews by the other devs or merge the PR directly.

app/views/onboarding/success.html.erb
```diff
@@ -1,4 +1,5 @@
-<%= c 'onboarding_success',
-  title: Setting.onboarding_success_title,
-  text: Setting.onboarding_success_text
+<%= c 'onboarding_response',
+  style: :success,
+  heading: I18n.t('components.onboarding_success.heading', project_name: Setting.project_name),
+  text: I18n.t('components.onboarding_success.text').html_safe
 %>
```
tillprochaska, line 5: I think this is a regression, as editors won’t be able to update the UI texts for the onboarding page using the new settings UI. 

```suggestion
<%= c 'onboarding_response',
  style: :success,
  heading: Setting.onboarding_success_title,
  text: ISetting.onboarding_success_text
%>
```

app/views/onboarding/unauthorized.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= c 'onboarding_response',
+  style: :error,
+  heading: I18n.t('components.onboarding_unauthorized.heading'),
+  text: I18n.t('components.onboarding_unauthorized.text').html_safe
+%>
```
tillprochaska, line 5: @mattwr18: We would need to make this configurable using the settings UI as well. If you want to, I can add the necessary fields.

mattwr18: done

app/views/onboarding/success.html.erb
mattwr18: good catch

tillprochaska:
Hi @mattwr18, sorry – I found another thing we’d need to change before this can be merged. Nothing critical, but we should remove the setting for the onboarding token:

* https://github.com/tactilenews/100eyes/blob/426-implement-single-use-invite-urls/app/components/settings_form/settings_form.html.erb#L19-L21
* https://github.com/tactilenews/100eyes/blob/426-implement-single-use-invite-urls/app/controllers/settings_controller.rb#L20
* https://github.com/tactilenews/100eyes/blob/426-implement-single-use-invite-urls/app/models/setting.rb#L9
* https://github.com/tactilenews/100eyes/blob/426-implement-single-use-invite-urls/config/locales/de.yml#L123-L124

mattwr18:
No reason to be sorry @tillprochaska ... it's good you are being thorough... I also removed the onboarding token from the ansible setup @roschaefer since they can only be used once.

PR #438, opened 2020-10-19
Fix webpack dev server

tillprochaska:
Currently, the development setup is broken and you can’t make use of Webpack’s dev server which provides incremental, fast builds and automatic reloads.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,16 +1,14 @@
 version: '3.7'
 
-x-default-env: &x-default-env
-  RAILS_ENV: "${RAILS_ENV:-development}"
-  NODE_ENV: development
-  MAILSERVER_HOST: mailserver
-  MAILSERVER_PORT: 25
-
 x-defaults: &x-defaults
   build:
     context: .
   environment:
-    <<: *x-default-env
+    RAILS_ENV: "${RAILS_ENV:-development}"
+    NODE_ENV: development
+    MAILSERVER_HOST: mailserver
+    MAILSERVER_PORT: 25
+    WEBPACKER_DEV_SERVER_HOST: 0.0.0.0
   stdin_open: true
   tty: true
   volumes:
@@ -22,21 +20,10 @@ services:
 
   app:
     <<: *x-defaults
-    depends_on: [ webpack, db, mailserver ]
-    environment:
-      <<: *x-default-env
-      WEBPACKER_DEV_SERVER_HOST: webpack
-    ports: ['3000:3000']
-
-  webpack:
-    <<: *x-defaults
-    command: ./bin/webpack-dev-server
-    ports: [ '3035:3035' ]
-    environment:
-      <<: *x-default-env
-      WEBPACKER_DEV_SERVER_HOST: 0.0.0.0
-    volumes:
-      - .:/app
+    depends_on: [ db, mailserver ]
+    ports:
+      - 3000:3000
+      - 3035:3035
 
   db:
     environment:
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,13 +1,16 @@
 version: '3.7'
 
+x-default-env: &x-default-env
+  RAILS_ENV: "${RAILS_ENV:-development}"
+  NODE_ENV: development
+  MAILSERVER_HOST: mailserver
+  MAILSERVER_PORT: 25
+
 x-defaults: &x-defaults
   build:
     context: .
   environment:
-    RAILS_ENV: "${RAILS_ENV:-development}"
-    NODE_ENV: development
-    MAILSERVER_HOST: mailserver
-    MAILSERVER_PORT: 25
+    <<: *x-default-env
   stdin_open: true
   tty: true
   volumes:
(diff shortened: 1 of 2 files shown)


docker-compose.override.yml
```diff
@@ -1,13 +1,16 @@
 version: '3.7'
 
+x-default-env: &x-default-env
+  RAILS_ENV: "${RAILS_ENV:-development}"
+  NODE_ENV: development
+  MAILSERVER_HOST: mailserver
+  MAILSERVER_PORT: 25
```
tillprochaska, line 7: @roschaefer I’m unsure if this is the best way to extend the environment in services. Maybe there is a Docker-native solution that does not rely on YAML anchors.

docker-compose.override.yml
```diff
@@ -1,13 +1,16 @@
...
+
 x-defaults: &x-defaults
   build:
     context: .
   environment:
-    RAILS_ENV: "${RAILS_ENV:-development}"
-    NODE_ENV: development
-    MAILSERVER_HOST: mailserver
-    MAILSERVER_PORT: 25
+    <<: *x-default-env
```
roschaefer, line 13: Why do we have `x-default-env` which get imported into `x-defaults` and both are being imported in the services below?

mind blown

And why do we need `./bin/webpack-dev-server` as a separate service? I removed it in #435 because I could see the assets being compiled by `bin/rails server` already. It takes like a couple of seconds, is that too slow?

roschaefer (changes requested):

docker-compose.override.yml
roschaefer: @tillprochaska you can debug these configuration files with [docker-compose config](https://docs.docker.com/compose/reference/config/)

tillprochaska: > Why do we have x-default-env which get imported into x-defaults and both are being imported in the services below?

That’s what why I asked above if there’s a better solution… :D

tillprochaska: > It takes like a couple of seconds, is that too slow.

Yes, definitely.

docker-compose.override.yml
tillprochaska: We could have a look at this, but I’m not sure if it’ll work with the rest of the setup: https://docs.docker.com/compose/extends/#understand-the-extends-configuration

docker-compose.override.yml
roschaefer: I am able to run `docker-compose exec app bin/webpack-dev-server`, change the CSS and see the effect immediately. Can we close this @tillprochaska? Shall I reopen #435 ?

tillprochaska: Robert, this has been a blocker during development for me and others for the past few days now. I’d really appreciate if we could try to resolve this so we can get work done rather than arguing again and again that we should remove the Webpack dev server.

This is how live reloading works: Webpack dev server starts a small websockets server to notify clients about recompiled/updated assets. Webpack injects a small snippet to JavaScript bundles that opens a connection to that server and listens for changes, then reloads CSS or the whole page.

> I am able to run docker-compose exec app bin/webpack-dev-server, change the CSS and see the effect immediately.

For that to work, the correct web sockets URL needs to be injected into the JS bundles. If you tried the above command, that’ likely not gonna work as you’re starting the Webpack dev server in a different container than it is configured.

When you simply start the containers using `docker-compose up app`, it works fine for me (see [this screencast](https://www.dropbox.com/s/95v3ljrp3aq99mg/Screen%20Recording%202020-10-21%20at%2008.43.17.mov?dl=0)).

roschaefer: Wait, I'm here to help, not to argue :slightly_smiling_face: 

The command `docker-compose exec` uses the *same container*! That's the difference to `docker-compose run`. See here: https://docs.docker.com/engine/reference/commandline/exec/

To verify, it changed CSS on branch `415_configurable_settings` on commit af189ddf553da6072ab4a266302435062e2b8881. *While the containers are running* after `docker-compose up` and can see the delay of ~4 seconds. When I run `docker-compose exec app bin/webpack-dev-server` I see no delay. When I stop the additional command, the delay is there again. And so on. For me, this seems to work.

Could you please verify that this solution doesn't work? Please @tillprochaska?

roschaefer: Ah, one thing I noticed is the webpack hot-reload does not work, as port 3035 is not shared.

roschaefer: Now also hot-reloading works: f5af1900f6f6909e655a5104d11d721123ca8f15

tillprochaska: Sorry about my comment above, I think I misunderstood your comment. Thought you were complaining it’s *not* working with the command you mentioned. ❤️ Should have had a coffee before writing angry comments… 😴

I think on the master branch it should work in my machine as well. Can’t verify that until this afternoon though as I won’t be at my desk.

I still don’t really like that solution and don’t understand what’s the problem with including the webpack service in the development config. After all it’s about making it as easy as possible to set up a development environment and we also include e.g. a dev mailserver in there for convenience.

If I could start my dev env using a single ‘dcu’ command instead of having to open two separate shell sessions and starting two commands in there, I’d very much prefer the first option.

If you have strong feelings about removing the web pack service, go for it, but please make sure to include updated instructions for the dev set up in the read me and ping the other devs on the team about this change so we don’t waste more time figuring out why our dev setup behaves differently than it did before.

roschaefer: My primary motivation is to keep the `docker-compose.yml` maintanable.

Apart from that, it could also be a performance improvement not to start the webpack-dev-server unless needed. In contrast to `maildev` we don't need the `webpack-dev-server` for the application to work, so I think it's justified to start `webpack-dev-server` on-demand.

roschaefer: ping @mattwr18 @phoeinx see a3987ab76b2b91a53e0ac0338a4b308c8b8501af

roschaefer (approved):

PR #440, opened 2020-10-19
Render telegram widget, verify the has if from telegram

mattwr18:
For this to work locally, you must register you local bot in our @BotFather with:

>/setdomain
Choose a bot to set or change its linked domain. See Login Widget documentation for details:
https://core.telegram.org/widgets/login
@HundredEyesMattBot
Link your website with your bot to use the Telegram Login Widget. Widget documentation:
https://core.telegram.org/widgets/login
Use /empty to remove.
https://odd-shrimp-41.loca.lt

Note: this only works over https, so you can set up a localtunnel with
`npx localtunnel --port 3000`

You must start it and then add the host to your `development.rb` config `  config.hosts << 'odd-shrimp-41.loca.lt'
` and restart the server.

For some reason, loading the button through the CDN breaks the styling of the page... maybe you have an idea why @tillprochaska 
See screenshot
Screenshot from 2020-10-19 19-52-37

To this point, it is redirecting to `onboarding#telegram_auth`, which is verifying the has comes from `Telegram` and if so finding or creating the user and responding with a success. Maybe this should be tying into our decorator, since I've needed to ignore rubocop complaining to push up and it would be cleaner code.


tillprochaska:
> For some reason, loading the button through the CDN breaks the styling of the page... maybe you have an idea why @tillprochaska

Will have a look!

tillprochaska:
Just one thing I already noticed from a UX perspective: I think I’d keep the list of steps for Telegram as it is right now (only changing the contents). Explaining what Telegram is and how to install the app might still be useful:

I think with #50survivors, it has happened quite often that contributors wanted to sign up using a chat app, but had never used/installed Telegram before. (Maybe @FrauCsu has more insights…)

tillprochaska:
> Note: this only works over https

I was wondering if using the authorization callback instead of letting Telegram redirect might make the integration easier:

* Would we still need to expose/tunnel the app during development?
* Would we still need to configure a domain for the Telegram bot? (If not, this would reduce complexity when setting up a new app instance and bot for a new project.)

We could listen for the authorization callback and then manually redirect to the respective backend route. So backend-wise, nothing should change, only like a few lines of JavaScript would be required to handle the callback and redirect.

***

Screen Shot 2020-10-20 at 12 06 01

mattwr18:
From what I saw @tillprochaska , we still would, but I could give it a go, maybe when I tried it, it wasn't working for another rrason

FrauCsu:
> Just one thing I already noticed from a UX perspective: I think I’d keep the list of steps for Telegram as it is right now (only changing the contents). Explaining what Telegram is and how to install the app might still be useful:
> 
> I think with #50survivors, it has happened quite often that contributors wanted to sign up using a chat app, but had never used/installed Telegram before. (Maybe @FrauCsu has more insights…)

Yes, that happened. But as soon as we explained "like Whatsapp" quite a few contributors were okay with installing Telegram.

app/controllers/onboarding_controller.rb
```diff
@@ -24,6 +27,31 @@ def create
...
+    raise ActionController::BadRequest if Time.now.to_i - telegram_auth_params[:auth_date].to_i > 86_400
+
+    @user = User.find_or_create_by(
+      telegram_id: telegram_auth_params[:id],
+      first_name: telegram_auth_params[:first_name],
+      last_name: telegram_auth_params[:last_name]
+    )
+    if @user.save!
+      render json: { message: 'Success' }, status: :ok
+    else
```
roschaefer, line 49: How could we reach the `else` branch here?

config/environments/development.rb
```diff
@@ -64,4 +64,6 @@
   }
 
   config.project_name = ENV['HUNDRED_EYES_PROJECT_NAME'] || '100eyes'
+  # FIXME: this is needed to use localtunnel, the telegram widget only works over https
+  config.hosts << 'odd-shrimp-41.loca.lt'
```
roschaefer, line 68: I have `APPLICATION_HOSTNAME` in #445 for that purpose, you can overwrite it with an environment variable.

roschaefer: With [localtunnel](https://www.npmjs.com/package/localtunnel) you can request particular subdomains, so we could also register a dedicated domain here for development purposes.

app/controllers/onboarding_controller.rb
```diff
@@ -24,6 +27,31 @@ def create
...
   def success; end
 
+  # rubocop:disable Metrics/AbcSize
+  def telegram_auth
+    check_hash = telegram_auth_params[:hash]
+    auth_data = telegram_auth_params.slice(:auth_date, :first_name, :id, :last_name)
+    check_string = auth_data.to_unsafe_h.map { |k, v| "#{k}=#{v}" }.sort.join("\n")
+
+    secret_key = OpenSSL::Digest::SHA256.new.digest(ENV['TELEGRAM_BOT_API_KEY'])
+    hash = OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new('sha256'), secret_key, check_string)
```
roschaefer, line 37: Does it work similar like JWT? Would it be ok if we write a test for that method and throw in an example token?

app/controllers/onboarding_controller.rb
```diff
@@ -24,6 +27,31 @@ def create
...
+
+    raise ActionController::BadRequest if hash.casecmp(check_hash) != 0
+    raise ActionController::BadRequest if Time.now.to_i - telegram_auth_params[:auth_date].to_i > 86_400
+
+    @user = User.find_or_create_by(
+      telegram_id: telegram_auth_params[:id],
```
roschaefer, line 43: When we have the telegram id through login, are we allowed to proactively send a message to the user?

app/components/telegram_widget/telegram_widget.html.erb
```diff
@@ -0,0 +1,8 @@
+<script 
+  async 
+  src="https://telegram.org/js/telegram-widget.js?12" 
+  data-telegram-login="HundredEyesMattBot"
```
roschaefer, line 4: at least:
```suggestion
  data-telegram-login="<%= ENV['TELEGRAM_BOT_USERNAME'] %>" 
```

after #434:
```suggestion
  data-telegram-login="<%= Setting.telegram_bot_username %>" 
```

app/controllers/onboarding_controller.rb
```diff
@@ -24,6 +27,31 @@ def create
...
+  def telegram_auth
+    check_hash = telegram_auth_params[:hash]
+    auth_data = telegram_auth_params.slice(:auth_date, :first_name, :id, :last_name)
+    check_string = auth_data.to_unsafe_h.map { |k, v| "#{k}=#{v}" }.sort.join("\n")
+
+    secret_key = OpenSSL::Digest::SHA256.new.digest(ENV['TELEGRAM_BOT_API_KEY'])
+    hash = OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new('sha256'), secret_key, check_string)
+
+    raise ActionController::BadRequest if hash.casecmp(check_hash) != 0
+    raise ActionController::BadRequest if Time.now.to_i - telegram_auth_params[:auth_date].to_i > 86_400
```
roschaefer, line 40:
```suggestion
    raise ActionController::BadRequest if Time.now.to_i - telegram_auth_params[:auth_date].to_i > 24 * 60 * 60
```
or even better
```suggestion
    raise ActionController::BadRequest if telegram_auth_params[:auth_date] + 1.day > Time.current
```

roschaefer (changes requested):
testing

@mattwr18 looks good and simple enough :+1: just needs some testing and refactoring in my eyes. I didn't try it out yet, but will do tomorrow.

mattwr18:
hey thanks for the early review @roschaefer , but I guess I'm far from done... I'll take into consideration your comments, but this was as far as I could get at the end of the day and just wanted to push it up, maybe I should have added a `[WIP]` to the title, or a label? 
I don't know the culture on the team, I have been in the habit of just requesting a review when I'm ready, but let me know what you prefer.

app/controllers/onboarding_controller.rb
mattwr18: > > Note: this only works over https
> 
> I was wondering if using the authorization callback instead of letting Telegram redirect might make the integration easier:
> 
>     * Would we still need to expose/tunnel the app during development? (If not, I’d prefer that as we wouldn’t need to change the dev setup.)
> 
>     * Would we still need to configure a domain for the Telegram bot? (If not, this would reduce complexity when setting up a new app instance and bot for a new project.)
> 
> 
> We could listen for the authorization callback and then manually redirect to the respective backend route. So backend-wise, nothing should change, only like a few lines of JavaScript would be required to handle the callback and redirect.
> 
> Screen Shot 2020-10-20 at 12 06 01

So, @tillprochaska having a look at this, we would still need to configure the telegram to `/setdomain`, if you look at the instructions on the [Linking your domain to the bot](https://core.telegram.org/widgets/login#linking-your-domain-to-the-bot), you will see that these instructions

> Once you have chosen a bot, send the /setdomain command to @Botfather to link your website's domain to the bot. Then configure your widget below and embed the code on your website.

are not specifically for the redirect method. I checked it out for the callback method as well, and it only works when you have configured an `https` domain. We could still use the callback method, if we wanted to handle the [Checking authorization](https://core.telegram.org/widgets/login#linking-your-domain-to-the-bot), but I think we might not want to do that in JS because I'm not sure how we can organize this logic into a stimulus controller, we would need to use inline JS script tags, or maybe you know how else we could do it?

Oh, re-reading your comment, you were just suggesting using the JS to redirect to the rails endpoint, as a workaround to the domain registration, right? I think that's not possible, and we have a way to redirect provided by telegram, so it makes sense to just favor that.

mattwr18:
> > Note: this only works over https
> 
> I was wondering if using the authorization callback instead of letting Telegram redirect might make the integration easier:
> 
>     * Would we still need to expose/tunnel the app during development? (If not, I’d prefer that as we wouldn’t need to change the dev setup.)
> 
>     * Would we still need to configure a domain for the Telegram bot? (If not, this would reduce complexity when setting up a new app instance and bot for a new project.)
> 
> 
> We could listen for the authorization callback and then manually redirect to the respective backend route. So backend-wise, nothing should change, only like a few lines of JavaScript would be required to handle the callback and redirect.
> 
> Screen Shot 2020-10-20 at 12 06 01

So, @tillprochaska having a look at this, we would still need to configure the telegram to `/setdomain`, if you look at the instructions on the [Linking your domain to the bot](https://core.telegram.org/widgets/login#linking-your-domain-to-the-bot), you will see that these instructions

> Once you have chosen a bot, send the /setdomain command to @Botfather to link your website's domain to the bot. Then configure your widget below and embed the code on your website.

are not specifically for the redirect method. I checked it out for the callback method as well, and it only works when you have configured an `https` domain. We could still use the callback method, if we wanted to handle the [Checking authorization](https://core.telegram.org/widgets/login#linking-your-domain-to-the-bot), but I think we might not want to do that in JS because I'm not sure how we can organize this logic into a stimulus controller, we would need to use inline JS script tags, or maybe you know how else we could do it?

Oh, re-reading your comment, you were just suggesting using the JS to redirect to the rails endpoint, as a workaround to the domain registration, right? I think that's not possible, and we have a way to redirect provided by telegram, so it makes sense to just favor that.

app/controllers/onboarding_controller.rb
mattwr18: yes, we can ask for this permission with the `data-request-access="write">` part of the script.

app/controllers/onboarding_controller.rb
mattwr18: I mean I haven't thought this through too much cause I am still working on the specifics, but I can imagine if the user is saved we render the success page and if they are not saved, we could render the error page to indicate they might need to get in contact with us?

I'm not sure if that answers your question, or if you are trying to imply something? The thing is that this feature is actually a login feature that we are trying to re-purpose as a authorize me to send you messages and create an account for you in our database, right? So, I think we'd need to use `find_or_create_by`, since the user could already exist? Not sure about this though. Ideally, when this is deployed there will be single use invite urls so that once a user has clicked on the link, they would not be able to access the page again, but would it be possible that a user already exists in our database when they use the link for the first time?

Or maybe you were just pointing out that by using `save!` instead of `save`, it would throw an exception if there is an issue with save and would then never reach the else clause? 

Are you being cryptic on purpose @roschaefer :stuck_out_tongue:

roschaefer: > Or maybe you were just pointing out that by using `save!` instead of `save`, it would throw an exception if there is an issue with save and would then never reach the else clause?

Yes

> Are you being cryptic on purpose @roschaefer

No, I think I was assuming that `find_or_create_by` would always return a valid user record, so `@user.save` can never get `false`.

roschaefer:
> hey thanks for the early review @roschaefer , but I guess I'm far from done... I'll take into consideration your comments, but this was as far as I could get at the end of the day and just wanted to push it up, maybe I should have added a `[WIP]` to the title, or a label?
> I don't know the culture on the team, I have been in the habit of just requesting a review when I'm ready, but let me know what you prefer.

I just like to reduce the number of open PRs :slightly_smiling_face:

tillprochaska:
> Oh, re-reading your comment, you were just suggesting using the JS to redirect to the rails endpoint, as a workaround to the domain registration, right? I think that's not possible, and we have a way to redirect provided by telegram, so it makes sense to just favor that.

Yeah, that’s what I was thinking. But reading it again, even if that way we could’ve circumvented the domain registration thing, it’d have been quite hacky and setting up the tunnel is actually quite easy. (Would be still good to mention that in the readme as well, though, I think.)

config/locales/de.yml
```diff
@@ -78,6 +78,7 @@ de:
         intro: |
           Um per Telegram teilzunehmen, folgen Sie bitte diesen vier Schritten:
         steps:
+          - Wenn Sie bereits Telegramm haben, können Sie sich anmelden, indem Sie auf die Schaltfläche unten klicken, um uns zu autorisieren, Ihnen Nachrichten zu senden.
```
mattwr18, line 81: please @FrauCsu can you give me what you want this step to say actually? This is a placeholder.

app/controllers/onboarding_controller.rb
mattwr18: true statement @roschaefer

mattwr18:
This is what the UI looks like
Screenshot from 2020-10-27 12-52-25

The button is only visible after a user chooses the Telegram option. I think this takes care of the concern you had @tillprochaska where some might not have Telegram, but be willing to install it. The would still have the instructions on how to manually find it, but users who have Telegram would have a minimal click solution to create an account and give us permissions to send them messages.

app/components/telegram_widget/telegram_widget.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="TelegramWidget">
+  <script
+    async
+    src="https://telegram.org/js/telegram-widget.js?12"
+    data-telegram-login="<%= ENV['TELEGRAM_BOT_USERNAME'] %>"
+    data-size="large"
+    data-auth-url="https://<%= ENV['APPLICATION_HOSTNAME'] %>/onboarding/telegram-auth"
+    data-request-access="write">
+  </script>
+</div>
```
mattwr18, line 10: I considered not creating a component for this and just rendering this in the onboarding instructions template, but decided that we might want to use it on other pages if we support logging in and letting users manage settings, or something similar.

app/components/telegram_widget/telegram_widget.css
```diff
@@ -0,0 +1,3 @@
+.TelegramWidget {
+  margin-top: 2rem;
+}
```
tillprochaska, line 3: This is an example for what I tried to explain when we spoke about presentational components on Tuesday and trying to make components context-independent.

The goal in this case is that there is some whitespace between the Telegram widget and the list that is rendered before.

So the context of this is actually a parent element of the widget and a list -- the white space should be between the two of them.

By adding the whitespace to the `TelegramWidget` component, the component becomes dependent on its context. Also, anytime you change the spacing between elements in the `OnboardingInstructions` component, you also need to adjust them in the `TelegramWidget` component, which is an indicator for tight coupling.

Now this might not be a huge problem in this case, and it probably won’t become a problem in the future, as the `TelegramWidget` component is likely going to be a one-off component, so the possible implications I mentioned above might be a bit constructed in this case.

But in general, I think it increases maintainability to reduce those kinds of side effects.

tillprochaska: In this case, you could actually leverage the `Stack` component that is already being used to add the spacing between the bold text and the list. A stack adds whitespace between its child elements.

Screen Shot 2020-10-31 at 12 52 11

tillprochaska:
> The button is only visible after a user chooses the Telegram option. I think this takes care of the concern you had @tillprochaska where some might not have Telegram, but be willing to install it. The would still have the instructions on how to manually find it, but users who have Telegram would have a minimal click solution to create an account and give us permissions to send them messages.

Sorry, I think I didn’t really think this through and have been wrong here. These are the reasons why we need/want to have the Telegram Widget:

1. We want to restrict Telegram sign-ups to contributors who have a valid invite link any more.
2. Many Telegram users haven’t added any name, only their first name or nicknames etc. to their Telegram profiles. During the onboarding process, we want to ask contributors for their first and last name, and use them instead of the data they provide in their profile.
3. We want to make it easier for contributors to sign up using Telegram, in order to increase conversions and reduce the amount of situations where editors need to assist contributors with sign up.

In order to achieve 1) and 2), we need to switch completely to sign-ups using the Telegram Widget, i.e. **contributors shouldn’t be able to sign up by starting the bot manually**. (If they do, no new contributor accounts should be created.)

We also shouldn’t mix two different ways to sign up with Telegram. We should still add some info where to get the Telegram app though, just to make sure contributors who haven’t been Telegram users so far have all the info they need. But I’m not sure we actually need a list for that.

In order to ask contributors for their name, I think there are two options:

1) Ask them before we send them over to Telegram, similar to the email form:

Screen Shot 2020-10-31 at 13 34 58

2) Ask them to complete or update their data after they’ve authorized us. The onboarding would then only consist of the Telegram widget.

Screen Shot 2020-10-31 at 13 35 30

app/components/telegram_widget/telegram_widget.css
mattwr18: I understand your logic, and I agree that if we decide to reuse, we might need to update the styling so it can be used in different contexts. I disagree that adding html elements to add this styling is a good direction.

The way I would prefer to solve this is to target the `TelegramWidget` in CSS in a way that is namespaced, so that it would only affect the styling in that one instance. I have added a commit https://github.com/tactilenews/100eyes/pull/440/commits/db962391c922e89b0de681e64bb40701399ace07 for that.

mattwr18:
Fair enough @tillprochaska ... yes, it totally makes sense that if we will require single invite urls that we should not allow other forms of signing up.

I think it makes sense to continue this only after the other `PR` gets merged.

tillprochaska:
> I think it makes sense to continue this only after the other PR gets merged.

Yes, you’re right, for validating the invites, we’d have to wait for the PR being merged. You could already have a look at the part of the logic that creates a new telegram user when you send the `/start` command, as we’d have to remove that as well.

I’m not super familiar with that, but it’s in the `TelegramMessage` decorator. If I remember correctly, this is what it does:

* Anytime we get a new message (including the message sent by the `/start` command), check if we already have a contributor with the given Telegram user id in our database.
* If a contributor account exists, update the record with data from the Telegram profile.
* If a contributor accounts doesn’t yet exist, create it with the data from the Telegram profile.

https://github.com/tactilenews/100eyes/blob/db962391c922e89b0de681e64bb40701399ace07/app/decorators/telegram_message.rb#L45-L51

mattwr18:
hey @roschaefer could you give me early feedback on https://github.com/tactilenews/100eyes/pull/440/commits/7f3709c0cf0059443c0074e158a803aaa3cd9d66 and https://github.com/tactilenews/100eyes/pull/440/commits/04ce00ea31f3c00db65207a8a3bb373be0ea729bjust to see that I'm on the right track... Obviously, we wouldn't actually respond with `Who are you?`, or would we?

Am I missing something? There are some failing tests I'd need to have a look at, but I think it's more or less what we need to do to disable auto-signing up through telegram, or?

PR #441, opened 2020-10-19
Use classic `FileUpdateChecker` instead of watcher-based implementation

tillprochaska:
There seems to be a bug in Docker for Mac (and probably also for Windows) which prevents file system events to propagate into containers. As a result, Rails doesn’t automatically reload changed source files in the development environment.

It’s a [known issue](https://github.com/docker/for-mac/issues/4962) and a fix should land in the near future. For the meantime, it’d be nice if we could use the classic `FileSystemChecker` which doesn’t rely on file system events. I’ve subscribed to updates on the issues mentioned and will revert that change once the bug has been fixed.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -56,7 +56,7 @@ Rails.application.configure do
 
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.
-  config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+  config.file_watcher = ActiveSupport::FileUpdateChecker
 
   # CUSTOM
   config.mailer = {


kylepeeler:
I was also noticing my file watcher for React (Hot Module Reloading) was not picking up changes, even with gRPC fuse off, I was running latest edge (2.4.2.0), and had to downgrade to be able to develop on my workflow... is this an issue w/ the latest edge?

tillprochaska:
Hi @kylepeeler, unfortunately, I can’t help with this as we’re working on a Ruby project and I don't have the latest edge build installed. You might want to have a look at the relevant issue in the docker repository (or ask your question there):

https://github.com/docker/for-mac/issues/4962

kylepeeler:
Yeah I posted here since it said in that issue it was fixed, but didn't seem to be the case for me... followed up in there as well! 😎

roschaefer:
@tillprochaska @kylepeeler a service similar to @dependabot would be great, which automatically reverts a commit that was marked as `#auto-revert: https://github.com/docker/for-mac/issues/4962` as soon as the referenced issue is closed.

We have plenty of [FIXME](https://github.com/tactilenews/100eyes/blob/master/Gemfile#L69) and [TODO](https://github.com/tactilenews/100eyes/blob/d30f2f69b5542774ffa67fd8d2cd8e7fb6f94d75/Gemfile#L84).

roschaefer:
@phoeinx @mattwr18 on my Linux machine, changes are propagated to the container, so no objections from my side. Could sb. of you who is on Mac review this PR?

tillprochaska:
For some reason, the fix suggested in the linked issue (disabling the gfuse file system) seems to have solved that problem on my machine. Not sure what made the difference. In case anyone’s having issues, feel free to reopen the PR.

PR #445, opened 2020-10-22
Use telegram webhooks in production

roschaefer:
In order to cleanup #434 I'm going to create this PR which sets up telegram webhooks in production first (avoid code creep @mattwr18 :wink: )

This will remove the telegram bot poller from our `docker-compose.prod.yml` and only use it for development. You can run
```
env RAILS_ENV=production bin/rails telegram:bot:set_webhook
```
 to set it up. You have to use `bin/rails c` to remove it, see https://github.com/telegram-bot-rb/telegram-bot/issues/143


tillprochaska:
I remember that there is info on how Telegram handles failed web hook requests (i.e. due to downtimes) and that it’ll retry sending the request a couple times. I’m not sure what delay there will before it retries to send the request or how often it’ll make an attempt, so that’d be good to know in advance. I couldn’t find it when I tried to search the docs just know, do you happen to know anything about this? 

And just to make sure I understood the `telegram-bot` docs correctly: This uses a route like `/telegram/<SECRET>` where `<SECRET>` is derived from the bot token?

roschaefer:
> I remember that there is info on how Telegram handles failed web hook requests (i.e. due to downtimes) and that it’ll retry sending the request a couple times. I’m not sure what delay there will before it retries to send the request or how often it’ll make an attempt, so that’d be good to know in advance. I couldn’t find it when I tried to search the docs just know, do you happen to know anything about this?

According to their docs, it's 24 hours: https://core.telegram.org/bots/api#getting-updates
> 
> And just to make sure I understood the `telegram-bot` docs correctly: This uses a route like `/telegram/<SECRET>` where `<SECRET>` is derived from the bot token?

Yes, you can check your bot's webhook with:
```
$ curl https://api.telegram.org/bot<TOKEN>/getWebhookInfo
{"ok":true,"result":{"url":"https://100eyes.roschaefer.de/telegram/<TOKEN>","has_custom_certificate":false,"pending_update_count":0,"max_connections":40}}
```
See: https://core.telegram.org/bots/api#setwebhook

tillprochaska:
> According to their docs, it's 24 hours: https://core.telegram.org/bots/api#getting-updates

Yes, I found that as well, what I could not find is info on how often and in which intervals Telegram will retry a webhook request in case a previous request has failed.

mattwr18 (changes requested):
Can you add instructions on how to test this out @roschaefer ??
I have the netcup.yml you sent me, but I don't know how to deploy, and it's good to have those instructions for our future selves

roschaefer:
> Can you add instructions on how to test this out @roschaefer ??
> I have the netcup.yml you sent me, but I don't know how to deploy, and it's good to have those instructions for our future selves

See https://github.com/tactilenews/100eyes/pull/445/commits/cfe988dfb79065b4f155facd28b00984b80ccffa

What's different in your case is that I've already send you the encrypted `inventories/custom/host_vars/netcup.yml` file. The nickname for the server in `inventories/hosts` would be `netcup`. Check out the bash script which I provided. Things like sendgrid configuration must be done manually through Sendgrid UI (and for this netcup server, this is already done).

mattwr18:
> > Can you add instructions on how to test this out @roschaefer ??
> > I have the netcup.yml you sent me, but I don't know how to deploy, and it's good to have those instructions for our future selves
> 
> See [cfe988d](https://github.com/tactilenews/100eyes/commit/cfe988dfb79065b4f155facd28b00984b80ccffa)
> 
> What's different in your case is that I've already send you the encrypted `inventories/custom/host_vars/netcup.yml` file. The nickname for the server in `inventories/hosts` would be `netcup`. Check out the bash script which I provided. Things like sendgrid configuration must be done manually through Sendgrid UI (and for this netcup server, this is already done).

ok, but what I got that the nickname should be netcup, but `What will be the new domain?`

ansible/README.md
```diff
@@ -11,8 +11,7 @@ ansible-galaxy install geerlingguy.pip
 
 Generate configuration files:
 ```bash
-$ cd ansible
-$ ./generate_config.sh
+$ ./ansible/generate_config.sh
 # follow the instructions
```
mattwr18, line 15: I don't think this is sufficient, like  I can see in the script that there are two questions, I guess. 
What's the nickname of your server?
I could reason that this was the same name as the `.yml`, in this case `netcup`
But the next question
What will be the new domain?
I can see that it is used on line 12, `traefik_domain="traefik.${domain}`, but what should I choose there? 
I see on link 34, you have given an example `traefik.tactile.news`, should I choose that for the test?

roschaefer: the domain is user-specific, so we cannot put it in the README, mine is `100eyes.roschaefer.de`

mattwr18: cool, sorry, I hadn't understood that I could skip to the part https://github.com/tactilenews/100eyes/blob/cfe988dfb79065b4f155facd28b00984b80ccffa/ansible/README.md#installation

mattwr18 (approved):
Works for me, thanks for walking me through how to set it up @roschaefer

tillprochaska:
> > According to their docs, it's 24 hours: https://core.telegram.org/bots/api#getting-updates
> Yes, I found that as well, what I could not find is info on how often and in which intervals Telegram will retry a webhook request in case a previous request has failed.

Any update on this, @roschaefer?

PR #446, opened 2020-10-22
@mattwr18 remove FIXME

roschaefer:


mattwr18 (approved):
yeah, I caught this too, and have removed it in https://github.com/tactilenews/100eyes/pull/437/files#diff-d09ea66f8227784ff4393d88a19836f321c915ae10031d16c93d67e6283ab55fL69-L70

But if you want to merge it in before that one, sure why not?

roschaefer:
saw that in #441

Issue #439, opened 2020-10-19
Investigate semantic alternatives for generic tags


tillprochaska:
https://github.com/tactilenews/100eyes/pull/432#issuecomment-712334103

mattwr18:
Yes! Thanks for opening this ticket. I think we could extend the scope of this ticket to also include the other concern I brought up, `divitis`, otherwise we could open a second ticket to discuss that? I will wait to respond to your comment https://github.com/tactilenews/100eyes/pull/432#issuecomment-712334103 until we've decided where to further that convesation.

As far as this topic, semantic HTML, I think we can and should continue to educate ourselves on the semantic alternatives to generic tags, as you put it. The documentation for this topic is vast, and has changed considerably in recent years, I guess. As I pointed out in the `PR` review, `divs`, according to the [MDN docs on %div](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div) 

>  should be used only when no other semantic element (such as %article or %nav) is appropriate.

And this is by far not the only place this type of declaration can be found, for example take the [W3C docs on %div](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-div-element.html), where they point out the accessibility issues, but also mention the maintainability issue as well.

> NOTE: Authors are strongly encouraged to view the div element as an element of last resort, for when no other element is suitable. Use of the div element instead of more appropriate elements leads to poor accessibility for readers and poor maintainability for authors.

A resource I find helpful is [this blog - Stop using so many divs! An intro to semantic HTML ](https://dev.to/kenbellows/stop-using-so-many-divs-an-intro-to-semantic-html-3i9i), which also has a second parter [Why I care about the Semantic Web](https://dev.to/kenbellows/why-i-care-about-the-semantic-web-2kn7)

I think the accessibility concerns is enough of an argument for us to prioritize this! I think it's important to point out that `a11y` concerns are not only about users with disabilities, as explained in the [MDN docs on accessibility](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Accessibility)

> However, it is wrong to say that accessibility is just about disabilities. Really, the aim of accessibility is to make your websites/apps usable by as many people in as many contexts as possible, not just those users using high-powered desktop computers.

[...]

PR #449, opened 2020-10-26
Add `lang` attribute to root element

tillprochaska:
@mattwr18: One positive outcome (hopefully of many others to come 😉) of our discussion over at #439 is I noticed we’re currently missing the `lang` attribute on the root element. This PR adds it so screen readers know which voice to use when reading out the contents. :)


mattwr18 (approved):
Nice, I am getting more and more in the habit of running the `aXe` accessibility tool to audit our site during `PR` reviews, and noticed this as well, but didn't act on it. Good job!!

PR #463, opened 2020-10-29
ci: bump ruby version for deployment

roschaefer:
I don't know why but Github actions was complaining about not finding ruby `2.7.1`.


tillprochaska (approved):
Hosted runners for GitHub Actions only support support selected versions for most languages. They seem to remove older patch versions when there are new releases. In this case, they removed 2.7.1 after 2.7.2 was released.

PR #464, opened 2020-10-29
feat: scope User::active

### LINKED ISSUE #417: Deactivate User Button
As editor
I want to have a deactivate-user-button
in order to soft-delete users who don't want to get messages anymore.

roschaefer:
close #417


mattwr18 (commented):
I know this is a draft, but I just thought I'd chime in here to say that I like the idea of not just relying on `active` as a boolean for this. I think it would be nice to have a `deactivated_at`, and possibly even `deactivation_cause`... Is that in your plan? What do you think?

roschaefer:
> I know this is a draft, but I just thought I'd chime in here to say that I like the idea of not just relying on `active` as a boolean for this. I think it would be nice to have a `deactivated_at`, and possibly even `deactivation_cause`... Is that in your plan? What do you think?

Hm, not really. I don't know what could be a possible use case for the `deactivated_at`. Furthermore it would require double-negation and more code.

If you can tell me a good reason, I'll reconsider that.

app/components/user_row/user_row.html.erb
```diff
@@ -11,7 +11,7 @@
       <span class="UserRow-name"><%= user.name %></span>
     </strong>
 
-    <span class="UserRow-channels">
+    <span class="<%= channel_css_classes %>">
```
roschaefer, line 14: What do you think, necessary to write a test for that? The entire `user_row` component is untested :shrug:

mattwr18:
> Hm, not really. I don't know what could be a possible use case for the deactivated_at. Furthermore it would require double-negation and more code.
If you can tell me a good reason, I'll reconsider that.
@roschaefer I would favor having more information than less. We could have a user that gets in contact and wants to know what's happened with their account, and it would be nice to be able to tell them more than it's not active.

You could just as easily have a method active?, which checks that deactivated_at is nil ... What's the argument for being less informative?

tillprochaska:
> De-active users have greyed-out channel icons ☝️

I feel like, from a UX perspective, we should try to omit encoding too much information in icons/colors. I think there are open tickets for the channel icons to either include explicit text labels or at least tooltips, and I’d like to make sure we don’t add additional inaccessible features.

Also, I remember that being able to see the number of signed-up contributors was quite important for @FrauCsu. I’d assume that it would also be important to see the numbers of active and inactive contributors separately.

Especially when a campaign like #50survivors runs over a longer period of time, I could imagine the list of inactive becoming bigger and bigger, so maybe we could just have two separate lists, one for active and one for inactive contributors?

We could use a tabbed navigation bar (similar e.g. to the repo nav on GitHub) to allow switching between the two lists and to show the counts of (in-)active contributors.


Screen Shot 2020-10-31 at 20 08 09

tillprochaska:
> Hm, not really. I don't know what could be a possible use case for the deactivated_at. Furthermore it would require double-negation and more code.

I’ve observed quite a few situations where editors on the #50survivors campaign where like "Wait, who added this to that user’s notes?" or "Do you know who added this user?" etc. I could imagine having a paper trail for an action as important as deactivating a user will be quite important.

That’s why I like @mattwr18’s idea of adding a note/reason for the deactivation, which would allow editors to explain why the deactivated a user and where to find more information (e.g. "Max Musterman called on Oct 31st asking to remove him from the list. Questions weren’t relevant to him anymore.").

While I don’t think this feature would have a super-high priority, and therefore doesn’t need to be included in this PR, I’d still like a `deactivated_at` timestamp instead of the boolean flag, if it’s not too time-consuming:

* This would mean we don’t need to migrate this column if we add the outlined feature later and we’d already have data for accounts deactivated before.
* I feel like it’s quite common for functionality like this, e.g. soft-delete usually is implemented using a `deleted_at` timestamp.

app/components/checkbox/checkbox.css
```diff
@@ -14,3 +14,52 @@
   flex-grow: 1;
   flex-shrink: 1;
 }
+
```
tillprochaska, line 17: Just curious, is there any particular reason for using a switch-style checkbox instead of just a default one?

tillprochaska (changes requested):

app/components/checkbox/checkbox.css
roschaefer: No, no real motivation for using a switch. I used a checkbox first and was unhappy with its look. Since @drjakob likes switches, I went for a switch.

roschaefer:
> Screen Shot 2020-10-31 at 20 08 09


Yes, I thought about that as well. However I wasn't quite sure how our users would use this feature. So, that's why I thought, maybe we should deploy it first and then see what exactly they need further.

PR #475, opened 2020-11-02
Postmark

tillprochaska:


config/environments/common/production.rb
```diff
@@ -104,23 +104,17 @@
...
 
-  config.mailer = {
-    from: Setting.sendgrid_from
+  config.action_mailer.postmark_settings = {
+    api_token: Setting.postmark_api_token
   }
 
-  config.action_mailbox.ingress = :sendgrid
-  config.project_name = ENV['HUNDRED_EYES_PROJECT_NAME']
+  config.action_mailer.default_options = {
```
tillprochaska, line 115: `config.action_mailer.default_options` is a Rails-native way to set the default `from` parameter (and other params to the `mail()` function, so we don’t need to do that manually.

app/mailers/message_mailer.rb
```diff
@@ -5,7 +5,18 @@ def new_message_email
...
+
+  private
+
+  def message_stream
+    if params[:broadcasted]
+      Setting.postmark_broadcasts_stream
+    else
+      Setting.postmark_transactional_stream
+    end
+  end
```
tillprochaska, line 21: Any ideas how to test this?

config/environments/common/production.rb
```diff
@@ -104,23 +104,17 @@
...
+    from: "#{Setting.project_name} <#{Setting.email_from_address}>"
```
tillprochaska, line 116: This is not directly related to the Postmark change. But I thought it might be a good idea to reduce redundancy and simply use the project name as the from name. Also, the from address isn’t really provider-dependant, right?

tillprochaska:
@roschaefer: If you have some spare time, it would be great if you could help me update the production setup to this change! Maybe we could have call to identify the necessary changes and I could try to apply them on my own then (I know you said you’d be quite busy on Tue and Wed.)

PR #767, opened 2021-03-11
build(update): ruby v3

roschaefer:
This should update only ruby to v3. I specified `alpine3.12` because
`alpine3.13` already has just postgres v13 tooling in their package
registry.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:2.7.2-alpine3.11
+FROM ruby:3.0.2-alpine3.12
 
 # Install dependencies
 RUN apk --update add \
--- a/app/components/attribute_bag.rb
+++ b/app/components/attribute_bag.rb
@@ -5,7 +5,7 @@ class AttributeBag
 
   attr_reader :attrs
 
-  def initialize(**attrs)
+  def initialize(attrs = {})
     @attrs = attrs
   end
 
@@ -19,7 +19,7 @@ class AttributeBag
     attrs
   end
 
-  def merge(**additional_attrs)
+  def merge(additional_attrs = {})
     new_attrs = attrs.dup
 
     if attrs.key?(:class) && additional_attrs.key?(:class)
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -7,8 +7,8 @@ RUN apk --update add \
     git \
     nodejs \
     yarn \
-    postgresql-dev=~12.6 \
-    postgresql-client=~12.6 \
+    postgresql-dev=~13.2 \
```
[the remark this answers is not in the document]
tillprochaska: @roschaefer I think we’d still want to upgrade to Ruby 3, right?

app/components/attribute_bag.rb
```diff
@@ -30,7 +30,7 @@ def merge(**additional_attrs)
     self.class.new(**new_attrs.deep_merge(additional_attrs))
   end
 
-  def defaults(**default_attrs)
+  def defaults(default_attrs = {})
```
roschaefer, line 33: @tillprochaska there is more in this file, I'm talking about `slice` and `except`. Apparently it's not executed during tests, can we remove?

tillprochaska: @roschaefer 

Both are used in components and tested:

https://github.com/tactilenews/100eyes/search?q=slice
https://github.com/tactilenews/100eyes/search?q=except

config/initializers/new_framework_defaults_6_1.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
```
tillprochaska, line 1: Why is this file added in this PR? We are already on Rails 6.1…

app/components/attribute_bag.rb
roschaefer: Right! I oversaw they have only one asterisk. So let's merge?

tillprochaska (approved):

PR #823, opened 2021-04-27
Restart Docker containers on reboot

### LINKED ISSUE #763: Ansible: Rebooting a VPS should restart Docker containers
Our current setup doesn’t seem to restart Docker containers after reboots. This is the reason why the `schweine` host wasn’t available today.

tillprochaska:
Previously, rebooting the docker host wouldn’t automatically restart the Docker containers, potentially causing downtime.

This patch adds an “always” restart policy to the Traefik reverse proxy, app, and database containers.

From the [documentation on restart policies](https://docs.docker.com/config/containers/start-containers-automatically/#use-a-restart-policy):

> Always restart the container if it stops. If it is manually stopped, it is restarted only when Docker daemon restarts or the container itself is manually restarted.

and

> If you manually stop a container, its restart policy is ignored until the Docker daemon restarts or the container is manually restarted. This is another attempt to prevent a restart loop.

I verified this actually works as expected following these steps:

* I ran the `site.yml` playbook, deploying to our staging environment using the changed Compose configuration.
* I verified the restart policies have been applied using `docker inspect -f "{{ .HostConfig.RestartPolicy }}" container_name`.
* I restarted the staging VPS using `sudo reboot`.
* I verified all containers are up after the restart using `docker ps`.

Closes #763.

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -28,9 +28,11 @@ services:
     volumes:
       - ./log:/app/log
       - ./storage:/app/storage
+    restart: always
 
   db:
     environment:
       POSTGRES_PASSWORD: "${POSTGRES_PASSWORD}"
       POSTGRES_USER: "${POSTGRES_USER}"
       POSTGRES_DB: "${POSTGRES_DB:-app_production}"
+    restart: always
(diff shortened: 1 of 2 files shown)


roschaefer (approved):
sure

PR #825, opened 2021-04-28
Add required data processing checkbox to onboarding forms

phoeinx:
Adds a mandatory consent to data processing checkbox on each onboarding form, this way only contributors that have given their consent can be created. The given consent is documented / saved as a timestamp in a `data_processing_consented_at` attribute for each contributor.

image

image


roschaefer (approved):
very nice

PR #832, opened 2021-05-02
Improved browsing UX / multi-column contributors layout

tillprochaska:
The primary goal of this PR is to improve the UX while browsing the contributor database. I started out adding a two-column layout to contributor profiles, allowing editors to easily switch between profiles without having to constantly go navigate back to the index page.

I ended up also optimizing some n+1 queries, as they lead to noticeably long server response times with 50+ contributors.

For in-detail reasoning about the changes, have a look at the individual commit messages :)

https://user-images.githubusercontent.com/1512805/116846812-7d215c00-abe9-11eb-83de-442fbeba19c9.mov


frontend/init/turbo.js
```diff
@@ -0,0 +1,30 @@
...
+
+  containers.forEach(container => {
+    const scrollPosition = scrollPositions[container.id];
+
+    if (scrollPosition && scrollPosition !== container.scrollTop) {
+      container.scrollTop = scrollPosition;
+    }
+  });
+};
+
```
roschaefer, line 28: Can we just use `id`s and `#` fragment identifier, please?

roschaefer: In Firefox and Chromium, this works fine with navigating back and forth.

app/components/two_column_layout/two_column_layout.js
```diff
@@ -0,0 +1,40 @@
...
+
+    // Calculate item bounds relative to sidebar
+    const itemTop = itemBounds.top - sidebarBounds.top - sidebar.scrollTop;
+    const itemBottom = itemTop + item.offsetHeight;
+
+    // Check if item is at least partially visible
+    return itemTop < sidebar.offsetHeight && itemBottom > 0;
+  }
+
+  scrollActiveItemIntoView() {
```
roschaefer, line 31: This would be obsolete with fragment identifier.

app/components/sidebar_item/sidebar_item.rb
```diff
@@ -6,11 +6,24 @@ def initialize(active: false, **)
...
 
-    attr_reader :active
+    def styles
+      return super << :active if active?
+
+      super
+    end
+
+    def attrs
+      return super.merge(data: { two_column_layout_target: 'activeItem' }) if active?
```
roschaefer, line 20: You told me about the `::target` pseudo-class. https://developer.mozilla.org/en-US/docs/Web/CSS/:target

spec/models/contributor_spec.rb
```diff
@@ -360,7 +360,7 @@
 
     describe 'number of database calls' do
       subject { -> { contributor.recent_replies.first.request } }
-      it { should make_database_queries(count: 1) }
+      it { should make_database_queries(count: 2) }
```
roschaefer, line 363: Why :thinking: ?

frontend/global/utils.css
```diff
@@ -33,3 +33,20 @@
   white-space: nowrap;
   width: 1px;
 }
+
+.skip-link {
+  position: absolute;
+  top: var(--spacing-unit-s);
+  left: var(--spacing-unit-s);
+  transform: translateY(-9999px);
+  z-index: 9999;
```
roschaefer, line 42: I'd prefer a `display: none`

app/views/contributors/show.html.erb
```diff
@@ -1,6 +1,10 @@
 <%= c 'two_column_layout', id: :contributors do |layout| %>
 
   <% layout.sidebar do %>
+    <a class="skip-link" href="#profile" data-turbo="false">
```
roschaefer, line 4: Nice idea! This is a whole new feature.

app/components/request_notification/request_notification.js
```diff
@@ -13,8 +13,14 @@ export default class extends Controller {
...
-    this.fetchMessages();
-    setInterval(() => this.fetchMessages(), POLLING_INTERVAL);
+    this.intervalHandle = setInterval(
+      () => this.fetchMessages(),
+      POLLING_INTERVAL
+    );
+  }
+
+  disconnect() {
+    clearInterval(this.intervalHandle);
```
roschaefer, line 23: I'll keep an eye out for this in the future :eyes:

roschaefer, line 23: I'll keep an eye out for this in the future :eyes:

roschaefer (approved):
nice

Quite nice @tillprochaska I learned a lot. However, I took the time to look carefully through the code and I'd prefer to remove obsolete code and discuss easier solutions before we merge.

app/components/two_column_layout/two_column_layout.js
tillprochaska: I thought linking to an anchor on the page might actually make this obsolete (even though it isn’t a solution to the code you commented on in [this comment](https://github.com/tactilenews/100eyes/pull/832/files/7f1c7204540d8ef864d936d1abd1bc361fe9c7c2..0c24ae9bbbf8e92de5764fd5f6688975719bf119#diff-11cfae26e2309368d917533b49425662d7d0206d3f7352e8746ac9116e7f618b). Unfortunately though, the browser isn’t smart enough to figure out the correct scroll position on its own… :(

Screenshot 2021-05-09 at 15 34 30

frontend/init/turbo.js
tillprochaska: As explained, unfortunately, this won’t prevent the scroll position from changing while navigating between different contributor profiles.

spec/models/contributor_spec.rb
tillprochaska: Just for reference, this is the solution recommended in the [Rails Guide](https://guides.rubyonrails.org/active_record_querying.html#eager-loading-associations) to load associated records. It requires an additional database query per association (i.e. the total numbers of queries is still constant), but requires less memory.

frontend/global/utils.css
tillprochaska: Just for reference. The reason not to use `display: none` or `visibility: hidden` in this case is that elements hidden like this will be removed from the tab order and screen readers will ignore them – in this case that would defeat the whole purpose of the skip link.

roschaefer (approved):
agreed

PR #839, opened 2021-05-03
Use relative base font size

### LINKED ISSUE #729: Update font-sizes/spacing-units to use rem instead of px
```
As a user
I would like to be able to adjust the default font-size and have it reflected on your site
Since I might have trouble reading small print
```

We should not use `px` to enforce a specific size, but rather use `rem` to be responsive to a user changing their default font size.

See:
https://engageinteractive.co.uk/blog/em-vs-rem-vs-px
https://zellwk.com/blog/media-query-units/

tillprochaska:
Closes #729


roschaefer (approved):
fantastic

PR #854, opened 2021-05-06
feat: automated deployment on `staging`

roschaefer:
How was this tested? In 21f7dd0 I included the current branch in the workflow.

I added the SSH key pair to 1Password.


.github/workflows/cd.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Install Ansible galaxies
+        run: |
+          ansible-galaxy collection install community.general
+          ansible-galaxy collection install devsec.hardening
+          ansible-galaxy install geerlingguy.docker
+          ansible-galaxy install geerlingguy.pip
+          ansible-galaxy install gantsign.inotify
+
+      - name: Run playbook
+        uses: dawidd6/action-ansible-playbook@v2
```
tillprochaska, line 48: Is it really necessary that we use a third-party action to run the playbook? Couldn’t we achieve the same using plain `run` steps? If you really want to use that action, could you pin it to a specific commit hash?

roschaefer:
@tillprochaska ping. I am not sure how much security we get, when we pin the action to a commit. An attacker has access to a dedicated SSH key-pair and the inventory file of `staging`. Not nice, but at least production shouldn't be affected.

I checked the source code of the Github action for security and found out about undocumented configuration `requirements`. Please re-review.

roschaefer:
I think rather than pinning the Github action to some commit, I'd prefer to add more security to our Dockerhub. E.g. we would win a lot if we'd hard-code docker image tags used in our ansible configuration. Before a deployment happens, a new tag would be pushed based on current `latest`.

mattwr18 (approved):
Sounds good to me! Let's give it a run! Have you created an issue for the security concern @tillprochaska

roschaefer:
issue created here #864

PR #862, opened 2021-05-07
Fix onboarding bugs and add server-side data processing consent validation

phoeinx:
This PR adds the following three things:

1. Server-side validation for the data processing consent attribute for contributors
 

2. A fix for a bug that caused the timestamp of data processing consent not being saved for contributors signing up via Telegram

3. A fix for a bug that enabled contributors to update the data of an other contributor if they were signing up via Telegram and updating their data after their telegram_id cookie had expired

### DIFF
--- a/app/controllers/onboarding/telegram_controller.rb
+++ b/app/controllers/onboarding/telegram_controller.rb
@@ -39,10 +39,14 @@ module Onboarding
     def update
       @contributor = Contributor.find_by(telegram_id: cookies.encrypted[:telegram_id])
 
+      unless @contributor
+        return redirect_to 'onboarding/unauthorized', status: :unauthorized
+      end
+
       if @contributor&.update(first_name: contributor_params[:first_name], last_name: contributor_params[:last_name])
         redirect_to_success
       else
-        render 'onboarding/unauthorized', status: :unauthorized
+        render :create
       end
     end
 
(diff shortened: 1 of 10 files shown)


app/controllers/onboarding/telegram_controller.rb
```diff
@@ -53,7 +59,7 @@ def redirect_to_success
     end
 
     def contributor_params
-      params.require(:contributor).permit(:first_name, :last_name, :email)
+      params.require(:contributor).permit(:first_name, :last_name, :email, :data_processing_consent)
```
roschaefer, line 62: GraphQL! :tired_face:

app/controllers/onboarding/telegram_controller.rb
```diff
@@ -37,12 +37,18 @@ def create
...
+      if @contributor&.update(
+        first_name: contributor_params[:first_name],
+        last_name: contributor_params[:last_name],
+        data_processing_consent: contributor_params[:data_processing_consent])
 
-      if @contributor&.update(first_name: contributor_params[:first_name], last_name: contributor_params[:last_name])
         redirect_to_success
       else
-        render 'onboarding/unauthorized', status: :unauthorized
+        render :create
```
roschaefer, line 51: When do we get here?

roschaefer (approved):
for now

As discussed, I'll merge for now and get the telegram onboarding flow refactored.

PR #867, opened 2021-05-08
fix(backup): treat blank values like `nil`

roschaefer:
See: https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#defaulting-undefined-variables


phoeinx:
Good catch! And shouldn't the same also be applied to the defaults of the installation role?

phoeinx:
For future reference - Robert and I discussed that it might be better to make users aware of blank values for the Postgres variables in their install configuration, than to silently set the provided default values. As missing values are written to `docker-compose.production.yml` the current behaviour in the installation role (aka no treating blank as nil) works better for this.

phoeinx (approved):

PR #870, opened 2021-05-08
feat: implement and test avatar_url= setter

### LINKED ISSUE #866: feat: Download+Save Telegram avatar
```gherkin
Feature: Download and save Telegram avatar
  As a user
  I would like the system to automatically download profile pictures of contributors who sign up via Telegram
  To distinguish contributors more easily
```

roschaefer:
swappy-20210508_212405

closes #866


app/models/contributor.rb
```diff
@@ -112,6 +112,15 @@ def active?
   end
   alias active active?
 
+  def avatar_url=(url)
+    begin
+      remote_file_location = URI(url)
+    rescue URI::InvalidURIError, ArgumentError
```
tillprochaska, line 118: You’re handling the `InvalidURIError` in case `url` is `nil` (or similar), e.g. in case the contributor has no profile picture set, right?

What’s the case for the `ArgumentError`?

tillprochaska:
Hi @roschaefer,

tested this using 1) a Telegram account with a profile picture and 2) with a Telegram account without a profile picture and it worked as expected.

Screenshot 2021-05-10 at 14 52 59

Screenshot 2021-05-10 at 15 03 27

app/models/contributor.rb
roschaefer: `InvalidURIError` handles "lksjdflsjdfl" and `ArgumentError` handles `nil`.

roschaefer: I believe a timeout would occur on line 121. Do you want me to test this?

tillprochaska: Regarding the use of `rescue` in order to handle `nil`, I think I’d have preferred testing for that explicitly. That would have made the intent clearer for me, e.g.

```ruby
return unless url
```

But that’s fine, I think. I’ll approve and you can decide if you want to change anything else.

tillprochaska (approved):

app/models/contributor.rb
roschaefer: Makes sense, :+1: for clear intention.

PR #881, opened 2021-05-10
Do not validate data processing consent on update

### LINKED ISSUE #869: bug: Cannot update contributor who was created manually
Steps to reproduce:
1. Create a contributor manually
swappy-20210508_182555
2. Click on contributor
3. Hit "Save"
swappy-20210508_182633

The problem is that contributors have not given `data_consent`. The only reason why it's possible to create contributors manually in the first place is that we have a special case for Telegram. The validation will kick in on the next `update`: https://github.com/tactilenews/100eyes/blob/master/app/models/contributor.rb#L19

tillprochaska:
Currently, we’re validating the presence of a data processing consent on every update. This is however only a workaround to make sure consent is required when contributors onboard using Telegram.

This patch removes the consent validation on every update, instead validates consent explcitly during the Telegram onboarding flow using the `contributor_signup` context.

Closes #869


app/controllers/onboarding/telegram_controller.rb
```diff
@@ -42,11 +42,13 @@ def update
 
       return render 'onboarding/unauthorized', status: :unauthorized unless @contributor
 
-      if @contributor&.update(
+      @contributor.assign_attributes(
```
tillprochaska, line 45: There’s no need for null-safe navigation in this case, as we’re returning in case `@contributor` isn’t set in the line before.

roschaefer (approved):
thumbs up

PR #882, opened 2021-05-10
refactor: telegram login flow

roschaefer:
Note: Telegram does not send the photo_url within a message. So after the PR our software will not save telegram avatars automatically.

reopen #866 

TODO:
- [x] Localize "Submit" button
- [x] "Drücke auf Start" instead of "Schick uns eine Nachricht"
- [x] Instead of `tg://resolve?domain=HundredEyesRobertDevelopmentBot&start=2GZVryjG` link to `t.me/HundredEyesRobertDevelopmentBot&start=2GZVryjG` - some devices cannot open `tg://` links but they are able to open Telegram after visiting the `t.me` website
- [x] Initially hide alternative options to connect. They should be visible after clicking a button "Any troubles to connect?" 
- [x] Show 8-digit `telegram_onboarding_token` as well as the name of the bot. This is another alternative way to connect with Telegram: A contributor searches for the bot in the Telegram app and sends the 8-digit code.
- [x] When a user already signed up and has `telegram_onboarding_token` not nil but `telegram_id: nil`, visiting `http://localhost:3000/onboarding?jwt=<JWT>` should redirect to `localhost:3000/onboarding/telegram/link/<TOKEN>`. This way, a contributor who clicks on the invitation link in the email (e.g. from the phone) will get to the correct page immediately. This should be doable as we save the JWT along with the contributor.


app/adapters/telegram_adapter/inbound.rb
```diff
@@ -1,64 +1,77 @@
...
+        telegram_onboarding_token = text.delete_prefix('/start').strip
+        sender = Contributor.find_by(telegram_id: nil, telegram_onboarding_token: telegram_onboarding_token)
+        if sender
+          sender.username = username
+          sender.telegram_id = telegram_id
+          @callbacks[CONNECT].call(sender)
+        else
+          @callbacks[UNKNOWN_CONTRIBUTOR].call
+        end
+      end
```
tillprochaska, line 70: I find the nested if blocks a little difficult to read/understand and would prefer early returns.

app/components/onboarding_response/onboarding_response.html.erb
```diff
@@ -3,5 +3,6 @@
     <%= c 'icon', icon: icon, class: 'OnboardingResponse-icon' %>
     <%= c('heading', tag: :h1) { heading } %>
     <p><%= text %></p>
+    <%= content %>
```
tillprochaska, line 6: Why is this necessary?

app/controllers/onboarding/telegram_controller.rb
```diff
@@ -1,87 +1,46 @@
...
-    layout 'onboarding'
+module Onboarding
+  class TelegramController < ChannelController
+    skip_before_action :verify_jwt, only: %i[link fallback]
+    before_action :ensure_contributor_exists, only: %i[link fallback]
 
     def show
-      @jwt = jwt_param
+      super
+      @contributor.telegram_onboarding_token = SecureRandom.alphanumeric(8)
```
tillprochaska, line 12: I think for UX reasons, we should either use upper- or lowercase characters only and make sure to handle onboarding tokens case-insensitively.

app/controllers/telegram/webhook_controller.rb
```diff
@@ -1,15 +1,53 @@
...
+    end
+  end
+
+  private
+
+  def telegram_welcome_message
+    [
+      Setting.onboarding_success_heading,
+      Setting.onboarding_success_text
+    ].join("\n")
```
tillprochaska, line 51: This is a temporary solution and we’re probably going to change that in #205 anyways, right?

db/migrate/20210510092738_add_telegram_onboarding_token.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddTelegramOnboardingToken < ActiveRecord::Migration[6.1]
+  def change
+    add_column :contributors, :telegram_onboarding_token, :string
```
tillprochaska, line 5: This should probably have a unique constraint for the unlikely case that the random token we generate already exists, right?

app/components/onboarding_response/onboarding_response.html.erb
tillprochaska: Ah, nevermind – it was necessary before, but became irrelevant with my changes! :)

config/locales/de.yml
```diff
@@ -90,12 +90,21 @@ de:
...
+      copy: Link kopieren
+      help: Über den Button gelangen Sie direkt zum Telegram-Profil von „%{project_name}”. Alternativ können Sie einen Link kopieren und auf Ihrem Smartphone öffnen.
+      didnt_work: Hat das nicht funktioniert? Oder werden Sie nach einem Code gefragt?
+    onboarding_telegram_fallback:
+      heading: Fast geschafft!
+      text: Verbinden Sie Ihr Telegram-Konto in der Telegram-App, um die Anmeldung abzuschließen.
+      steps:
+        - "Finden Sie uns auf Telegram. Suchen Sie dazu nach <strong>%{telegram_handle}</strong>."
+        - "Wenn Sie unser Telegram-Konto gefunden haben, klicken Sie auf <strong>Start</strong>."
+        - "Geben Sie diesen Code ein, sobald sie danach gefragt werden: <strong>%{token}</strong>."
```
tillprochaska, line 107: Let’s remove the period at the end so people can’t get confused over wether the period is part of the code or not.

```suggestion
        - "Geben Sie diesen Code ein, sobald sie danach gefragt werden: <strong>%{token}</strong>"
```

app/controllers/telegram/webhook_controller.rb
```diff
@@ -1,15 +1,53 @@
...
-    contributor = telegram_message.sender
-    TelegramAdapter::Inbound.bounce!(chat) and return unless contributor
+    adapter.on(TelegramAdapter::UNKNOWN_CONTRIBUTOR) do
+      respond_with :message, text: Setting.telegram_contributor_not_found_message
+    end
 
-    respond_with :message, text: Setting.telegram_unknown_content_message if telegram_message.unknown_content
-    contributor.reply(telegram_message)
+    adapter.consume(payload) do
+      respond_with :message, text: telegram_welcome_message
```
tillprochaska, line 14: Why can’t we reply with the welcome message directly in the callback for the `CONNECT` event?

roschaefer: I did that because you might want to `/start` after being connected

app/controllers/onboarding_controller.rb
```diff
@@ -14,13 +12,86 @@ def index
...
+
+    json_web_token = JsonWebToken.includes(:contributor).find_by(invalidated_jwt: jwt_param)
+    return unless json_web_token
+
+    contributor = json_web_token.contributor
+    if contributor && contributor.telegram_id.blank? && contributor.telegram_onboarding_token.present?
+      return if %w[link fallback].include? action_name
+
+      redirect_to onboarding_telegram_link_path(telegram_onboarding_token: contributor.telegram_onboarding_token) and return
+    end
```
roschaefer, line 92: This part is a little unfortunate as the `onboarding_controller` knows two controller actions of `telegram_controller` and also redirects to a specific path there. I don't know how to handle this any better.

app/controllers/onboarding_controller.rb
```diff
@@ -14,13 +12,86 @@ def index
...
+    raise ActionController::BadRequest
+  rescue ActionController::BadRequest, JWT::DecodeError
```
roschaefer, line 94: Any reason why we caught `StandardError` here before? @mattwr18

roschaefer: https://github.com/jwt/ruby-jwt/blob/3434f585d8b5b88c483314b193077fc2e8ef0b72/lib/jwt/error.rb#L12 looks safe enough to me, no?

drjakob:
Please also check this scmall bug: https://github.com/tactilenews/100eyes/issues/895

app/components/onboarding_telegram_form/onboarding_telegram_form.html.erb
```diff
@@ -1,27 +1,35 @@
...
+    <%= c 'stack' do %>
+      <%= c 'field', object: contributor, id: :first_name do |field| %>
+        <%= c 'input', field.input_defaults.merge(required: true) %>
+      <% end %>
+
+      <%= c 'field', object: contributor, id: :last_name do |field| %>
+        <%= c 'input', field.input_defaults.merge(required: true) %>
+      <% end %>
+
+      <%= c 'input', id: 'contributor[telegram_onboarding_token]', type: 'hidden', value: contributor.telegram_onboarding_token %>
```
tillprochaska, line 21: Any specific reason to set the Telegram token in `#show`, potentially allowing a contributor to alter it? Instead of setting it on the fly in `#create` before persisting the record to the database?

I don’t really see if/how this would be a problem, but it seems a little unusual.

app/controllers/onboarding_controller.rb
tillprochaska: I don’t have a good idea how to solve the specific problem, but maybe it makes sense to extract that logic into a separate method?

I tried something like the following. Not sure I really like that so much more than the current implementation, though. Thought I’d push it to a separate branch so you can see if you like it…

https://github.com/tactilenews/100eyes/blob/refactor_telegram_login_flow_till/app/controllers/onboarding_controller.rb#L85

spec/requests/contributors_spec.rb
```diff
@@ -93,7 +93,7 @@
     end
 
     context 'given a manually created contributor' do
-      let(:contributor) { create(:contributor, :manually_created) }
+      let(:contributor) { build(:contributor, data_processing_consent: false).tap { |c| c.save(validate: false) } }
```
tillprochaska, line 96: This is to workaround the consent validation on `create`, right?

app/models/contributor.rb
```diff
@@ -10,13 +10,15 @@ class Contributor < ApplicationRecord
...
+  has_one :json_web_token, dependent: :destroy
+  accepts_nested_attributes_for :json_web_token
 
   acts_as_taggable_on :tags
 
   default_scope { order(:first_name, :last_name) }
   scope :active, -> { where(deactivated_at: nil) }
 
-  validates :data_processing_consent, acceptance: true, on: :contributor_signup
+  validates :data_processing_consent, acceptance: true, on: :create
```
tillprochaska, line 21: This introduces a new bug (related to the bug we initially introduced when adding validation for the data processing consent). This time, you won’t be able to manually create a contributor via the UI.

Why can’t we keep the previous `contributor_signup` context?

spec/requests/onboarding/telegram_spec.rb
```diff
@@ -1,170 +1,111 @@
...
-        expect { subject.call }.to change(JsonWebToken, :count).by(1)
+    describe 'with invalidated jwt' do
+      let!(:invalidated_jwt) { create(:json_web_token, invalidated_jwt: 'INVALID_JWT') }
+      let(:jwt) { 'INVALID_JWT' }
 
-        json_web_token = JsonWebToken.where(invalidated_jwt: jwt)
-        expect(json_web_token).to exist
+      describe 'http status' do
+        subject { super().call && response }
+        it { is_expected.to have_http_status(:unauthorized) }
```
tillprochaska, line 108: I’m not sure I understand this. Why do we need `super().call`?

spec/requests/onboarding/telegram_spec.rb
```diff
@@ -173,78 +114,63 @@
...
+    describe 'redirects' do
+      it 'are skipped' do
+        is_expected.to have_http_status(:ok)
+      end
 
-      it 'updates the contributor' do
-        subject.call
-        contributor = Contributor.find_by(telegram_id: 789)
-        expect(contributor).to have_attributes(first_name: 'Update', last_name: 'MyNames')
+      describe '(sanity check: / redirects to /link)' do
```
tillprochaska, line 170: I don’t really understand why this test is in this place, as a sanity check? Shouldn’t this either be a separate test for the `GET /onboarding/telegram` request? (If necessary at all, I think it’s covered by other requests, right?)

tillprochaska (changes requested):

tillprochaska:
Hi @roschaefer, requesting changes as this introduces a regression/bug (see [this comment](https://github.com/tactilenews/100eyes/pull/882#discussion_r633620336)). 

Apart from that, I’m not sure if [this](https://github.com/tactilenews/100eyes/pull/882#discussion_r633620336) might be problem.

app/models/contributor.rb
roschaefer: I'll check this as soon as I have time. From the law perspective I'd say we should validate `data_processing_context` when personal data is added to our system. The validation shouldn't be called explicitly, but by default instead.

tillprochaska: > From the law perspective I'd say we should validate data_processing_context when personal data is added to our system. The validation shouldn't be called explicitly, but by default instead.

Yes, that’s fine for me. Then we’d need to make sure we explicitly skip the data processing validation when manually creating contributors.

spec/requests/onboarding/telegram_spec.rb
roschaefer: `have_http_status` is defined on `response`. In order to write `is_expected.to have_http_status(...)` I first call `super()` for the actual request and make `response` the new return value.

spec/requests/onboarding/telegram_spec.rb
roschaefer: This is a sanity check for the test setup. By changing `let(:action)` I make sure the test setup is alright and stays up-to-date even after we (maybe) refactor the routing behaviour.

app/components/onboarding_telegram_form/onboarding_telegram_form.html.erb
roschaefer: Yes, makes sense. I can overwrite `contributor_params` and `merge(telegram_onboarding_token: Securerandom..)`

spec/requests/contributors_spec.rb
roschaefer: Si

app/components/onboarding_telegram_form/onboarding_telegram_form.html.erb
roschaefer: No, I decided against this. See 5ac3f6101a2e57e399d86720ce5982bc7c277237

app/models/contributor.rb
roschaefer: I added a hidden input when `contributor.new_record?` in 5ac3f6101a2e57e399d86720ce5982bc7c277237

roschaefer: Speaking of 5ac3f6101a2e57e399d86720ce5982bc7c277237 I removed `status: bad_request` when validations fail. It's rails default behaviour to send 200 on data validations and show error messages to the user.

app/components/contributor_form/contributor_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= form_with(model: contributor, local: true, class: "ContributorForm", multipart: true) do %>
+<%= form_with(model: contributor, local: true, class: "ContributorForm", multipart: true, data: { turbo: false }) do %>
```
tillprochaska, line 1: Why’s this necessary?

tillprochaska (approved):

app/components/contributor_form/contributor_form.html.erb
roschaefer: That's something I caught along the way: Flash messages (validation errors) won't be displayed otherwise.

PR #912, opened 2021-05-24
Add new icons to dashboard / homepage

phoeinx:
This PR:

- adds new icons to `icons.svg,` and  old icons that are not needed anymore
- adds `icon-background.svg `because using SVG fragments as the content of a CSS pseudo-element (for the background of icons) didn't work
- updates dashboard view to accommodate new icons
- for this adds a new component: `icon_list`
- deletes not used any more component `steps`

The updated part of the dashboard now looks like this:

image

image


app/components/steps_list/steps_list.html.erb
```diff
@@ -1,4 +1,4 @@
-<ol <%= attrs %>>
+<ol class="StepsList">
```
phoeinx, line 1:
```suggestion
<ol <%= attrs %>>
```

app/components/icon_list/icon_list.css
```diff
@@ -30,6 +30,7 @@
 
 .IconList--row-3 .IconList-element {
   flex: 1 0 30%;
+  gap: 0 calc(2/3 * var(--spacing-unit));
```
tillprochaska, line 33: Before, the gap between the icon and the text and between multiple elements of the icon list where the same. In some cases, this can make it *slightly* more difficult to visually associate which icons and texts belong to each other. Making the gap between the icon and text of an individual elements slightly smaller than the gap between multiple elements helps.

Screen Shot 2021-06-07 at 11 13 02

Screen Shot 2021-06-07 at 11 12 33

(Of course, there are other factors that impact how we associate elements laid out visually, e.g. text alignment.)

app/views/dashboard/index.html.erb
```diff
@@ -2,26 +2,37 @@
 
 <section>
   <%= c 'box', style: :large do %>
-    <%= c 'wrapper' do %>
+    <%= c 'wrapper', style: :narrow do %>
```
tillprochaska, line 5: I’d personally prefer the heading and icon list to be aligned, as they have a closer association.

Screen Shot 2021-06-07 at 10 44 24

app/components/icon_list/icon_list.html.erb
```diff
@@ -8,11 +8,13 @@
       
       <div class="IconList-text">
         <% if elem[:title] %>
-          <div class="IconList-title"> <%= elem[:title] %> </div>
+          <%= c 'heading', tag: :h3, class: 'IconList-title' do %>
+            <%= elem[:title] %>
+          <% end %>
```
tillprochaska, line 13: Actually, I’m not 100% sure if using headings inside of list items is valid and well-supported by assistive tech, I’ll check later. As an alternative, we could use a `strong` tag.

app/components/icon_list/icon_list.html.erb
```diff
@@ -1,5 +1,5 @@
 <ul <%= attrs %>>
-  <% I18n.t('components.lists.'.concat(list)).each do |elem| %>
```
tillprochaska: Maybe there was a specific reason for this implementation that I’m missing here. If that’s the case, feel free to drop this commit.

In general, I think I’d prefer explicitly passing the contents to the list via the component’s properties instead of just the I18n key. Technically, the icon list might be used for anything, that the contents come from the localization files in this particular case is irrelevant to the component.

tillprochaska (approved):
Looks good. I’m not going to merge it directly because you might want to have a look at [this comment](https://github.com/tactilenews/100eyes/pull/912#discussion_r646516752). Feel free to merge afterwards! :)

app/components/icon_list/icon_list.html.erb
phoeinx: I think that was just a legacy from the previous `steps` component, adapted a bit to fit the new requirements. But I like this version much better.

app/components/icon_list/icon_list.html.erb
phoeinx: Have you had a look already? 
From what I've found, using headings inside of list items seems to be valid because they are block-level elements. The only thing to watch out for is that the semantic structure is consistent for each list element, which is the case here. 
And in regards to assistive tech, I found a WCAG guideline to always use structural markup in a way that does represent relationships in the content. This is also the case here. 

I am not really sure where to refer to make sure that the actual assistive tech implementations support headings inside of list elements. If you have resources for that in mind, I would be happy for a pointer.

But so far, so good, and I am gonna merge now. :) 
Thank you for the review!

tillprochaska: > Have you had a look already?

No, but will do later and open a new PR in case smth needs to be fixed.

> I am not really sure where to refer to make sure that the actual assistive tech implementations support headings inside of list elements. If you have resources for that in mind, I would be happy for a pointer.

If unsure, I’m usually testing in VoiceOver on macOS, although that’s probably not the best way, as e.g. other screen readers for other platforms have higher market shares and might behaves differently.

Apart from that, I usually simply google or browse through blogs by authors who’ve written a lot on that topic.

PR #921, opened 2021-06-02
test: TelegramAdapter::Inbound#avatar_url

### LINKED ISSUE #899: Automatically show Telegram profile pic (re installment)
As editor
I want to see the profile picture my contributor uses in telegram 
after she/he/div signs in.

roschaefer:
This includes a refactoring. There are helper methods on `Telegram::Bot` for every API method. I'd like to use them, but in our tests we have a stubbed telegram bot. So I decided to enable actual HTTP requests for one particular spec. In order to do that, I'd had to filter the API keys and I also removed the assignments to `ENV`.

I did *not* write a migration to refresh all `avatar_url`s, simply because one should not use migrations to set up the data itself. If we want, we can do this manually:
```rb
adapter = TelegramAdapter::Inbound.new
Contributor.where.missing(:avatar_attachment).where.not(telegram_id: nil).find_each do |contributor|
  contributor.avatar_url = adapter.avatar_url(contributor)
end
```

The behaviour is now: For every incoming telegram request, if the `avatar_url` of `contributor` is `nil` then try to download the avatar. If the `avatar` is already set, don't do anything. Ie. the avatar will never be updated. If a contributor has no avatar, a request will be made for every incoming message.

- [x] I'd like to test this with a telegram user who has no profile picture. I seem not to be able to remove mine.

close #899


app/adapters/telegram_adapter/inbound.rb
```diff
@@ -48,6 +50,17 @@ def consume(telegram_message)
...
     end
 
+    def avatar_url(contributor)
+      return unless contributor.telegram_id
+
+      profile_photos = Telegram.bot.get_user_profile_photos(user_id: contributor.telegram_id, limit: 1).with_indifferent_access
+      first_photo = profile_photos.dig(:result, :photos, 0)
+      return unless first_photo
+
+      largest_size = first_photo.max { |a, b| a[:file_size] <=> b[:file_size] }
```
tillprochaska, line 60: I was wondering if we might want to reduce duplication of this line (we select the highest resolution for images attached to messages as well as profile photos) and extract it into something like `photo_url`? On the other hand, it’s just a single line duplicated in two places, so definitely nothing super important.

spec/components/chat_preview_spec.rb
```diff
@@ -4,6 +4,9 @@
 
 RSpec.describe ChatPreview::ChatPreview, type: :component do
   subject { render_inline(described_class.new(**params)) }
+  before do
```
tillprochaska, line 7: I haven’t really understood why removing the respective environment variables from the setup is necessary or why you prefer stubbing the settings in all those specs?

tillprochaska (approved):
Hey @roschaefer, sorry for the late review. I added a few comments, but that’s because I’m not sure wether I understood everything correctly, would be great if you could clarify :)

app/adapters/telegram_adapter/inbound.rb
roschaefer: Yep, I thought about putting that line in a central location but decided against it.

spec/components/chat_preview_spec.rb
roschaefer: That was a refactoring for code quality: it's considered bad practice to write into `ENV` and we do this for our test environment. Second, you have to look into several locations to understand the test, the assertion is in the spec and the stubbed value in `config/environments/test.rb`.

Besides that I wanted to make sure that we never commit our Threema API keys by adding:
```
  c.filter_sensitive_data('THREEMARB_API_SECRET') { ENV['THREEMARB_API_SECRET'] }
  c.filter_sensitive_data('THREEMARB_PRIVATE') { ENV['THREEMARB_PRIVATE'] }
  ```
which reads from `ENV`.

PR #924, opened 2021-06-04
Fix Ansible deployment

tillprochaska:
The `dev-sec/os-hardening` role changed a UFW default causing IPv6 to be enabled by default. Our DO hosts don’t have IPv6 addresses, resulting in errors during deployment.

As the UFW defaults are set in the `os-hardening` role, we needed to change the execution order of roles, i.e. the `os-hardening` roles is now executed before the `initial_setup ` role.

### DIFF
--- a/ansible/site.yml
+++ b/ansible/site.yml
@@ -11,9 +11,11 @@
   collections:
     - devsec.hardening
   roles:
-    - role: initial_setup
     - role: devsec.hardening.ssh_hardening
     - role: devsec.hardening.os_hardening
+      vars:
+        ufw_enable_ipv6: false
+    - role: initial_setup
     - role: gantsign.inotify
       inotify_max_user_watches: 524288
     - role: geerlingguy.pip


ansible/site.yml
```diff
@@ -11,9 +11,11 @@
   collections:
     - devsec.hardening
   roles:
-    - role: initial_setup
     - role: devsec.hardening.ssh_hardening
     - role: devsec.hardening.os_hardening
+      vars:
+        ufw_enable_ipv6: false
+    - role: initial_setup
```
roschaefer, line 18: The PR description makes sense to me, except why the order of `initial_setup` and `devsec.hardening.os_hardening` must be changed? Can you explain? Where is the dependency?

tillprochaska: Yep, sure, that could’ve been clearer. In `initial_setup`, we set up new UFW rules, causing the error. I.e. the actual error occurs in `initial_setup`, although the cause lies in `os_hardening`.

If kept the order of the roles, the playbook would always fail in the `initial_setup` role, before the updated UFW default would be applied by the `os_hardening` role.

As the `os_hardening` and `ssh_hardening` roles apply a basic set of changes that do not (at least as far as we could see) depend on tasks from `initial_setup`, it might make sense to run these first.

Could changing the order be a problem?

roschaefer: OK.. so the problem is that setting a default fails because a UFW rule is already in place? There is no problem changing the order, I just wanted to learn something.

roschaefer (approved):
learn

ansible/site.yml
tillprochaska: Just to be clear, the problem is not that setting the default fails, but that running Ufw with the wrong defaults fails. And if we run Ufw before setting the defaults, we'd never get to the point where we could update the defaults…

tillprochaska:
Turns out, this didn’t fix the issue when deploying via GitHub Actions… 🙃

PR #948, opened 2021-06-27
Fix controller bug when updating contributor with invalid values

tillprochaska:
I discovered this bug that I probably introduced when adding the new contributors layout. When updating a contributor with invalid values, `@contributors` was not set which led to an error when rendering the view. This only happened when updating a contributor with invalid values because we redirect on successful form submissions.


phoeinx (approved):
lgtm

PR #949, opened 2021-06-27
894 passwords feedback

### LINKED ISSUE #639: Allow manually setting `Field` id
Right now, we have to rely on the magic the `Field` component does (e.g. generating the input id based on an object) in all cases. It works well for all kinds of classic CRUD forms for a specific record, but requires workarounds to make it work with other kinds of forms.

I’d suggest to implement a backward compatible change that allows us to use that magic when we need it (e.g. CRUD forms), but allows us to use the field component without an associated object in other cases without workarounds.

This is handy, and I think we should keep it:
```ruby
    <%= c 'field', object: contributor, id: :first_name do |field| %>
      <%= c 'input', field.input_defaults %>
    <% end %>
```

This currently would throw an error, but should work imo:
```ruby
    <%= c 'field', id: '2fa[otp]' do |field| %>
      <%= c 'input', field.input_defaults %>
    <% end %>
```

### LINKED ISSUE #892: Fix banner height
tbd

Screenshot 2021-05-14 at 22 57 14

### LINKED ISSUE #894: Missing feedback for bad passwords
As editor
I want to get feedback if resetting my password an password is too short (or failing other requirements). Deactivated SET-NEW-PASSWORD Button is insufficient.

### LINKED ISSUE #969: Password-Length: Relax password strength
We're already using 2FA and yet we enforce a password length of at least 20 characters(!).

swappy-20210716_134326

I think this is overkill. I'm sitting in a demo in which a group of users was onboarded. They had some difficulties to come up with such a long password.

tillprochaska:
https://user-images.githubusercontent.com/1512805/131676542-a5a61bc4-c116-4507-bf24-1546ebea2b97.mov

<details>
<summary>Show screenshots of complete sign in/password reset flow</summary>

## Sign in
Screen Shot 2021-08-01 at 18 44 57

## Reset Password
Screen Shot 2021-08-01 at 18 45 02

Screen Shot 2021-08-01 at 18 45 08

Screen Shot 2021-08-01 at 18 45 37
</details>

Close #894, close #892, close #639, close #969


roschaefer (changes requested):
looks good

This PR already looks really good and it is ready to be merged, no?

I would like to improve the UX though. I'm getting confused by the `password` and `otp_code` inputs in the very same UI component. If we cannot put them on different pages, could we at least make a visual split between the two?

swappy-20210901_235245


Also, the validation error message is not quite true. The password *is* 8 characters but the `otp_code` was wrong.

tillprochaska:
> Also, the validation error message is not quite true. The password is 8 characters but the otp_code was wrong.

True! I updated the error message.

> I would like to improve the UX though. I'm getting confused by the password and otp_code inputs in the very same UI component. If we cannot put them on different pages, could we at least make a visual split between the two?

I think we should still put them on different pages given the feedback we got from editors. I have a POC for that, but I’m not 100% sure about that yet, so I’d like to do that in a separate PR.

For users that don’t have OTP enabled, the OTP field isn’t displayed. As the primary use case for password resets is onboarding of new editors (i.e. they don’t have OTP enabled yet) and given that the ideal solution is a true two-step flow, maybe we can keep it like this for now?

Screen Shot 2021-09-02 at 12 19 59

tillprochaska:
@roschaefer Ich habe meinen Ansatz mal ein bisschen aufgeräumt und in #1035 gepusht (ab https://github.com/tactilenews/100eyes/pull/1035/commits/c66d8562789c45e470798852c98dae1437f2d1ef).

roschaefer (approved):
OK, so the changes are done via a follow-up PR? Fine for me. I'll approve.


fine for me

PR #950, opened 2021-06-28
Update README.md

drjakob:
Updatet ##funding


FrauCsu (changes requested):
wording: change supportet into "suppported" and modules into "features"
license: discuss MIT and review different licenses with me, @drjakob

PR #962, opened 2021-07-05
feat: notify users + bold text for threema

roschaefer:
This was a follow-up from @tillprochaska's review here:
https://github.com/tactilenews/100eyes/pull/930#pullrequestreview-690620976

swappy-20210705_215225

### DIFF
--- a/app/adapters/threema_adapter/outbound.rb
+++ b/app/adapters/threema_adapter/outbound.rb
@@ -15,10 +15,13 @@ module ThreemaAdapter
       perform_later(recipient: recipient, text: message.text)
     end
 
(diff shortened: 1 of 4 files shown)


drjakob:
I like it **very much**.

phoeinx (approved):
Looks good to me! I have one small suggestion regarding test readability, feel free to merge either way.
(When I was reading the `'strips whitespace to not break basic formatting'` test I was confused as to why you also added whitespace in front of the string and only wrote about trailing whitespace inside of the string. It makes a lot of sense to test that leading and trailing whitespace is removed and writing that down in the string, would improve readability imo.)

PR #972, opened 2021-07-19
958 add signal messenger

### LINKED ISSUE #958: Add Signal messenger
```gherkin
Feature: Add Signal messenger
  As a user
  I would like to communicate via Signal private messenger
  For the sake of encryption and to cover more people who use only Signal
```

I did 2 hours of research and I can confirm that it is technically possible to use Signal for our application. It requires an extra service which we need to deploy (+ complexity) and a separate phone number for every instance (+ fees, + rollout time). @drjakob and @FrauCsu told me via phone that they want this feature and are OK with the costs.

Have a look into [this repo](https://github.com/bbernhard/signal-cli-rest-api) and the corresponding [swagger api](https://bbernhard.github.io/signal-cli-rest-api/). 

**Beware** If you [register your phone number](https://bbernhard.github.io/signal-cli-rest-api/#/Devices/post_v1_register__number_) and then [verify your phone number](https://bbernhard.github.io/signal-cli-rest-api/#/Devices/post_v1_register__number__verify__token_) you will loose access to your signal messages from your phone. You can change your device after a period of 7 days only.

1. Download the `docker-compose.yml` from the `README.md`.
2. Run `docker-compose.yml`
3. Check the [Swagger API](https://bbernhard.github.io/signal-cli-rest-api/#/Devices/post_v1_register__number_) and copy the command to verify your phone.
4. If you get `curl: (35) error:1408F10B:SSL routines:ssl3_get_record:wrong version number` replace `https` with `http`:
```
curl -X 'POST' 'http://127.0.0.1:8080/v1/register/<phone_number>' -H 'accept: application/json' -H 'Content-Type: application/json'
```
You will see an error message like: 
```
{"error":"Captcha required for verification, use --captcha CAPTCHA\nTo get the token, go to https://signalcaptchas.org/registration/generate.html\nCheck the developer tools (F12) console for a failed redirect to signalcaptcha://\nEverything after signalcaptcha:// is the captcha token.\n"}
```
5. Go to https://signalcaptchas.org/registration/generate.html and solve a captcha. Observe your JavaScript console for the respective token:
swappy-20210701_213129
6. Run the command again, this time with captcha:
```
curl -X 'POST' \
   'http://127.0.0.1:8080/v1/register/<phone_number>' \
   -H 'accept: application/json' \
   -H 'Content-Type: application/json' \
   -d '{ "captcha": "<captcha>" }'
```
7. You will receive a text message with your verification code. When you use this code to [verify your phone number](https://bbernhard.github.io/signal-cli-rest-api/#/Devices/post_v1_register__number__verify__token_) you unlink any other device and you are able to receive messages from your service.
8. Enjoy? :thinking:

roschaefer:
close #958


roschaefer:
Here's a summary of what's included (as of now):

* signal-cli-rest endpoint in `docker-compose.yml`
* onboarding via Signal
* phone number normalization and validation
* receiving text messages
* sending text messages
* display of signal channel in contributor overview

What's missing:
* deployment configurations
* conveniently setup new phone numbers
* webhooks in favour of busy-wait (requires contribution to `signal-rest-cli`)

config/locales/signal/unknown_content.txt
```diff
@@ -0,0 +1 @@
+Unsere Software kann derzeit Textnachrichten verarbeiten, aber noch nicht andere Inhalte wie Bilder, Sprachnachrichten oder Dateianhänge. In der Zwischenzeit senden Sie bitte Nachrichten mit solchen Inhalten an Astrid Csuraji, Telefonnummer <ASTRIDS_TELEFONNUMMER>. Herzlichen Dank!
```
roschaefer, line 1: @FrauCsu can I add your real phone number here?

FrauCsu: No, please don't

FrauCsu: Can't we use the prepaid number @drjakob bought?

tillprochaska:
For future reference, this is what I discussed with @roschaefer today:

> phone number normalization and validation

It would be nice to keep the naming scheme of database columns/attributes of identifiers for the different channels consistent, e.g. using `signal_phone_number`. The current naming is a little ambiguous.

> raise error when someone completely unknown sends us a message (-> Sentry)
  
  It feel it’s a little hacky to use Sentry for errors we actually expect. However, implementing a real notification systems seems a little out of scope (at least for this PR). Apart from that, Sentry actually offers handy features e.g. to send issues to a #100eyes-support channel on Slack. I’d prefer if we could make sure that all of these issues do have a separate tag in Sentry, so that we can easily distinguish between actual runtime errors of the application vs. those contributor-related errors.

   Additionally, as a follow-up to this PR, we should handle unknown senders consistently across all communication channels.

> deployment configurations
   
> conveniently setup new phone numbers

   We briefly discussed the idea to have a single `signal-cli-rest-api` service for all our 100eyes instances. However, this likely won’t simplify the deployment/setup process for new 100eyes instances (instead, it might actually make more work). The primary factor in terms of manual work/developer time is the registration/verification process for new phone numbers/account creation. Wether we use a single `signal-cli-rest-api` for all 100eyes instances vs. one per 100eyes instance is irrelevant, as we’ll need separate phone numbers for every 100eyes instance anyways.

   In order to make deployment/Signal setup as easy as possible, it would be preferable if we used a telco/service provider that allows us to receive SMS via a web interface, so that any developer is able to access Signal verification codes via SMS at any time.

   Apart from that, we should automate the steps to setup a new Signal account (e.g. sending the registration, verification request to `signal-cli-rest-api`) using a shell script or Ansible playbook.

> webhooks in favour of busy-wait (if possible at all, that would require a contribution to signal-rest-cli)

   In my opinion, implementing web hooks in `signal-cli-rest-api` is a little overkill for our current use case. However, the current implementation (dispatching a background job on every request) isn’t really a clean polling solution. Instead, we should configure a rake task or similar to poll for new Signal messages and then set up a cron job to run that task every minute or so.

phoeinx:
**Additionally to what Robert noted above, now this works, too:**

- deployment configurations, tested on staging where contributors can now register, receive requests and answer with text-based messages
- conveniently setting up new phone numbers via an ansible script

An issue is that Signal to fight spam [apparently blocks automated sending of messages (from a server)](https://github.com/AsamK/signal-cli/issues/650) to signal users. Sending only works if the user has already sent a message to the number from which we are sending messages. For us that would mean that contributors would have to send a first message to the instance signal phone number before they could receive any messages from us.

**Open questions / blockers:**
- Is it acceptable that in the signal onboarding workflow contributors would have to manually send a message to a displayed phone number? (There currently seems to be no other way.)
- How can we receive messages from the virtual phone numbers? (Something seems to not work yet.)

**What's still to do:**
- Adapt signal onboarding pages / workflow
- Support images
- Support voice messages
- Test (including with virtual phone numbers)

phoeinx:
Update re: deployment with virtual phone numbers.

I tried deploying staging with the virtual phone number dedicated for staging. Registering the number worked well and Tills signal phone number was able to receive requests as well as send answering messages with no problem. Which means, that there at least doesn't seem to be a general ban on virtual phone numbers.

Some bad news however: Sending a message from my signal phone number did not work and resulted in the following error:
photo_2021-08-23 21 19 45

This error is caused by an error when trying to pull my message from the signal-cli:
tg_image_2357588297
I'll investigate this further this week.

Good news again tho:
It seems to be possible to create links for message sending in Signal after all, which could improve the onboarding workflow.

> The link structure is as follows:
> 
> https://signal.me/#p/+15555555555
> 
> apparently you can also use:
> 
> sgnl://signal.me/#p/+15555555555
> 
> Hope that helps someone.

(See https://github.com/AsamK/signal-cli/issues/650#issuecomment-901852197)

app/adapters/signal_adapter/inbound.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+module SignalAdapter
+  UNKNOWN_CONTRIBUTOR = :unknown_contributor
+  UNKNOWN_CONTENT = :unknown_content
+
+  class Inbound
+    UNKNOWN_CONTENT_KEYS = %w[mentions attachments contacts].freeze
```
phoeinx, line 8: Where did you find the unknown content keys? We did not find any documentation, but according to this [implementation file](https://github.com/AsamK/signal-cli/blob/master/src/main/java/org/asamk/signal/json/JsonDataMessage.java) e.g. a sticker would have a dedicated sticker property.

app/adapters/signal_adapter/inbound.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    def trigger(event, *args)
+      return unless @callbacks.key?(event)
+
+      @callbacks[event].call(*args)
+    end
+
+    def initialize_sender(signal_message)
+      phone_number = signal_message.dig(:envelope, :source)
+      sender = Contributor.find_by(phone_number: phone_number)
```
phoeinx, line 43: Let's rename this to `signal_phone_number` in order to keep the naming of channel-specific attributes consistent

app/adapters/signal_adapter/inbound.rb
```diff
@@ -0,0 +1,69 @@
...
+      return sender if sender
+
+      trigger(UNKNOWN_CONTRIBUTOR, phone_number) and return nil
+    end
+
+    def initialize_message(signal_message)
+      is_receipt = signal_message.dig(:envelope, :receiptMessage)
+      return nil if is_receipt
```
phoeinx, line 51: We should probably handle other message types like `reaction`s or perhaps `remoteDelete` messages equally

app/adapters/signal_adapter/outbound.rb
```diff
@@ -0,0 +1,39 @@
...
+        number: Setting.signal_phone_number,
+        recipients: [recipient.phone_number],
+        message: text
+      }
+      req = Net::HTTP::Post.new(url.to_s, header)
+      req.body = data.to_json
+      res = Net::HTTP.start(url.host, url.port) do |http|
+        http.request(req)
+      end
+      res.value # may raise exception
```
phoeinx, line 36: When and why does this raise an exception?

phoeinx (commented):
Looks good so far, we had some points we should discuss and still need to review the testing parts.

app/jobs/signal_adapter/receive_polling_job.rb
```diff
@@ -0,0 +1,41 @@
...
+      req['Accept'] = 'application/json'
+      res = Net::HTTP.start(url.host, url.port) do |http|
+        http.request(req)
+      end
+      signal_messages = JSON.parse(res.body)
+
+      adapter = SignalAdapter::Inbound.new
+
+      adapter.on(SignalAdapter::UNKNOWN_CONTRIBUTOR) do |phone_number|
+        raise SignalAdapter::UnknownContributorError.new(phone_number: phone_number)
```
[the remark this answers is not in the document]
roschaefer: Ouch! This is true! Good spot.

This should get fixed before the PR gets merged. We want to process (and raise) errors for each message separately.

app/adapters/signal_adapter/inbound.rb
roschaefer: I just had a look into the JSON that gets parsed :laughing: and I used all the buttons that were available in my Signal app.

app/adapters/signal_adapter/inbound.rb
roschaefer: You have to distinguish between the one `signal_phone_number` of the server and many `signal_phone_number`s of each contributor. Suggestions?

ansible/roles/signal/templates/enter_captcha.j2
```diff
@@ -0,0 +1,10 @@
+#############################
+
+Visit 'https://signalcaptchas.org/registration/generate.html' and solve a captcha.
+
+Observe your JavaScript console or the url you are redirected to after successful captcha completion.
```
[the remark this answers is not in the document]
phoeinx: So you probably don't have Signal Desktop installed, right?

app/adapters/signal_adapter/inbound.rb
roschaefer: Ignore `remoteDelete` :smiling_imp: :wink:

app/adapters/signal_adapter/outbound.rb
roschaefer: I think when the service is not reachable, it will respond with a non-JSON HTTP response and then parsing it will fail.

app/jobs/signal_adapter/receive_polling_job.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def perform(*_args)
+      return if Setting.signal_phone_number.blank?
+
+      url = URI.parse("#{Setting.signal_rest_cli_endpoint}/v1/receive/#{Setting.signal_phone_number}")
+      req = Net::HTTP::Get.new(url.to_s)
+      req['Accept'] = 'application/json'
+      res = Net::HTTP.start(url.host, url.port) do |http|
+        http.request(req)
+      end
```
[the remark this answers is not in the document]
roschaefer: No, go ahead. Our services responds with JSON anyways.

app/jobs/signal_adapter/receive_polling_job.rb
```diff
@@ -0,0 +1,41 @@
...
+      signal_messages = JSON.parse(res.body)
+
+      adapter = SignalAdapter::Inbound.new
+
+      adapter.on(SignalAdapter::UNKNOWN_CONTRIBUTOR) do |phone_number|
```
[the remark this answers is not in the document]
roschaefer: That's true and wasn't anticipated by me before. Do we have to fix this before merge?

ansible/roles/signal/templates/enter_captcha.j2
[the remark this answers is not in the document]
roschaefer: I do have it installed. But I'm a little lost here, what's the problem you're facing :thinking: ?

app/adapters/signal_adapter/inbound.rb
roschaefer: What about `server_signal_phone_number` for disambiguation?

spec/components/onboarding_signal_splash_spec.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe OnboardingSignalSplash::OnboardingSignalSplash, type: :component do
+  subject { render_inline(described_class.new(**params)) }
+
+  let(:params) { {} }
+  it { should have_css('.OnboardingSignalSplash') }
```
tillprochaska, line 9: We might want to think about a better name than splash screen.

config/locales/signal/unknown_content.txt
phoeinx: Another option (which I think I would prefer) is just leaving a placeholder. The signal unknown content text is editable in the settings and every project can customize it with contact data they prefer. Additionally, if we put in an all_caps placeholder like <KONTAKTDATEN HINZUFÜGEN> people might even be more aware of having to adapt this part when setting up their own project.

ansible/roles/signal/templates/enter_captcha.j2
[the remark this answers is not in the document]
phoeinx: For me, the Signal Desktop app automatically tries to open the generated captcha link without displaying it in the Console, if the Signal Desktop app is installed. Might be different browser configurations, but shortly deinstalling Signal Desktop works as a work-around, so I think that's fine.

app/adapters/signal_adapter/inbound.rb
phoeinx: changed it to `signal_server_phone_number` for a consistent naming scheme

spec/adapters/signal_adapter/outbound_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  describe '::send!' do
+    subject { -> { described_class.send!(message) } }
+    before { message } # we don't count the extra ::send here
+    it { should_not enqueue_job }
+
+    describe 'contributor has a phone number' do
+      let(:contributor) { create(:contributor, email: nil, signal_phone_number: '+491511234567') }
+
```
phoeinx, line 28:
```suggestion
```

spec/components/onboarding_signal_splash_spec.rb
phoeinx: +1, I found the naming confusing when I first saw it. But I also don't have better ideas.
As the page serves to get Contributors to send a first message via Signal something like `OnboardingSignalMessagePage`, `OnboardingSignalMessageScreen` or `OnboardingSignalMessageView` might be fitting, but don't know if that conveys the meaning better.

spec/models/contributor_spec.rb
```diff
@@ -77,10 +77,49 @@
...
+      contributor = create(:contributor, signal_phone_number: '+49(0)/123-456789')
+      expect(contributor.signal_phone_number).to eq '+49123456789'
+    end
+
+    it 'replaces leading zeros' do
+      contributor = create(:contributor, signal_phone_number: '0049123456789')
+      expect(contributor.signal_phone_number).to eq '+49123456789'
+    end
+
+    it 'assumes German country code by default' do
```
phoeinx, line 96: maybe add a test for accepting other country codes / storing numbers with the given country code if one is given

config/locales/de.yml
```diff
@@ -398,6 +414,9 @@ de:
           has_one: Datensatz kann nicht gelöscht werden, da ein abhängiger %{record}-Datensatz
             existiert.
           has_many: Datensatz kann nicht gelöscht werden, da abhängige %{record} existieren.
+    attributes:
```
phoeinx, line 417: Where does this get used ? (Asking for my information)

spec/requests/onboarding/signal_spec.rb
```diff
@@ -0,0 +1,141 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Onboarding::Signals', type: :request do
```
phoeinx, line 5:
```suggestion
RSpec.describe 'Onboarding::Signal', type: :request do
```

phoeinx: Why do the tests run through with both versions?

app/components/onboarding_signal_splash/onboarding_signal_splash.html.erb
```diff
@@ -0,0 +1,58 @@
...
+    <p>
+      <%= t('components.onboarding_signal_splash.action', number: Setting.signal_server_phone_number).html_safe %>
+    </p>
+  <% end %>
+
+  <%= c 'stack', space: :small do %>
+    <div data-action="click->onboarding-signal-splash#highlightNext">
+      <%= c 'button',
+        styles: [:block, :primary],
+        link: "https://signal.me/#p/#{Setting.signal_server_phone_number}",
```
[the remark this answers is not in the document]
phoeinx: Works in my Signal iOS App.

spec/requests/onboarding/signal_spec.rb
tillprochaska: I think as long as you don’t use [`described_class`](https://relishapp.com/rspec/rspec-core/docs/metadata/described-class) somewhere in your tests, what you pass in to `describe` has no effect on your tests (except that it’ll be used in the console output when running the tests).

roschaefer: it changes the `subject`

tillprochaska:
Okay, I manually tested the following:

* Plain text message
* Message with image attachment
* Message with location
* Quote reply
* GIFs 🥳
* Voice message

I also tried sending a sticker message, a message containing a contact, and a PDF attachment. For all of them, a message record was created and the `unknown_content` flag was set accordingly, however, I did not receive an auto-reply.

docker-compose.yml
```diff
@@ -7,6 +7,8 @@ services:
       PORT: 3000
       POSTGRES_HOST: "${POSTGRES_HOST:-db}"
       POSTGRES_PORT: "${POSTGRES_PORT:-5432}"
+      SIGNAL_REST_CLI_ENDPOINT: "http://signal:8080"
+      SIGNAL_REST_CLI_ATTACHMENT_PATH: "signal-cli-config/attachments/"
```
tillprochaska, line 11: It’s a little confusing that this is named `SIGNAL_REST_CLI` while the project is called `signal-cli-rest-api` (inverse order of cli and rest).

spec/components/onboarding_signal_splash_spec.rb
tillprochaska: I think I’m going to rename this to `OnboardingSignalLink` -- this is what we call the respective component in the Telegram flow as well.

tillprochaska:
> Edit: Tried this again and could see the job being enqueued, but ut failed to send the message before the timeout. Not sure if I understand the log output correctly, but do jobs that are enqueued during another job run need to complete within the timeout of the parent job? I.e. would the Outbound job need to be completed within the timeout of the Receive job?

@phoeinx @roschaefer: Okay, I think the problem is related to how `sucker_punch` works. It stores jobs in memory and processes them within the application process.

> Sucker Punch is a single-process Ruby asynchronous processing library. This reduces costs of hosting on a service like Heroku along with the memory footprint of having to maintain additional jobs if hosting on a dedicated server. All queues can run within a single application (eg. Rails, Sinatra, etc.) process.

(From the [readme](https://github.com/brandonhilkert/sucker_punch))

So when we execute `bin/rails signal:receive_messages`, the receive job (and any outbound jobs to send unknown content messages) will run within that Rails processes (the one started to execute the Rake task, not the long-running process handling web requests).

The Rails process for the Rake task will exit after the task has been executed (i.e. after it has enqueued the `ReceivePollingJob`, but before the jobs have actually been executed). When the process exists, `sucker_punch`’s `shutdown_timeout` kicks in. It waits for the timeout configured, then kills all jobs that haven’t yet completed.

To summarize: The problem is that `sucker_punch` processes jobs within the parent process, which works okay for long-running processes (like the server process), but is problematic when jobs are enqueued within the context of a short-running process like a Rake task.

I could see two possible solutions:

* Enqueue the `ReceivePollingJob` from the webserver processes. (And enqueue a new job [after completion](https://edgeapi.rubyonrails.org/classes/ActiveJob/Callbacks/ClassMethods.html#method-i-after_perform).) This is a hacky solution and error prone.
* Use a different queue backend, for example [Delayed Job](https://github.com/collectiveidea/delayed_job). It uses the database to store jobs and one or multiple worker processes. Thus, processing jobs isn’t limited to the process they are enqueued in, and they are not lost if the process exists. As it uses a simple DB table to store jobs, we also wouldn’t have to maintain additional infrastructure (like Redis). Should be fine given the low volume of jobs we process.

drjakob:
I like the solution with different queue backend

tillprochaska:
@phoeinx @roschaefer FYI, I pushed a WIP commit that changes the queue backend to `delayed_job`.

phoeinx:
> @phoeinx @roschaefer FYI, I pushed a WIP commit that changes the queue backend to `delayed_job`. Please have a look at the open questions in the commit message! :)

Thank you for the research and the fixing proposal! 

Agreeing with @drjakob here; I think that using a different queueing backend makes more sense. Delayed Job also looks good to me. I like that with persisting jobs in the database we won't lose jobs if something goes wrong. The only issue that I could see would be long processing/querying times if the `delayed_job` table in the database gets too big. But as you've already mentioned, given our low job volume, that should not be a problem.

Regarding your questions in the commit message:

> Make sure we only send a single concurrent request to the signal-cli REST API.

How are you planning on achieving this? Simply check if a job with a request to the signal-cli is currently running before you enqueue a new one? That would make sense for the polling / receiving messages job. I am not sure if it would work well for jobs that send out request messages, as we would have to keep trying to enqueue them, adding error-prone logic.
Having one dedicated Signal Job Queue and one worker to process the signal jobs should also solve it, but perhaps make Signal request processing slower. (Sending only one request to the signal-cli per request/message sent out to the contributors (see my idea below might improve that.)

> As requests to the signal-cli tend to take quite long, should we use a separate queue for sending and receiving Signal messages, in order not to block sending out e.g. emails?

I think separate queues would make sense. Given that we want to poll Signal messages continuously with short intervals (like a minute or so) and the jobs themselves can take 20, 30 seconds or more, blocking actually might be an issue. Plus, setting up different queues doesn't seem to be much work with [Delayed Job](https://github.com/collectiveidea/delayed_job#running-jobs).

One more thing we might be able to do to reduce the job volume of long-running signal jobs is to adapt our request message sending for Signal. Currently, we send an individual message to each contributor who is registered with Signal, causing an individual job for each contributor. We have not tried sending a request to many Signal contributors at the same time, but if we ever end up with a relevant number of Signal contributors, that might actually block the queue for a longer time.
image
It wouldn't fit well with our current architecture for broadcasting requests, but as you can see in the `perform` method of the `SignalAdapter`, the signal-cli allows sending the same message to an Array of recipients. Which would mean one long-running request for all contributors instead of a long-running request for each contributor.

tillprochaska:
> How are you planning on achieving this? Simply check if a job with a request to the signal-cli is currently running before you enqueue a new one? That would make sense for the polling / receiving messages job. 

Yes, that would work well if we had a separate queue for polling jobs, I think. Something like this might work:

```ruby
return if Delayed::Job.where(queue: 'signal_receive').count > 0
```

That should prevent the job queue being polluted with polling jobs. Regarding outbound Signal jobs, it might just be fine not to handle them in any special way, although we might want to check wether calling the signal-cli concurrently could be a problem. But I think we wouldn’t necessarily need to do that before merging.

tillprochaska:
> One more thing we might be able to do to reduce the job volume of long-running signal jobs is to adapt our request message sending for Signal. Currently, we send an individual message to each contributor who is registered with Signal, causing an individual job for each contributor. We have not tried sending a request to many Signal contributors at the same time, but if we ever end up with a relevant number of Signal contributors, that might actually block the queue for a longer time.

👍

tillprochaska:
I think the the `signal-cli` is also looking into building a native [binary executable](https://github.com/AsamK/signal-cli#building-a-native-binary-with-graalvm-experimental). Once that is stable, it might drastically reduce the time it takes to send/receive message, as most of it seems to be due to the Java VM.

ansible/roles/signal/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+
+- name: Save container id of app
+  set_fact:
+    app_container_id: "{{ get_container_id.stdout }}"
+
+- name: Setup cronjob to receive messages
+  ansible.builtin.cron:
+    name: "Receive signal messages"
+    minute: "*" # job gets run every minute
+    job: "cd {{ ansible_facts.env.HOME }} && docker exec -it {{ app_container_id }} bin/rails signal:receive_messages"
```
tillprochaska, line 48: As the playbook is only run once during set up, this fails as soon as the container ID changes (e.g. when the container is recreated).

```suggestion
    job: "cd /home/ansible && docker-compose -f docker-compose.yml -f docker-compose.prod.yml exec app bin/rails signal:receive_message"
```

tillprochaska: Also, this currently adds a job to the root user’s crontab, we might want to change that!

tillprochaska: Okay, even with these changes, when executed by cron, the rake task isn’t run for some reason. I’ve got no time to debug this right now. Any ideas @phoeinx @roschaefer?

* It works when I invoke it manually, so I thought it might be related to the environment?

phoeinx: Oh, that makes a lot of sense, of course, and changing the playbook as you suggested makes sense to me.
Regarding the still not run rake task, I wonder why you changed `signal:receive_messages` to `signal:receive_message`? I just checked and on the latest state on this branch, the rake task is still called `receive_messages`.

tillprochaska: @phoeinx That’s a typo in the suggestion, sorry about the confusion. (I’m pretty sure I set up the job with the correct rake task, but maybe double-check that.)

tillprochaska: @phoeinx and I isolated the issue to the `docker-compose` command itself. It seems to work for me now when I use absolute paths not only for the `docker-compose` command, but also for the `docker-compose.yml` files. (Changing to `/home/ansible` before executing the command wasn’t enough.)

tillprochaska: Additionally, we need to set the `-T` flag for the `exec` command. From the docs:

> Disable pseudo-tty allocation. By default `docker-compose exec` allocates a TTY.

phoeinx:
**Status:** Polling messages on Staging works 🎉 

**What needs to be done?**

- [ ] Check that everyone currently signed in with Signal on staging (@roschaefer, @drjakob,@FrauCsu) can receive and send messages.
- [ ] Potentially only send Welcome-Message as soon as we've received an initial message from a new contributor.
- [ ] Set-up polling monitoring


_Finish implementation for our switch to a `DelayedJob` queue:_
- [ ] Add a second queue with lower priority than the standard queue for receiving / polling Signal messages and only add one polling job at a time.
- [ ] Test Sentry error reporting for the queues


_Housekeeping:_
- [ ] Rename `OnboardingSignalSplash` to `OnboardingSignalLink`
- [ ] Collect other small (non-urgent) todos into issue(s) before we close and merge this PR.

docker-compose.yml
```diff
@@ -1,19 +1,46 @@
...
+    command: bundle exec rails jobs:work
+    depends_on: [ db, signal ]
 
   db:
     image: postgres:12.2-alpine
     volumes:
       - postgres_data:/var/lib/postgresql/data
 
+  signal:
+    image: bbernhard/signal-cli-rest-api:latest
```
tillprochaska, line 36: We should pin this to the current version and set up Dependabot updates for Docker.

docker-compose.yml
```diff
@@ -1,19 +1,46 @@
...
+    SIGNAL_REST_CLI_ENDPOINT: "http://signal:8080"
+    SIGNAL_REST_CLI_ATTACHMENT_PATH: "signal-cli-config/attachments/"
+
 services:
   app:
+    <<: *x-defaults
+    command: bundle exec rails server
+    depends_on: [ db, queue ]
+
+  queue:
```
tillprochaska, line 18: We should rename this to `delayed_job`, `worker`, `queue_worker` or something like that, just for clarity.

ansible/roles/signal/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+
+- name: Query id of app container
+  command:
+    cmd: docker-compose -f docker-compose.yml -f docker-compose.prod.yml ps -q app
+    chdir: /home/ansible
+  register: get_container_id
+
+- name: Save container id of app
+  set_fact:
+    app_container_id: "{{ get_container_id.stdout }}"
```
tillprochaska, line 42: This isn’t necessary anymore as we now use docker compose

```suggestion
```

ansible/roles/installation/tasks/main.yml
```diff
@@ -44,6 +44,8 @@
     THREEMARB_API_IDENTITY: "{{ rails.threema.api_identity }}"
     THREEMARB_API_SECRET: "{{ rails.threema.api_secret }}"
     THREEMARB_PRIVATE: "{{ rails.threema.private_key }}"
+    SIGNAL_SERVER_PHONE_NUMBER: "{{ rails.signal.server_phone_number | default('') }}"
```
roschaefer, line 47: If it has a `default` it shouldn't be marked as `REQUIRED` in the `ansible/generate_config.sh`, no?

ansible/roles/signal/templates/enter_captcha.j2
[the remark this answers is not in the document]
roschaefer: My browser always asks me if I want to open a particular type of link with another application. I guess all you need is to reset that so it again asks you.

app/adapters/signal_adapter/inbound.rb
```diff
@@ -0,0 +1,117 @@
...
+      @message = initialize_message(signal_message)
+      return unless @message
+
+      files = initialize_files(signal_message)
+      @message.files = files
+
+      has_content = @message.text || @message.files.any? || @message.unknown_content
+      return unless has_content
+
+      yield(@message) if block_given?
```
roschaefer, line 37: I know I was the first to introduce this "style" of coding the adapters, however, I believe it's quite complex and hard to read. I'm a little sorry to not have it refactored earlier and see this code being repeated for other adapters :see_no_evil: :sweat_smile:

app/adapters/signal_adapter/outbound.rb
roschaefer: Well, if the server is not reachable, it will not respond at all. However, if Signal is not reachable, it will respond with a non-JSON response or, I guess, if we send a bad request.

app/components/button/button.css
```diff
@@ -15,6 +15,8 @@
   border: none;
   font-weight: 500;
   font-size: inherit;
+
+  transition: 0.1s all ease-out;
```
roschaefer, line 19: uuuuuh :fire:

bin/delayed_job
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env ruby
+
+require File.expand_path(File.join(File.dirname(__FILE__), '..', 'config', 'environment'))
+require 'delayed/command'
+Delayed::Command.new(ARGV).daemonize
```
roschaefer, line 5: What's this? :thinking:

config/application.rb
```diff
@@ -35,7 +35,7 @@ class Application < Rails::Application
     config.generators.assets = false
     config.generators.helper = false
 
-    config.active_job.queue_adapter = :sucker_punch
+    config.active_job.queue_adapter = :delayed_job
```
roschaefer, line 38: Ahh :laughing:

roschaefer: I'm sure we will regret this step. Any background job framework adds plenty of complexity. I had only bad developer experience in the past with either of these.

config/locales/de.yml
roschaefer: I think it's a default translation when a model attribute gets translated somewhere in a view file. Rails comes with this magic lookup path for attributes. It's good to investigate though where this is actually being used.

config/locales/signal/unknown_content.txt
```diff
@@ -0,0 +1 @@
+Unsere Software kann derzeit Textnachrichten, Sprachnachrichten und Bilder verarbeiten, jedoch noch keine weiteren Inhalte wie z.B. Sticker, Kontakte oder Reaktionen auf Nachrichten. In der Zwischenzeit senden Sie bitte Nachrichten mit solchen Inhalten an <KONTAKTDATEN EINFÜGEN>. Herzlichen Dank!
```
roschaefer, line 1: Alright, I'll send my stuff to `<KONTAKTDATEN EINFÜGEN>` :+1:

db/migrate/20210902094654_change_contributor_phone_number_column_name.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class ChangeContributorPhoneNumberColumnName < ActiveRecord::Migration[6.1]
+  def change
+    rename_column :contributors, :phone_number, :signal_phone_number
```
roschaefer, line 5: Hm I don't get it. What comes first? Do we migrate all `phone_number`s we had to `signal_phone_number`s? I thought one column is for internal use only (curated by users/editors) and one is the number that the user has entered.

db/migrate/20210906101108_create_delayed_jobs.rb
```diff
@@ -0,0 +1,33 @@
...
+      # Set when a client is working on this object
+      table.datetime :locked_at
+      # Set when all retries have failed (actually, by default, the record is deleted instead)
+      table.datetime :failed_at
+      # Who is working on this object (if locked)
+      table.string :locked_by
+      # The name of the queue this job is in
+      table.string :queue
+      table.timestamps null: true
+    end
```
roschaefer, line 25: At least, we don't have another dependency on e.g. redis.

docker-compose.yml
roschaefer: Yes please. And I'm already weary of the future troubles we will get from the extra service.

docker-compose.yml
```diff
@@ -15,5 +16,15 @@ services:
     volumes:
       - postgres_data:/var/lib/postgresql/data
 
+  signal:
+    image: bbernhard/signal-cli-rest-api:latest
+    environment:
+      - USE_NATIVE=0
+        #- AUTO_RECEIVE_SCHEDULE=0 22 * * * #enable this parameter on demand (see description below)
```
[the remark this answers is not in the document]
roschaefer: Yes, because I doubt this feature can process the data in the way we want it? But I would be glad about anyone who looks into this feature and sees if it has any value for us.

spec/models/contributor_spec.rb
roschaefer: This is done a few lines below, no?

roschaefer (commented):
it took a long time

Alright. I had a look into the code and you have my blessing to merge this. It looks quite nice actually. Well done @tillprochaska @phoeinx! :raised_hands: 

OOps: I opened this PR originally. I cannot click "Approve". But you have my approval. (Github needs a feature to disown a PR)

bin/delayed_job
tillprochaska: Its auto-generated when installing delayed_job, but we're not actually using it.

tillprochaska: See https://github.com/collectiveidea/delayed_job#running-jobs

db/migrate/20210902094654_change_contributor_phone_number_column_name.rb
tillprochaska: The phone column is for internal use only and to make that clearer @phoeinx renamed the phone_number column (which is used for signal) to signal_phone_number.

db/migrate/20210906101108_create_delayed_jobs.rb
tillprochaska: Yes, that is exactly the reason we chose delayed_job.

docker-compose.yml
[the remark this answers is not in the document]
tillprochaska: I checked this some time ago. The only use case is if you do not need to process incoming messages. (Signal requires that you regularly connect to the servers to receive messages even if you're not interested in them/don't receive any.)

ansible/roles/signal/templates/enter_captcha.j2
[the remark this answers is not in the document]
tillprochaska: @roschaefer: I think @phoeinx updated the message in the meantime. This is the relevant difference:

Note: You have to enable 'Persist Logs' in the Firefox console preferences or 'Preserve log' in the
network tab of Chromes console to be able to see the redirect link.

app/adapters/signal_adapter/inbound.rb
tillprochaska: Are you talking about the general structure of adapters or about the yield/block to handle incoming message? If the latter is the case, maybe we could handle messages the same way we handle any other event? That might be less confusing…

```ruby
adapter.on(SignalAdapter::MESSAGE) do |message|
  ...
end
```

config/locales/de.yml
tillprochaska: It’s used e.g. for validation error messages. For example, if a contributor enters an invalid phone number.

config/locales/de.yml
```diff
@@ -424,6 +442,9 @@ de:
...
+      contributor:
+        signal_phone_number: Telefonnummer
```
tillprochaska, line 447:
```suggestion
        signal_phone_number: Handynummer
```

app/adapters/signal_adapter/inbound.rb
roschaefer: I think the yield/block has quite a nice API, I think that all these private methods have quite some dependencies on each other, ie. they have side effects and must run in a specific order. This is quite bad I'd say.

tillprochaska (approved):

roschaefer:
party

PR #981, opened 2021-07-23
build(ci): avoid concurrent deploys

roschaefer:
See:
* https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/
* https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idconcurrency

This should fix this problem:

swappy-20210723_172605

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -31,6 +31,7 @@ jobs:
   deploy:
     needs: [docker_push]
     runs-on: ubuntu-latest
+    concurrency: staging_environment
 
     steps:
       - name: Checkout repository


tillprochaska (approved):

tillprochaska:
@roschaefer Here’s one thing to think about, but I don’t think it’s a problem. Right now, only the `deploy` job is limited to one concurrent run. However, the `docker_push` job will still run concurrently. I was wondering if we should may add the concurrency policy to the whole workflow, not just the `deploy` job.

The following could happen:

1. Start workflow at commit `123`
2. Start workflow at commit `abc`
3. `docker_push` job runs for commit `123`
4. `docker_push` job runs concurrently for commit `abc` and completes shortly after the first run, i.e. now the Docker image for commit `abc` is published.
5. `deploy` job for commit `123` runs and completes. This job uses the latest Docker image, i.e. the image built at commit `abc`.
6. `deploy` job for commit `abc` runs and completes.

I don’t really think this would be a problem, but the behavior might be unexpected. For example, imagine you’ve just merged commit `abc` and then notice a bug in that commit, so you cancel the workflow run. You’d expect that the Docker image at commit `123` would be deployed, but that’s not actually the case. As we’re only deploying to staging on push, consequences are limited though.

PR #1032, opened 2021-09-01
Filter for active/inactive contributors

tillprochaska:
Screen Shot 2021-09-01 at 18 30 41

### DIFF
--- a/app/components/tab_bar/tab_bar.css
+++ b/app/components/tab_bar/tab_bar.css
@@ -1,0 +1,36 @@
+.TabBar {
+  list-style: none;
+}
+
+.TabBar-item {
+  display: inline-block;
+  padding: var(--spacing-unit-xs) 0;
+  border-bottom: 2px solid transparent;
+  color: var(--color-text-light);
+  font-weight: 500;
+  transition: 0.1s all ease-out;
+}
+
+.TabBar-item + .TabBar-item {
+  margin-left: var(--spacing-unit-s);
+}
+
+.TabBar-item:not(.TabBar-item--active):hover {
+  border-bottom-color: var(--color-gray);
+  color: var(--color-text);
+}
+
+.TabBar-item--active {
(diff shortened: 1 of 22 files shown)


app/components/tab_bar/tab_bar.html.erb
```diff
@@ -0,0 +1,17 @@
+<nav class="TabBar">
+  <ul>
+    <% items.each do |url, options| %>
+      <li class="TabBar-item <%= 'TabBar-item--active' if options.dig(:active) %>">
+        <a href="<%= url %>">
+          <%= options.dig(:label) %>
```
roschaefer, line 6: Why `dig` :shrug: and a couple of lines below you use `[:count]`?

app/components/tab_bar/tab_bar.css
```diff
@@ -0,0 +1,36 @@
...
+}
+
+.TabBar-count {
+  display: inline-block;
+  margin-left: 0.25em;
+  padding: 0 0.75em;
+  background-color: var(--color-gray);
+  color: #fff;
+  font-size: var(--font-size-xs);
+  border-radius: 9999px;
```
roschaefer, line 35: You could fix that after https://github.com/tactilenews/100eyes/pull/1033

roschaefer (approved):
looks good

This looks very nice. I leave it open so you can add the few changes, if you want.

Feel free to merge.

app/components/tab_bar/tab_bar.html.erb
tillprochaska: The `options[:count]` is inside an `if options.key?(:count)` so that key is guaranteed to be present in that case, but I can also replace that with `dig` for consistency…

app/components/tab_bar/tab_bar.css
tillprochaska: Yes, will do that!

PR #1033, opened 2021-09-01
Use more subtle tag pills

tillprochaska:
| Contributor listings | Tags Input |
| - | - |
| Screen Shot 2021-09-08 at 14 44 42 | Screen Shot 2021-09-08 at 14 45 05 |

### DIFF
--- a/app/components/tags_input/tags_input.css
+++ b/app/components/tags_input/tags_input.css
@@ -33,6 +33,7 @@
 
   font-size: var(--font-size-s);
   border: 1px solid var(--tag-color);
+  background-color: #fff;
   border-radius: 9999px;
 
   transition: none !important;
--- a/app/components/tags_input/tags_input.css
+++ b/app/components/tags_input/tags_input.css
@@ -1,53 +1,143 @@
+/* These styles overwrite lots of Tagify's default styles
+ * in rather hacky ways. Be extra careful when changing
+ * styles and make sure to test your changes manually.
+ */
 @import '@yaireo/tagify/dist/tagify.css';
 
 .TagsInput {
   width: 100%;
 }
 
-.TagsInput tags {
-  --tag-text-color: #fff;
-
-  --tag-remove-btn-color: #fff;
-  --tag-remove-btn-bg--hover: var(--color-pink);
-  --tag-remove-bg: var(--color-pink);
-
-  --tags-border-color: var(--color-border);
-  --tags-focus-border-color: var(--color-focus);
-
+.tagify {
   border-radius: var(--border-radius);
-
   transition: none;
 }
 
-.TagsInput tags:focus-within {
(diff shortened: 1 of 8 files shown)


app/components/tags_input/tags_input.css
```diff
@@ -1,53 +1,144 @@
...
+  filter: brightness(0.66);
+
+  transition: none;
+  opacity: 1 !important;
 }
 
 .TagsInput-name {
   padding: 0 var(--spacing-unit-xs);
-  border-radius: var(--border-radius);
+  border-radius: 9999px;
```
roschaefer, line 77: :unamused:

frontend/global/variables.css
```diff
@@ -71,7 +71,7 @@
   --font-family-serif: PT Serif;
 
   --input-border-width: 1px;
-  --input-focus-shadow: 0 0 0 3px #007eff40;
+  --input-focus-shadow: 0 0 0 3px var(--color-blue-lightest);
```
roschaefer, line 74: Probably the only line I really like about this PR

app/components/tags_input/tags_input.css
```diff
@@ -1,53 +1,144 @@
...
+  transition: none !important;
+  overflow: hidden;
 }
 
-.TagsInput-dropdownItem {
-  display: flex;
-  width: 100%;
-  justify-content: space-between;
+.tagify__tag > div {
+  padding: 0.25em 0.75em;
```
roschaefer, line 44: Instead of hacking the CSS retrospectively, can we maybe configure taggify to produce the desired CSS output? 

CSS variables: https://github.com/yairEO/tagify#css-variables

SCSS variables: https://github.com/yairEO/tagify/blob/master/src/tagify.scss#L9-L24

roschaefer (approved):
if you must

What's the issue being addressed here?

I'm not a big fan of overwriting CSS from libraries. Renderless components would have come nicely here.

Maybe we could use the dedicated configuration from the library itself to update the CSS?

Anyways, I'll approve this PR as I tested it on my machine and it looks good to me.

tillprochaska:
> What's the issue being addressed here?

1. For some colors, tags are difficult to read due to a bad text/background contrast ratio.
2. @FrauCsu mentioned that some editors have described the flashy colors as annoying and/or unprofessional.

> We have been using some of the CSS variables before. However, they don’t allow for the leve of customization we need. Given that this is a pretty isolated component and we rarely need to make any changes to it, I think it’s okay to have a hacky solution.

app/components/tags_input/tags_input.css
tillprochaska: Fair: https://github.com/tactilenews/100eyes/pull/1033/commits/4ef0716d8915d845eeb40092c1a96c62955515b5

PR #1041, opened 2021-09-07
Admin dashboard

tillprochaska:


roschaefer:
According to description, this should be merged after #1035 which in turn is a draft. Shouldn't this also be a draft and get reviewed when the other PR is merged?

FrauCsu:
What's missing for merching into master? @roschaefer prio here

tillprochaska:
@FrauCsu I need to complete work on #1035 first.

tillprochaska:
Just for future reference: As this will make it easier to irreversibly delete contributors, messages, and request, we should maybe implement a pattern like this to confirm such "dangerous" actions:

Screen Shot 2021-09-30 at 16 01 01

FrauCsu:
Do not attach to #1035 @tillprochaska @roschaefer Try to re-factor and get rid of the dependency of #1035.

drjakob:
Whats blocking this issue to be merged @roschaefer @tillprochaska ?

Issue #1004, opened 2021-08-13
UI Improvements Contributor Profiles and Listing


tillprochaska:
This is a collection of multiple minor UI issues.

The contributor profiles UI has evolved in a somewhat unstructured way. The currently UI has some problems and potential for improvements:

- [x] **1) Master data vs. communication channel preferences**
   It’s unclear which fields are merely a way to store additional information about a contributor vs. which fields are used to send messages to contributors. For example, an editor reported they wanted to simply store an email address in a contributor’s profile as additional information. They did not expect that this would cause the contributor to receive all messages via email.
   
   We should probably separate the fields that store communication channel information and those that store master data that isn’t used for anything else. We should also add help texts where appropriate to set clear expectations.

||||
|-|-|-|
|![profil_email](https://user-images.githubusercontent.com/1512805/133920295-d75101b9-e543-4107-abab-1191957ac80d.png)|![profil_inaktiv](https://user-images.githubusercontent.com/1512805/133920298-3ddadce2-baa0-4724-9dda-310d848b4cea.png)|![profil_onboarding](https://user-images.githubusercontent.com/1512805/133920303-06034fc4-046d-4390-bd5c-bbd37135b07b.png)|

- [x] **2) Display note/warning if contributor hasn’t completed onboarding**
   Telegram and Signal have a two-step onboarding process where contributors need to send us an initial message. We should display a warning if a contributor hasn’t completed the last step, with instructions for editors how they can support contributors in completing the process. For example, if a contributor hasn’t completed Telegram onboarding, we’d display something like this:

   > "Dieses Mitglied hat sich per Telegram angemeldet, das Onboarding aber noch nicht abgeschlossen. Unter [diesem Link](#) findet das Mitglied eine Anleitung, um das Onboarding abzuschließen."

- [x] **3) Filter contributor listing by active/inactive contributors**
   Currently, the listings page for contributors lists all contributors, wether they are active or inactive. This is confusing, as the listings don’t indicate which contributors are active/inactive (i.e. you have to visit the contributor profile to see the status). Also, the total number of contributors includes inactive contributors, which might give you a false impression of the community size.

   ![](https://user-images.githubusercontent.com/1512805/97787897-2cd32e00-1bb5-11eb-8e2d-4753091e2080.png)
https://github.com/tactilenews/100eyes/pull/464#issuecomment-719975476

- [x] **4) Use more subtle styles for tag pills**
[...]

PR #1042, opened 2021-09-08
Update `Contributor.telegram?` to return true with uncomplete onboarding

### LINKED ISSUE #1039: Display "via Telegram" for contributors that have started, but not completed Telegram onboarding

tillprochaska:
This will display "via Telegram" for contributors who haven’t completed the last step of the Telegram onboarding process. We should also display a warning so editors are aware that the contributor hasn’t yet complete the process (see #1004).

Close #1039

### DIFF
--- a/app/models/contributor.rb
+++ b/app/models/contributor.rb
@@ -90,7 +90,7 @@ class Contributor < ApplicationRecord
   end
 
   def telegram?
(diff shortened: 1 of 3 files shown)


roschaefer (approved):
yup

PR #1062, opened 2021-09-14
Update default message stream name

tillprochaska:
Postmark now creates a default broadcasts stream named `broadcast` for every new server.


roschaefer (approved):
sure

PR #1074, opened 2021-09-18
Calculate correct count of contributors per tag

### LINKED ISSUE #1051: tag counting fails
tag counting fails. just one member with tag, but says 5. two members, says six.
Bildschirmfoto 2021-09-13 um 08 57 22

tillprochaska:
Previously, we used the precalculated value of the `taggings_count` column. However, this count includes any record tagged with a tag, i.e. also e.g. requests. We’re interested in the number of contributors with a specific tag, though.

Close #1051


roschaefer (approved):
makes sense

Issue #1028, opened 2021-08-31
Ignore existing server.pid file when starting Rails server


tillprochaska:
`rails server` fails if a `server.pid` file exists in the `tmp` directory. The pid file is created when the server starts. However, if the server is killed or crashes unexpectedly, it may not be deleted. Thus, when Docker tries to automatically restart the container, this fails as well.

Possible solutions:

* There might be a flag for `rails server` (or puma) to ignore an existing `server.pid` file (preferred).

* Otherwise, we could adjust the start command to remove `server.pid` if it exists, e.g. like this: `rm -f /app/tmp/server.pid. && bundle exec rails server`

tillprochaska:
@roschaefer The latter actually is the recommended solution in this example from the Docker docs: https://docs.docker.com/samples/rails/

> Next, provide an entrypoint script to fix a Rails-specific issue that prevents the server from restarting when a certain server.pid file pre-exists.

tillprochaska:
I also looked into the `docker-compose` config to see if there is a way to always recreate a container on restarts (because the only reason we run into this problem is because compose reuses the failed/exited container), but didn’t find anything.

PR #1075, opened 2021-09-19
fix: ignore and remove server.pid in docker env

roschaefer:
See: https://docs.docker.com/samples/rails/#define-the-project

fix: #1028

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,10 @@ RUN apk --update add \
     less \
     libsodium
 
+COPY entrypoint.sh /usr/bin/
+RUN chmod +x /usr/bin/entrypoint.sh
+ENTRYPOINT ["entrypoint.sh"]
+
 RUN mkdir -p /app
 WORKDIR /app
 CMD ["bundle", "exec", "rails", "server"]
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+set -e
+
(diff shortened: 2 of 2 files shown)


entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+# Remove a potentially pre-existing server.pid for Rails.
+rm -f /app/tmp/pids/server.pid
+
+# Then exec the container's main process (what's set as CMD in the Dockerfile).
```
tillprochaska, line 7: This is a little inaccurate, as `CMD` only sets a default command that can be overwritten (e.g. via command line parameters or `docker-compose`)…

Dockerfile
```diff
@@ -16,6 +16,10 @@ RUN apk --update add \
     less \
     libsodium
 
+COPY entrypoint.sh /usr/bin/
+RUN chmod +x /usr/bin/entrypoint.sh
```
tillprochaska, line 20: What’s the reason to apply the execute permission only after copying to the container? Why can’t we just check in the script with the correct permissions in the beginning?

tillprochaska (approved):

Dockerfile
roschaefer: Because it shouldn't be executed in a non-virtualized environment like your local machine.

entrypoint.sh
roschaefer: :thinking: err, yes? That's what we want, right?

PR #1078, opened 2021-09-20
Downgrade webpack dev server

tillprochaska:
`webpack-dev-server` 4 is not compatible with our current webpacker setup. I’d like to move away from webpacker in the near future anyways as it doesn’t really provides any additional value (and use [rails/jsbundling-rails](https://github.com/rails/jsbundling-rails) instead), but until then I think it’s fine to simply downgrade to the latest version that’s compatible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,6 @@
   "devDependencies": {
     "@arkweid/lefthook": "^0.7.6",
     "prettier": "^2.4.1",
-    "webpack-dev-server": "^4.2.1"
+    "webpack-dev-server": "^3.11.2"
   }
 }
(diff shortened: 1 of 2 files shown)


tillprochaska:
Closing in favor of #1081

tillprochaska:
@roschaefer Vielleicht können wir den in der Zwischenzeit reinmergen?

roschaefer (approved):
Err, OK, I didn't pay attention when merging dependencies?

tillprochaska:
> Err, OK, I didn't pay attention when merging dependencies?

It was my fault :D

PR #1081, opened 2021-09-20
Replace webpacker

tillprochaska:
Webpack is super slow and webpacker adds an additional level of indirection on top of Webpack’s own configuration.

Rails 7 will provide a thin integration layer for JS and CSS bundling tools as a replacement for webpacker. This layer handles only a few basic tasks, e.g. providing view helpers to include CSS and JS assets. Configuration of the actual build process etc. is completely up to the respective build tools, i.e. no more additional complexity.

Using the new tooling already works with Rails 6. This PR completely removes webpacker and replaces it with PostCSS CLI for CSS and esbuild for JS.


Dockerfile
```diff
@@ -24,17 +24,16 @@ RUN mkdir -p /app
...
 RUN gem install bundler
 RUN bundle install
 
-ENV RAILS_ENV development
-ENV NODE_ENV production
-
+ENV NODE_ENV=production
 COPY package.json yarn.lock ./
-RUN yarn install
+RUN yarn install --production=false
```
roschaefer, line 34: why?

Gemfile
```diff
@@ -49,6 +46,8 @@ group :development do
   gem 'spring'
   gem 'spring-commands-rspec'
   gem 'spring-watcher-listen', '~> 2.0.0'
+
+  gem 'foreman'
```
roschaefer, line 50: No, please do not pull in foreman again. I removed it back then because having multiple processes running in a single docker service caused troubles when debugging the app.

Procfile.dev
```diff
@@ -0,0 +1,3 @@
+web: bin/rails server --binding 0.0.0.0 --port 3000
+css: yarn build:css --watch
```
roschaefer, line 2: Fine for me to have a separate service for `css` and `js`. Would it be possible to merge the two?

app/views/layouts/application.html.erb
```diff
@@ -5,8 +5,8 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
 
-    <%= stylesheet_pack_tag 'application', media: 'all' %>
-    <%= javascript_pack_tag 'application' %>
+    <%= stylesheet_link_tag "application" %>
+    <%= javascript_include_tag "application", "data-turbo-track": "reload", defer: true %>
```
roschaefer, line 9: What's this change?

docker-compose.override.yml
```diff
@@ -23,6 +22,7 @@ services:
 
   app:
     <<: *x-defaults
+    command: bin/dev
```
roschaefer, line 25: Ugh.

package.json
```diff
@@ -1,18 +1,26 @@
 {
   "scripts": {
-    "prettier-write": "prettier --write '{app,frontend}/**/*.{js,css}'",
-    "prettier-check": "prettier --check '{app,frontend}/**/*.{js,css}'"
+    "build:css": "postcss ./app/assets/stylesheets/application.css -o ./app/assets/builds/application.css",
+    "build:js": "bin/esbuild",
```
roschaefer, line 4: To parallelize the two in one script:
```suggestion
    "build:css": "postcss ./app/assets/stylesheets/application.css -o ./app/assets/builds/application.css",
    "build:js": "bin/esbuild",
    "build": "run-p build:*",
```

With [npm-run-all](https://www.npmjs.com/package/npm-run-all). I guess it's OK to bundle the two as wer'e not going to debug these that often from within a docker container.

roschaefer (changes requested):
replace

I would absolutely :heart_eyes: if we could remove webpack for good. However, I think this needs some updates before I hit "approve" and "merge".

Dockerfile
tillprochaska: Otherwise it won’t install dev dependencies

Gemfile
tillprochaska: Which issues exactly? Having to (re-)start the watchers for JS and CSS is super annoying. Using foreman is the default setup with Rails.

app/views/layouts/application.html.erb
tillprochaska: As we’re removing webpacker, we can’t use the `*_pack_tag` helpers anymore.

Gemfile
roschaefer: The issue of `binding.pry` interactively through `docker-compose`. The keystrokes would hit foreman but how should it know which process is the receiver?

roschaefer (approved):
foreman

I removed foreman :point_up: and now he and I are happy.

roschaefer:
@tillprochaska feel free to merge

docker-compose.override.yml
```diff
@@ -28,7 +27,11 @@ services:
       - 3000:3000
       - 3035:3035
 
-  queue:
+  assets:
```
tillprochaska, line 30: @roschaefer Can we add the `assets` service to the dependencies of the `app` service so it automatically starts if you run `docker-compose up app`?

tillprochaska: @roschaefer I rebased and added this change. Please merge if you’re okay with that! :)

PR #1092, opened 2021-09-27
Fix service name in dev setup

tillprochaska:


docker-compose.override.yml
```diff
@@ -23,14 +23,13 @@ services:
...
-    depends_on: [ db, mailserver ]
+    depends_on: [ db, signal, mailserver ]
     ports:
       - 3000:3000
       - 3035:3035
 
-  queue:
+  job-worker:
     <<: *x-dev-defaults
-    depends_on: [ db ]
```
roschaefer, line 33: The job-worker does not depend on the `db`? :thinking: Am I missing something? https://github.com/collectiveidea/delayed_job#installation

roschaefer (changes requested):
Sorry @tillprochaska here's another request changes.

docker-compose.override.yml
tillprochaska: It does actually depend on the `db` and `signal` services. That’s already defined in `docker-compose.yml` though (see [this line](https://github.com/tactilenews/100eyes/blob/master/docker-compose.yml#L21)), so there’s no need to overwrite it in `docker-compose.override.yml`.

tillprochaska: The change in the dependencies for the `app` services wasn’t necessary though, as the `app` service never interacts with the `signal` service directly, it only enqueues jobs that then send requests to the `signal` service.

roschaefer (approved):
LGTM! :+1:

PR #1141, opened 2021-11-05
Emoji Picker

tillprochaska:
Emoji are often used in personal, casual text messages. The latest versions of both Windows and macOS support a native emoji picker, but not everyone knows they exist.

So instead of implementing our own emoji picker or embedding a huge library, we simply display a small hint below textareas to make our users aware of the native picker if they are on a supported platform.

This is what it looks like on macOS:

Screen Shot 2021-11-05 at 12 25 16


drjakob:
Leave this one open for another solution

phoeinx (approved):

PR #1156, opened 2021-11-19
Remove unnecessary dependencies

tillprochaska:
When reviewing Dependabot PRs, I noticed we have quite a few dependencies I think we’re not actually using. If you think that’s not actually the case for some of them, let me know!

`jbuilder`: Is used to generate JSON responses, but for the few cases we actually return JSON, we don't use it.

`byebug`: I think everyone on the team is using `pry`, right?

`web-console`: Not sure about others on the team, but I prefer setting a breakpoint an debugging from my terminal instead of the browser. If any of you are using the web console, we can of course keep it!

`selenium-webdriver`, `webdrivers`: Currently, we’re not running any tests in actual browsers.

`postcss-flexbugs-fixes`: These issues have been fixed in most actively developed browsers quite some time ago, so they are likely unnecessary for us.


phoeinx (approved):
I would like to keep the `byebug` gem (see my commit to this branch) but otherwise agree. Thanks for the housekeeping!
@roschaefer if you are also ok with the other removed dependencies I would merge this.

roschaefer (approved):
clean

PR #1158, opened 2021-11-20
Improve readability of phone number during Signal onboardinig

tillprochaska:
This is one of the subtasks from the Signal follow-up ticket.

| Before | After |
|-|-|
|Screen Shot 2021-11-20 at 12 38 50|Screen Shot 2021-11-20 at 12 38 35|


phoeinx (approved):
👍

PR #1159, opened 2021-11-20
Fix CSS gradient in two column layout in Safari

tillprochaska:
This fixes a minor rendering bug in Safari. While we don’t officially support Safari, it’s a one-line change and this has constantly annoyed me.

https://stackoverflow.com/questions/11829410/css3-gradient-rendering-issues-from-transparent-to-white

Screen Shot 2021-11-20 at 12 46 10


phoeinx (approved):
Looks nice!

PR #1160, opened 2021-11-20
Use native sidecar directories

tillprochaska:
**To do:**

- [x] Rebase
- [ ] Remove `bin/view_component_sidecar.sh`
- [ ] Update generators

***

Refactor components to use native sidecar directories

When we initially started using the `view_component` gem, there were no conventions on how to handle CSS and JS files that belong to a component. That’s why back then, I implemented that myself, following a directory structure like this:

```
components/
├─ button/
│  ├─ button.rb
│  ├─ button.css
│  ├─ button.js
│  ├─ button.html.erb
```

Today, `view_component` has built-in support for those “sidecar” directories, using a directory layout like this:

```
components/
├─ button/
│  ├─ component.rb
│  ├─ component.css
│  ├─ component.js
│  ├─ component.html.erb
```

Following the standard way also has the benefti that it will allow us to nest/namespace components, e.g. we could group all components related to auth in a `auth` directory. That would make sense as a next step, I think…

Docs: https://viewcomponent.org/guide/javascript_and_css.html


roschaefer (commented):
Well, I'm a little annoyed that every component class is now called `Component` and nested in it's namespaced module. Good luck doing a `git grep 'class Component'` but since we can still `git grep 'module SomeComponent'` I'm OK with this.

fine for me

> Following the standard way also has the benefti that it will allow us to nest/namespace components, e.g. we could group all components related to auth in a auth directory. That would make sense as a next step, I think…

I didn't know that we're not able to group components logically together. Is this true? If so it would be a good reason for the switch.

drjakob:
What happened to this pr?

tillprochaska:
> What happened to this pr?

It’s nothing urgent and Robert has a point that this might make things more complicated/might not be the optimal solution, so I didn’t continue work on this so far.

PR #1161, opened 2021-11-21
feat: order contributor by most recent activity

roschaefer:
This is just a quick and dirty hack to see how much of an effort it
would be to implement it. If I'm going down this path, I will make sure
to refactor and add some tests.


drjakob:
What happened to this one?

Issue #1055, opened 2021-09-14
Follow up: Signal Improvements


tillprochaska:
This is a follow up issues, collecting various smaller improvements and issues with the Signal integration.

- [ ] **Do not discard initial "Hello" messages by contributors.**
   In order to complete the Signal onboarding process, contributors have to send us an initial message. Currently, we discard this message. Those messages might contain important information or questions by contributors and as such, we should keep and handle them in the same way we handle any other message.

- [ ] **Hide "Zur Signal-App" button on desktop**
   The Signal desktop client currently [doesn’t support linking to a conversation/profile](https://community.signalusers.org/t/add-custom-url-scheme-signal-or-sgnl-for-share-or-call-buttons-on-websites/2362/86) and displays a confusing error messages. Therefore, we should hide the button on desktops and make the "Copy phone number" button the primary action".

- [ ] **Format phone number in an easier-to-read way**
  We should add spaces and (probably) omit the leading country code. 
  <img width="514" alt="Screen Shot 2021-09-14 at 12 21 53" src="https://user-images.githubusercontent.com/1512805/133240793-2eaec6a9-48cc-4cd1-b278-c41859ef2377.png">

PR #1167, opened 2021-11-23
Add support for Signal reaction emojis

### LINKED ISSUE #1144: signal emoji fix
Please fix signal emojis, send them as seperate message

phoeinx:
This is a PR for the Signal Follow-Up Improvement (#1055) **Handle emoji reactions** and issue **Signal emoji fix** (#1144).

@FrauCsu asked to prioritise this improvement.

image

Signal allows users to react to messages with emojis (like e.g. the heart emoji in the screenshot above). These emojis are then displayed together with the message. We previously did not support reaction emojis and sent an `UNKNOWN_CONTENT` message to contributors who used reaction emojis.

This PR now implements the following logic for supporting signal reaction emojis:

- Each reaction emoji gets stored and displayed as a separate message from the contributor containing only the reaction emoji. [1]
- If a contributor changes her or his reaction emoji the new emoji gets stored and displayed in an additional message [2]
- If a contributor deletes a reaction emoji this is not registered / handled. [3]

[1]: **Displaying reaction emoji**
  - image
  - image

[2]: **Displaying changed reaction emoji**
  - image
  - image
  
[3]: **(Not) displaying removed reaction emoji**
- image
- image

Closes #1144.

### DIFF
--- a/app/adapters/signal_adapter/inbound.rb
+++ b/app/adapters/signal_adapter/inbound.rb
@@ -6,7 +6,7 @@ module SignalAdapter
   CONNECT = :connect
 
   class Inbound
-    UNKNOWN_CONTENT_KEYS = %w[mentions contacts reaction sticker].freeze
+    UNKNOWN_CONTENT_KEYS = %w[mentions contacts sticker].freeze
     SUPPORTED_ATTACHMENT_TYPES = %w[image/jpg image/jpeg image/png image/gif audio/oog audio/aac audio/mp4].freeze
 
     attr_reader :sender, :text, :message
@@ -68,7 +68,12 @@ module SignalAdapter
 
       data_message = signal_message.dig(:envelope, :dataMessage)
 
-      message = Message.new(text: data_message[:message], sender: sender)
+      reaction = data_message[:reaction]
+      return nil if reaction && reaction[:isRemove]
+
+      message_text = reaction ? reaction[:emoji] : data_message[:message]
+
+      message = Message.new(text: message_text, sender: sender)
       message.raw_data.attach(
         io: StringIO.new(JSON.generate(signal_message)),
         filename: 'signal_message.json',
--- a/app/adapters/signal_adapter/inbound.rb
+++ b/app/adapters/signal_adapter/inbound.rb
@@ -64,12 +64,11 @@ module SignalAdapter
 
     def initialize_message(signal_message)
       is_receipt = signal_message.dig(:envelope, :receiptMessage)
-      return nil if is_receipt
+      is_remove_emoji = signal_message.dig(:envelope, :dataMessage, :reaction, :isRemove)
+      return nil if is_receipt || is_remove_emoji
 
       data_message = signal_message.dig(:envelope, :dataMessage)
-
       reaction = data_message[:reaction]
-      return nil if reaction && reaction[:isRemove]
 
       message_text = reaction ? reaction[:emoji] : data_message[:message]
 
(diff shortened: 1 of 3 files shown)


app/adapters/signal_adapter/inbound.rb
```diff
@@ -68,7 +68,12 @@ def initialize_message(signal_message)
 
       data_message = signal_message.dig(:envelope, :dataMessage)
 
-      message = Message.new(text: data_message[:message], sender: sender)
+      reaction = data_message[:reaction]
+      return nil if reaction && reaction[:isRemove]
```
roschaefer, line 72: Maybe add some consistency here? Either `return` like :point_up: or change the lines above to `return nil` as well.

app/adapters/signal_adapter/inbound.rb
```diff
@@ -68,7 +68,12 @@ def initialize_message(signal_message)
...
+
+      message_text = reaction ? reaction[:emoji] : data_message[:message]
```
roschaefer, line 74: What does rubocop say about ternary operators in ruby? Isn't it used to be written with an `if` oneliner?

roschaefer (commented):
While this looks OK for me, I forgot how to test Signal locally and I couldn't find documentation. So, I'm giving this a "comment" status.

not sure

Besides that: Maybe we could quote the original message (if it's easy to implement) in the emoji message. This way, an editor would immediately see when the emoji referenced an older mesage.

app/adapters/signal_adapter/inbound.rb
phoeinx: Rubocop didn't complain and I just looked it up - it only has rules for [Multiline ternary operators ](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MultilineTernaryOperator) in which case it requests the `if` `then` `else` constructs, but it explicitly marks `a = cond ? b : c` as good. (And in my opinion that's also more common and readable for a ternary operator with little complexity.)

app/adapters/signal_adapter/inbound.rb
phoeinx: Not a hundred percent sure what you exactly you mean (which lines should return `nil`? because there isn't really another place where we should return `nil` or?)

But I assume this is mostly about the clarity / consistency to know when nil is returned and that can definitely be improved - I pushed a new version which moves all checks for when it's appropriate to return nil to the top of the method. What do you think ?

phoeinx:
> Besides that: Maybe we could quote the original message (if it's easy to implement) in the emoji message. This way, an editor would immediately see when the emoji referenced an older mesage.

We discussed this possibility, but it would actually require quite some refactoring, so it was decided to try the easy implementation first. The assumption is that this works well for 95 % of all usages of reaction emojis - that means quick replies to the most recent question. 

The higher refactoring need lies based in that Signal identifies the message that was reacted to via a `"targetSentTimestamp"`. To match this we would need to start storing the sent timestamps for all messages we receive and also - that is the point of resistance - of messages we ourselves sent out / the response to the message send request we send to the Signal cli. Definitely doable but a bit more of effort and possibly some thought needed about where we adapt the current broadcast abstraction where we assume that response values are not needed.
```
"reaction": {
                    "emoji": "❤️",
                    "targetAuthor": "+491xxxxxxxxxx",
                    "targetSentTimestamp": 1630442783119,
                    "isRemove": false
                }
```

tillprochaska (approved):

PR #1174, opened 2021-11-25
Run dependabot checks weekly instead of daily

tillprochaska:
Running daily Dependabot checks isn't necessary. In case of critical updates, Dependabot/GitHub will notify us anyways. Most of the updates we see are from dependencies with very frequent releases and few changes per release, thus causing a lot of noise.


roschaefer (approved):
Fine for me :+1:

PR #1186, opened 2021-12-04
refactor: replace yarn with npm

roschaefer:
So, `yarn` in version is in maintenance mode since early 2020 already.
See: https://dev.to/arcanis/introducing-yarn-2-4eh1

It receives less frequent releases than `npm`:
* https://github.com/npm/cli/releases
* https://github.com/yarnpkg/yarn/releases

I guess we want to move away from `yarn` in version1.
In another project, I'm learning about the new version of `yarn` I love it.
However, this one has major changes, some of them breaking, and I guess we want
to stick to the more conservative approach here?


tillprochaska (approved):
Sorry it took so long, but I finally had a look at this. I rebased the PR, converted the current `yarn.lock@master` to `package-lock.json` via [`synp`](https://github.com/imsnif/synp) and removed the `yarn-error.log` that you committed (probably accidentally).

Anything I could’ve missed while updating this PR?

PR #1187, opened 2021-12-05
Do not load view_component engine manually

tillprochaska:
This is deprecated:

> DEPRECATION WARNING: This manually engine loading is deprecated and will be removed in v3.0.0. Remove `require "view_component/engine"`. (called from <main> at /app/config/application.rb:19)


phoeinx (approved):
Very pleasant on my eyes, thanks for taking charge.

PR #1189, opened 2021-12-07
Fix profile picture upload

tillprochaska:


phoeinx:
Profile uploading for image types works again 👍 
During testing I came across another profile picture upload issue where a failed avatar update caused render issues. (See commit message for my proposed fix.)

phoeinx (approved):

tillprochaska:
@phoeinx I’ve slightly changed the implementation of your fix to only reset the changed avatar, but keep any other changed attributes (see commit message).

phoeinx (approved):
Looks good and makes a lot more sense to only handle invalid avatar input in this way. Thanks for the update! 👍 

What I noticed during testing is that when a user uploads a new profile picture, all not-saved changes to notes and tags are lost. This might be confusing because it specifically does not happen when saving a change to the email address.
If you agree @tillprochaska we might add changing this as a low-priority possible enhancement.

Another question: In some other tests e.g. `chat_message_spec.rb:65` we used i18n to check for displayed text. Not sure if this might be unnecessary overhead especially as we don't have configs for other languages than German currently. I saw the feature specs in general don't use i18n and wanted to ask for your opinion on that @tillprochaska.

tillprochaska:
>  If you agree @tillprochaska we might add changing this as a low-priority possible enhancement.

👍 

> I saw the feature specs in general don't use i18n and wanted to ask for your opinion on that @tillprochaska.

I think we used the I18n helpers in some places because it was cumbersome to always have to update tests when there were minor wording changes etc. In general however, I found tests that make use of the I18n helpers more difficult to read and especially in the case of feature tests, it’s kind of against the idea of testing end to end.

PR #1191, opened 2021-12-10
Change Telegram ID columns to bigint

### LINKED ISSUE #1190: Handle Telegram IDs larger than 4 bytes
Telegram chat and user IDs can be larger than 4 bytes. Our database schema currently uses integer columns to store those, which results in out of range errors when trying to save respective records.

See Sentry issue 100EYES-2E.

Relevant Telegram API documentation:
* https://core.telegram.org/bots/api#user
* https://core.telegram.org/bots/api#chat

tillprochaska:
Fixes #1190


roschaefer (approved):
integer

PR #1225, opened 2022-01-24
Set human-readable server name in sentry config

### LINKED ISSUE #1192: Configure Sentry server name
By default, the container ID is attached to new Sentry events as the `server_name`. This can be a little confusing, as you’d first need to check container IDs to find out which instances are affected. We can also explicitly set `server_name` via the Sentry configuration. I think we should use something that’s easier to understand, e.g. the project name or host.

tillprochaska:
This displays a human-readable server name (based on the application host name) in Sentry that allows for easier debugging:

| Before | After |
| - | - |
| Screen Shot 2022-01-24 at 13 11 00 | Screen Shot 2022-01-24 at 13 06 17 |

Fix #1192

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -3,4 +3,5 @@
 Sentry.init do |config|
   config.breadcrumbs_logger = [:active_support_logger]
   config.capture_exception_frame_locals = true
+  config.server_name = ENV['APPLICATION_HOSTNAME'] if ENV['APPLICATION_HOSTNAME'].present?
 end


phoeinx:
Looks good (and will be much easier to read)

One question: What happens when `ENV['APPLICATION_HOSTNAME']` is not set? 

In case of it not being present `ENV['APPLICATION_HOSTNAME'] if ENV['APPLICATION_HOSTNAME'].present?` would evaluate to `nil` or? Does Sentry smartly default to the container id in this case or does the server_name not get set in this case?

tillprochaska:
> In case of it not being present ENV['APPLICATION_HOSTNAME'] if ENV['APPLICATION_HOSTNAME'].present? would evaluate to nil or?

I just tested that it works as expected even if 'APPLICATION_HOSTNAME'  is not set. I think the `if` condition applies to the whole left-hand side expression, i.e. the assignment expression would not be evaluated, and thus the default value would not be overwritten.

For example, try this:

```irb
> class Test
*   def value=(value)
*     puts 'Assignment'
*   end
* end
=> :value=

> a = Test.new
=> #<Test:0x00007f93df8d7e60>

> a.value = 1 if true
Assignment
=> 1

> a.value = 0 if false
=> nil
```

tillprochaska:
Ah, it’s actually equivalent to the following except for parsing order (which is irrelevant in this case):

https://docs.ruby-lang.org/en/3.1/doc/syntax/control_expressions_rdoc.html#label-Modifier+if+and+unless

```ruby
if ENV['APPLICATION_HOSTNAME'].present?
  config.server_name = ENV['APPLICATION_HOSTNAME']
end
```

phoeinx:
ah that makes a lot of sense. So I confused return value of the assignment with the value actually being assigned.
Thanks for the research and explanation!

phoeinx (approved):

PR #1236, opened 2022-01-28
Add about page

### LINKED ISSUE #1065: Display current commit hash
In order to quickly identify which revision of 100eyes we’re running, I’d like to be able to see the current commit hash somewhere in the app, e.g. in the footer.

tillprochaska:
This adds a simple page that displays information about the 100eyes instance. This includes the current commit hash and date (in order to easily see which version is currently deployed) and information about the communication channels set up.

The about page can be accessed via a small link in the footer.

Screen Shot 2022-01-28 at 16 06 18

Close #1065


app/components/about/about.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module About
+  class About < ApplicationComponent
+    private
+
+    def version
+      Setting.git_commit_sha[0, 8]
```
roschaefer, line 8:
```suggestion
      Setting.git_commit_sha
```
Why shortened? :shrug:

roschaefer (approved):
update

Great! I just checked the code (not tested it manually) and there is nothing important I complain about. :clap:

app/components/about/about.rb
tillprochaska: I find commit hashes a little easier to skim when they are shortened (especially as they are usually displayed shortened in tools like GitHub as well), especially as for most projects it’s very unlikely that there are collision with 8+ characters.

PR #1254, opened 2022-03-01
Make imprint customizable

tillprochaska:
Some clients would like to link to an imprint on their own website. The imprint link can now be customized via the settings. The default value is still the tactile.news imprint. Related, some clients preferred a more subtle tactile.news branding, so the logo is now removed.

new_footer

PR #1266, opened 2022-03-02
Settings form refactoring

tillprochaska:
Merge #1254 first!

This includes a small, additional and non-time-sensitive refactoring of the settings form.

### DIFF
--- a/app/components/settings_field/settings_field.rb
+++ b/app/components/settings_field/settings_field.rb
@@ -2,21 +2,21 @@
 
 module SettingsField
   class SettingsField < ApplicationComponent
-    def initialize(type:, setting:, **)
+    def initialize(type:, attr:, **)
       super
 
       @type = type
-      @setting = setting
+      @attr = attr
     end
 
     def call
-      component('field', object: Setting.new, value: Setting.send(setting), attr: setting) do |field|
+      component('field', object: Setting.new, value: Setting.send(attr), attr: attr) do |field|
         component(type, field.input_defaults)
       end
     end
 
     private
 
-    attr_reader :type, :setting
+    attr_reader :type, :attr
   end
 end
--- a/app/components/settings_field/settings_field.rb
+++ b/app/components/settings_field/settings_field.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module SettingsField
+  class SettingsField < ApplicationComponent
+    def initialize(type:, setting:, **)
+      super
+
+      @type = type
+      @setting = setting
+    end
+
+    def call
+      component('field', object: Setting.new, value: Setting.send(setting), attr: setting) do |field|
+        component(type, field.input_defaults)
+      end
+    end
+
+    private
+
+    attr_reader :type, :setting
+  end
+end
--- a/app/components/settings_form/settings_form.html.erb
+++ b/app/components/settings_form/settings_form.html.erb
@@ -18,7 +18,7 @@
     <% end %>
 
     <%= c 'stack' do %>
-      <%= c('heading', tag: :h2) { t 'setting.form.groups.additional_consent' } %>
+      <%= c('heading', tag: :h2) { t 'setting.form.groups.onboarding_additional_consent' } %>
 
       <%= c('field',
         object: Setting.new,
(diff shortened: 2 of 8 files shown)


app/components/settings_field/settings_field.rb
```diff
@@ -0,0 +1,22 @@
...
+  class SettingsField < ApplicationComponent
+    def initialize(type:, setting:, **)
+      super
+
+      @type = type
+      @setting = setting
+    end
+
+    def call
+      component('field', object: Setting.new, value: Setting.send(setting), attr: setting) do |field|
```
phoeinx, line 13: Why is the object always a new instance of the `Setting` class and not the `Setting` class itself?

And I was a bit confused as to what `Setting.send(setting)` does until I understood that `setting` refers to the name of the setting attribute that is referred to in this field. As `setting` could also describe the `Setting` which holds all `settings` / `Setting attributes` it might make it more clear to change the name of `setting` to `setting_name` or `setting_attribute`?

phoeinx (commented):
Makes the `SettingsForm` much cleaner. 👍 
I only have a small question and one suggestion for clearer / easier to understand naming.

app/components/settings_field/settings_field.rb
tillprochaska: > Why is the object always a new instance of the Setting class and not the Setting class itself?

This is a workaround that was already in place before ([see this line](https://github.com/tactilenews/100eyes/pull/1266/files#diff-bbc928fe5608c64d9a1d98217562d6b8a4fbbc5c8bc0d17335e3ae3825bdad2fL8)). I think this is due to the way the `Field` component derives the `id` and `name` attributes from the model name ([source](https://github.com/tactilenews/100eyes/blob/master/app/components/field/field.rb#L43)), as it uses ActiveRecord’s `model_name` method which is only available on ActiveRecord instances, not classes.

tillprochaska: > And I was a bit confused as to what Setting.send(setting) does until I understood that setting refers to the name of the setting attribute that is referred to in this field. As setting could also describe the Setting which holds all settings / Setting attributes it might make it more clear to change the name of setting to setting_name or setting_attribute?

I’d like to keep the component API simple and short, that’s why I went with `setting`:

```rb
<%= c('settings_field', type: :url_input, setting: :onboarding_logo) %>
<%= c('settings_field', type: :input, setting: :onboarding_byline) %>
```

How about `attr` as an alternative?

```rb
<%= c('settings_field', type: :url_input, attr: :onboarding_logo) %>
<%= c('settings_field', type: :input, attr: :onboarding_byline) %>
```

(We could also use `attribute` instead of `attr`, but we already use the abbreviated from in other places of the app.)

phoeinx: > > Why is the object always a new instance of the Setting class and not the Setting class itself?
> 
> This is a workaround that was already in place before ([see this line](https://github.com/tactilenews/100eyes/pull/1266/files#diff-bbc928fe5608c64d9a1d98217562d6b8a4fbbc5c8bc0d17335e3ae3825bdad2fL8)). I think this is due to the way the `Field` component derives the `id` and `name` attributes from the model name ([source](https://github.com/tactilenews/100eyes/blob/master/app/components/field/field.rb#L43)), as it uses ActiveRecord’s `model_name` method which is only available on ActiveRecord instances, not classes.

Makes sense, thanks for the explanation!

phoeinx: > How about attr as an alternative?

Seems like a good solution to me. It then also conforms nicely with the `field` API, where we would pass  `attr` to `attr`:

` component('field', object: Setting.new, value: Setting.send(attr), attr: attr)`

Much more readable to me. 👍

phoeinx (approved):
I like that the `SettingsForm` HTML is now much cleaner, got really a bit cluttered with the ever-growing list of Setting fields.

As commented I am happy with the renaming of `Setting.setting` to `Setting.attr` and have added a small commit with this refactoring. As you already updated the PR with the newly introduced `Setting` fields, everything looks good to go and I am gonna merge this PR. 👍

PR #1269, opened 2022-03-03
Fix checkbox size in mobile Safari

tillprochaska:
| Before | After |
| - | - |
| Before | After |


phoeinx (approved):
And the box is also aligned to the text. A joy to my eyes. ✨

PR #1288, opened 2022-03-30
Fix stale settings cache during job execution

### LINKED ISSUE #1276: Email adapter: Sender name isn’t set correctly when updated via the UI
**To reproduce:**

1. Open the 100eyes web UI.
2. Visit `/settings`.
3. Change the project name to something new.
4. Trigger any action that sends an email, e.g. send a new question.
5. The email uses the old project name as the sender.

However, this might as well affect other use cases where tasks are executed in background jobs.

**Why this happens:**

* Emails are sent in a background job, i.e. in a worker process that is different from the web server process.
* Settings aren’t cached once they’ve been fetched from the database for the first time.
* Caching using the default in-memory store is obviously not safe to use across multiple processes.

**Potential solutions:**

1. We could ensure that background jobs never never retrieve settings from the cache.
2. We could use a centralized cache store (Redis, …).

I think I’d prefer solution 1 here, although it’s a somewhat hacky solution. However, there should be only a handful of cases at max where this is relevant and the additional overhead of an  additional database request is negligible. This might be as easy as adding [a callback](https://api.rubyonrails.org/v7.0.2.3/classes/ActiveJob/Callbacks/ClassMethods.html#method-i-before_perform) to our `ApplicationJob` that always [resets the settings cache](https://github.com/huacnlee/rails-settings-cached/blob/main/lib/rails-settings/base.rb#L35) before any job is performed.

tillprochaska:
This is the simplest and most generic solution I could come up with. It’s not tested, as it’s somewhat difficult to test multi-process cache behavior, but I think this fix is quite isolated, so this might be fine?

In order to test this out locally, make sure you can reproduce the bug in your dev env following the instructions in #1276 first. Then switch to this branch and verify that the fix solves the bug.

Instead of registering a DelayedJob lifecyle callback, we could have used `ActiveJob`’s `before_perform` callback. However, this would require some code duplication in order to solve this problem for any async job (mail delivery and other jobs), as we’d have to implement the callback method in both `ApplicationJob` and a subclass of `ActionMailer::MailDeliveryJob`.

Additionally, this would clear the cache any time a job is performed – even if the job is performed synchronously. (This isn’t relevant right now, but it might be unexpected in case we ever use sync jobs in the future.)

Closes #1276

### DIFF
--- a/config/initializers/delayed_job_config.rb
+++ b/config/initializers/delayed_job_config.rb
@@ -8,3 +8,17 @@ Delayed::Worker.queue_attributes = {
 
 Delayed::Worker.max_attempts = 1
 Delayed::Worker.destroy_failed_jobs = false
+
+# We use an in-memory cache store that is *not* shared between
+# processes. In order to ensure we do not use outdated, cached
+# settings values, we need to clear the settings cache every
+# time a job is performed.
+class DelayedJobClearCachePlugin < Delayed::Plugin
+  callbacks do |lifecycle|
+    lifecycle.before(:perform) do
+      Setting.clear_cache
+    end
+  end
+end
+
+Delayed::Worker.plugins << DelayedJobClearCachePlugin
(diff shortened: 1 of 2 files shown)


phoeinx (approved):
The fix works well for me locally and I agree that in this case testing the fix would probably produce more overhead than actual use and the fix is quite isolated.
The `DelayedJob lifecyle callback` solution also seems good / more targeted than the `ActiveJob` solution. 👍

PR #1289, opened 2022-03-30
Add admin dashboard for delayed jobs

tillprochaska:
Currently, we try to execute a job at most once, but we do keep failed jobs in the database. We currently do not make use of this, but I think it would be very helpful to be able to see which jobs failed (and why), and (in the future) even to manually retry a failed job.

### DIFF
--- a/app/models/delayed_job.rb
+++ b/app/models/delayed_job.rb
@@ -1,0 +1,3 @@
+# frozen_string_literal: true
+
+class DelayedJob < Delayed::Job; end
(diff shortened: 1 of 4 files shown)


app/models/delayed_job.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+class DelayedJob < Delayed::Job; end
```
tillprochaska, line 3: This is required as administrate works best with non-namespaced models. (It is possible to create dashboard for namespaced models, however, as UI text and URLs are generated based on the model class, this is a little nicer, I think.)

PR #1290, opened 2022-03-30
Add messages dashboard

### LINKED ISSUE #1193: Admin Dashboard for Messages

tillprochaska:
In the past, there have been situations where support needed to delete individual messages.

Fix #1193


phoeinx (approved):
At least for my screen size, the messages table in the request view has weirdly breaking columns, e.g. 'Destroy split into two lines as 'Destro' and 'y'. 
But because the view will only be used for a few administrative tasks in between I think that works / is not worth the time to research a fix now.

PR #1294, opened 2022-03-30
Extract consent checkboxes into separate component

tillprochaska:
Currently, these two checkboxes (as well as the specs) are duplicated four times.


phoeinx (approved):
Makes it much nicer! 👍 

@tillprochaska I added some small additional improvements, if you could have a short look and if everything is alright merge, that would be great.

Issue #1275, opened 2022-03-24
Handle `typingMessage` Signal events


tillprochaska:
If a Signal users has typing indicators enabled, we’ll receive a respective Signal message that currently isn’t handled properly. See Sentry issue 100EYES-2V.

PR #1297, opened 2022-03-30
Handle exception in signal receive job

tillprochaska:
We’ve run into situations were `SignalAdapter::Inbound` (e.g. #1275) raises an error (and we can’t be sure this might not happen again in the future). It’s rather unlikely, but in case we did receive multiple Signal messages and the adapter raised an error for one of them, the other messages would be discarded.

This PR rescues any errors that are raised by `SignalAdapter::Inbound`, but keeps reporting them to Sentry so we’re still alerted about them.

### DIFF
--- a/spec/jobs/signal_adapter/receive_polling_job_spec.rb
+++ b/spec/jobs/signal_adapter/receive_polling_job_spec.rb
@@ -58,7 +58,7 @@ RSpec.describe SignalAdapter::ReceivePollingJob, type: :job do
         end
 
         it 'sends error to Sentry' do
-          subject.call
+          expect { subject.call }.not_to raise_error
           expect(Sentry).to have_received(:capture_exception)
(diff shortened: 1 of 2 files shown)


spec/jobs/signal_adapter/receive_polling_job_spec.rb
```diff
@@ -49,10 +49,24 @@
...
+      end
+
+      context 'if consuming the message fails' do
+        before do
+          allow(Sentry).to receive(:capture_exception).with(an_instance_of(StandardError))
+          allow_any_instance_of(SignalAdapter::Inbound).to receive(:consume).and_raise(StandardError)
+        end
+
+        it 'sends error to Sentry' do
+          subject.call
```
tillprochaska, line 61:
```suggestion
          expect { subject.call }.not_to raise_error
```

phoeinx (approved):
Good catch / fix!

PR #1299, opened 2022-03-30
Remove `spring-watcher-listen` and update `spring` to v4

tillprochaska:
`spring-watcher-listen` uses file system events (instead of polling) to reload files in the development environment, which should be a little more CPU efficient. However, `spring-watcher-listen` requires spring –– the most recent version of `spring` being v4.

spring v2 has been super slow for my setup (Docker for Mac), with initial boot times of up to 5 *minutes* (so in practice, I’ve rarely used it at all). Updating to v4 has solved this (at least for me).

Is there a noticeable performance impact when not using `spring-watcher-listen` with your setups? If not, can we remove this gem so we can update to spring v3.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,7 +32,6 @@ group :development do
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
   gem 'spring-commands-rspec'
-  gem 'spring-watcher-listen', '~> 2.0.0'
 end
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -347,9 +347,6 @@ GEM
     spring (2.1.1)
     spring-commands-rspec (1.0.4)
       spring (>= 0.9.1)
-    spring-watcher-listen (2.0.1)
-      listen (>= 2.7, < 4.0)
-      spring (>= 1.2, < 3.0)
     sprockets (4.0.3)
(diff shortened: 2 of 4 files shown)


phoeinx:
So I also haven't been using Spring previously. Testing it for me it actually doesn't work in the old version due to some exception within `spring-watcher-listen`. (no implicit conversion of Hash into Array (TypeError))
The version without `spring-watcher-listen` works well and without any noticeable performance issues. So completely fine by me to remove it.

(Just in case it gave you relevant performance improvements - did you see the [PR](https://github.com/rails/spring-watcher-listen/pull/32) that's currently open on `spring-watcher-listen` for Spring v4 support? Looks promising, the only problem is that the maintainers haven't answered in like a month. (Which might be a whole other issue on its own.))

phoeinx (approved):

phoeinx:
Re: my last comment - I discussed this with Till and with both agreed that deleting `spring-watcher-listen` and upgrading `spring` to v4 would be the best option. Especially, because `spring-watcher-listen` doesn't seem to be very maintained.

tillprochaska:
> I guess in newer versions, they don't have that dependency anymore, right?

Yes, I think Rails 7 doesn’t included `spring` by default at all (arguing that on modern dev machines, it doesn’t make much of a difference). However, in my experience when using Docker for macOS, it still speeds up many workflows quite a lot.

PR #1318, opened 2022-04-22
GDPR banner

### LINKED ISSUE #1317: Coockie-Banner
An optional coockie banner for data transfer to a third country on the onboarding page for Contributor.

tillprochaska:
This adds a banner/modal that displays a short note on the data processing 100eyes does and links to the privacy policy. The modal is only displayed on the first visit. The modal has to be enabled via the `ONBOARDING_DISPLAY_GDPR_MODAL` env variable. By default, it is disabled.

This is what it looks like on desktop and mobile devices:

| Desktop | Mobile |
| - | - |
| Desktop | Mobile

Close #1317


tillprochaska:
@phoeinx @roschaefer Please note that when reviewing, you might have to build assets yourself -- the dev setup seems to be broken and currently does not build assets automatically. I’ll put in a PR later, in the meantime, just execute the following:

```
docker-compose exec app bash
npm run build
```

app/components/modal/modal.js
```diff
@@ -0,0 +1,10 @@
+import { Controller } from '@hotwired/stimulus';
+import polyfill from 'dialog-polyfill';
+
+const STORAGE_KEY = 'gdpr_modal_state';
+
```
phoeinx, line 5:
```suggestion
```

phoeinx: Probably a leftover definition, pretty sure it's not needed here.

phoeinx (approved):
👍

Issue #450, opened 2020-10-26
Add file upload for onboarding hero image and project logo


tillprochaska:
Currently, in order to set the hero image and project logo on the onboarding pages, those image assets need to be hosted separately as the settings UI requires an url.

This means that in order to change the hero image or logo, editors need to have their own web space, upload the assets, copy the urls, and finally enter the urls in the settings UI.

This is complicated (although probably not a deal breaker) and might lead to errors (e.g. due to content security policies, non-HTTPS urls, …) and privacy concerns (editors might not be aware what it means to include resources from a third party).

PR #1323, opened 2022-04-25
feat: upload onboarding logo+header

### LINKED ISSUE #1143: store images on instance
as editor
i want to upload and store image data in myy system instead of uploading images anywhere and use links in order to set up my system easily..

roschaefer:
These TODOs are not implemented yet:

* ~Clean up existing blobs when uploading a new onboarding hero image or onboarding logo.~
* Make it possible to remove a blob and the corresponding `Setting` url altogether.

However, would say let's create a tracking issue for these two issues and make this ready for review.

The new component `setting_file` is now tightly coupled to the controller implementation - that's why I created a feature spec.

close #1143 #450


tillprochaska:
I’ll take a closer look later today, two things that already came to my mind:

* Will we migrate existing images manually? (I guess this would be fine, but I wanted to make sure that we communicate this explicitly before merging.)
* Could we store the blob ID (or similar) instead of the path on disk so that we can instantiate an ActiveStorage record from the setting? I think we’d run into problems/would need to manually migrate settings if we’d want to change the ActiveStorage driver/config, e.g. if we wanted to migrate to a managed object storage service. Also, it would be nice if we could use ActiveStorage’s built-in support for image transformations. (Users tend to upload images with inefficient encoding/suboptimal dimensions and it would be great if we could automatically generate optimized image variants.)

tillprochaska:
I have also quickly prototyped a UI for the image/file upload that would provide an enhanced UX:

image

image

drjakob:
I agree with Robert: let the cleanup be done later.
I think the simple UX should do asproposed by @tillprochaska . The instructions which file types are allowed and which size is recommended remain, right?

tillprochaska:
> The instructions which file types are allowed and which size is recommended remain, right?

The instruction will stay the same for now. In a separate step, we could introduce automatic image transformations/optimisations, so that editors do not have to worry about uploading images with specific dimensions/in a specific file format.

roschaefer:
@tillprochaska please re-review

spec/models/setting_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+RSpec.describe Setting, type: :model do
+  let(:uploaded_file) { fixture_file_upload('profile_picture.jpg') }
+  let(:blob) { ActiveStorage::Blob.create_and_upload!(io: uploaded_file, filename: uploaded_file.original_filename) }
+  subject { Setting }
+
+  %i[onboarding_logo onboarding_hero].each do |getter|
+    setter = "#{getter}="
+    blob_id_getter = "#{getter}_blob_id"
+    blob_id_setter = "#{getter}_blob_id="
```
roschaefer, line 13: Too much complexity for a test? :thinking:

FrauCsu: @phoeinx Can you review today? If not, maybe @tillprochaska ?

tillprochaska:
Ich habe schon mal kurz drauf geschaut, weiß aber nicht, ob ich heute in Ruhe Zeit finde. Ich versuch’s!

spec/models/setting_spec.rb
phoeinx: Today I won't make it, but if Till doesn't find time today, I'll try tomorrow.

tillprochaska:
Hi @roschaefer, sorry it took so long to do the final review. I have a few suggestions that I added to this PR directly in order to start working on #1334. I tried to explain the reasoning for all the changes, and would be interested to hear if they make sense for you as well.

spec/models/setting_spec.rb
tillprochaska: > Too much complexity for a test? 🤔

Using loops in test cases always feels a bit like this and I guess it would be a little easier to follow if the expectations were simply duplicated for `onboarding_logo` and `onboarding_hero`, but I’m fine with the current solution as well.

spec/requests/settings_request_spec.rb
```diff
@@ -15,15 +15,15 @@
     describe '`settting` params' do
       let(:params) { { setting: { project_name: 'Shiny new project' } } }
 
-      it { should change { Setting.project_name }.to('Shiny new project') }
+      it { will change { Setting.project_name }.to('Shiny new project') }
```
tillprochaska, line 18: Hmm, in this case, I don’t really see how "It will change…" is easier to understand than "It should change…"?

spec/models/setting_spec.rb
```diff
@@ -24,11 +24,11 @@
     describe setter do
       subject { -> { Setting.send(setter, blob) } }
       it "sets #{getter}" do
-        it_should change { Setting.send(getter) }.from(nil).to(blob)
+        will change { Setting.send(getter) }.from(nil).to(blob)
```
tillprochaska, line 27: I don’t really see how using `will` reads more natural in this case? What’s the reason against the default `expect` style that we use in other places?

```suggestion
        expect(subject.call).to change { Setting.send(getter) }.from(nil).to(blob)
```

spec/requests/settings_request_spec.rb
roschaefer: That fixes a deprecation warning in newer versions of rspec

tillprochaska (approved):

PR #1324, opened 2022-04-25
Fix gdpr modal defaults

tillprochaska:


ansible/roles/installation/tasks/main.yml
```diff
@@ -47,7 +47,7 @@
     SIGNAL_SERVER_PHONE_NUMBER: "{{ rails.signal.server_phone_number | default('') }}"
     SIGNAL_MONITORING_URL: "{{ rails.signal.monitoring_url | default('') }}"
     SENTRY_DSN: "{{ rails.sentry.dsn }}"
-    ONBOARDING_SHOW_GDPR_MODAL: "{{ rails.onboarding.show_gdpr_modal }}"
+    ONBOARDING_SHOW_GDPR_MODAL: "{{ rails.onboarding.show_gdpr_modal | default('') | string | lower }}"
```
phoeinx, line 50: Why did you add the string filter here? Makes sense as a defensive precaution, too, but I was wondering if there is a specific case you were thinking about.

phoeinx (approved):

ansible/roles/installation/tasks/main.yml
tillprochaska: `show_gdpr_modal` can be a boolean. I didn’t check if calling the `lower` filter on a boolean value might work (without explicitly converting it to a string first), but anyway, I think it’s better to cast values explicitly.

tillprochaska: So I guess we could do without the explicit conversion as it seems that `lower` will convert the value to a string if it isn’t already (using the `soft_str` method).

https://github.com/pallets/jinja/blob/main/src/jinja2/filters.py#L218

phoeinx: Makes sense! Thanks for the explanation and yeah, I think it's ok (and more stable) to do an explicit cast

PR #1325, opened 2022-04-25
Fix watch mode for npm `build:**` scripts

tillprochaska:
I think this broke when replacing yarn with npm. Unlike yarn, npm requires additional command line options for npm scripts to be separated with a double dash (`--`):

```
yarn build:css --watch
npm run build:css -- --watch
```

https://docs.npmjs.com/cli/v8/commands/npm-run-script

I have also changed the `build` script to run in default mode and added a new `dev` script that runs esbuild and postcss in watch mode as I think most people wouldn’t expect a default `build` script to run in watch mode.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -29,7 +29,7 @@ services:
       - 3035:3035
 
   assets:
-    command: npm run build
+    command: npm run dev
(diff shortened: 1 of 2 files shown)


roschaefer (approved):
:+1:

PR #1326, opened 2022-04-29
1315 request personalization

### LINKED ISSUE #1315: Individualized address in requests: Hello, (name)
As an editor
I would like to be able to address all my contributors by their name when sending out requests. Therefore I need to be able to use a button to individualize requests, displaying e.g. "Hallo ((VORNAME))"

tillprochaska:
**To do:**
- [x] Refactor server-side placeholder replacement

**Potential enhancements:**
- [ ] Add button/dropdown to insert placeholder
- [ ] Support additional placeholders
- [x] Highlight placeholders on `requests#show`
- [ ] Highlight incorrect placeholder markup / display warning
- [ ] Translate placeholders (e.g. allow `{{VORNAME}}` instead of `{{FIRST_NAME}}`)
- [ ] Use single braces for placeholders

https://user-images.githubusercontent.com/1512805/165991589-5c0f19ac-0737-46d9-9045-74ae13a3ee30.mp4

Close #1315

### DIFF
--- a/app/components/plaintext_message/plaintext_message.html.erb
+++ b/app/components/plaintext_message/plaintext_message.html.erb
@@ -1,1 +1,1 @@
-<div <%= attrs %>><%= rendered if !empty? %></div>
+<div <%= attrs %>><%= rendered.html_safe if !empty? %></div>
--- a/app/components/textarea/textarea.html.erb
+++ b/app/components/textarea/textarea.html.erb
@@ -1,5 +1,21 @@
-<div <%= attrs.slice(:class) %>>
-  <textarea <%= attrs.except(:class).merge(class: 'Textarea-input') %>><%= value %></textarea>
+<div
+  <%= attrs.slice(:class) %>
+  data-controller="textarea"
+  data-action="input->textarea#resize input->textarea#highlightPlaceholders"
+>
+  <div class="Textarea-wrapper">
+    <textarea
+      class="Textarea-input"
+      data-textarea-target="input"
+      <%= attrs.except(:class) %>
+    ><%= value %></textarea>
+
+    <div
+      class="Textarea-highlights"
+      data-textarea-target="highlights"
+      aria-hidden="true"
+    ></div>
+  </div>
 
   <% if show_emoji_picker_hint? && emoji_picker_supported? %>
     <p class="Textarea-emojiPickerHint text-small text-light">
--- a/app/components/textarea/textarea.html.erb
+++ b/app/components/textarea/textarea.html.erb
@@ -10,11 +10,13 @@
       <%= attrs.except(:class) %>
     ><%= value %></textarea>
 
-    <div
-      class="Textarea-highlights"
-      data-textarea-target="highlights"
-      aria-hidden="true"
-    ></div>
+    <% if highlight_placeholders? %>
+      <div
+        class="Textarea-highlights"
+        data-textarea-target="highlights"
+        aria-hidden="true"
+      ></div>
+    <% end %>
   </div>
 
   <% if show_emoji_picker_hint? && emoji_picker_supported? %>
--- a/app/components/textarea/textarea.html.erb
+++ b/app/components/textarea/textarea.html.erb
@@ -4,12 +4,6 @@
   data-action="input->textarea#resize input->textarea#highlightPlaceholders"
 >
   <div class="Textarea-wrapper">
-    <textarea
-      class="Textarea-input"
-      data-textarea-target="input"
-      <%= attrs.except(:class) %>
-    ><%= value %></textarea>
-
     <% if highlight_placeholders? %>
       <div
         class="Textarea-highlights"
@@ -17,6 +11,12 @@
         aria-hidden="true"
       ></div>
     <% end %>
+
+    <textarea
+      class="Textarea-input"
+      data-textarea-target="input"
+      <%= attrs.except(:class) %>
+    ><%= value %></textarea>
   </div>
 
   <% if show_emoji_picker_hint? && emoji_picker_supported? %>
--- a/spec/features/requests/personalization_spec.rb
+++ b/spec/features/requests/personalization_spec.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.feature 'Request personalization', type: :feature do
+  let(:user) { create(:user) }
+
+  context 'given two contributors'
+  before(:each) do
+    # `broadcast!` is stubbed in tests
+    allow(Request).to receive(:broadcast!).and_call_original
+
+    create(:contributor, first_name: 'Adam', last_name: 'Apfel', email: 'adam@example.org')
+    create(:contributor, first_name: 'Zora', last_name: 'Zimmermann', email: 'zora@example.org')
+  end
+
+  scenario 'sending a request with placeholders' do
+    visit new_request_path(as: user)
+
+    fill_in 'Interner Titel', with: 'Personalizes request'
+    fill_in 'Was möchtest du wissen?', with: 'Hi {{FIRST_NAME}}, how are you?'
+
+    click_button 'Frage an die Community senden'
+
+    # TODO: This isn't really what you'd do in a feature test. However, it's an easy to
+    # understand solution, given that we don't display broadcasted messages in the UI
+    expect(Message.pluck(:text)).to contain_exactly('Hi Adam, how are you?', 'Hi Zora, how are you?')
+  end
+end
--- a/app/assets/javascript/helpers/replace-placeholder.js
+++ b/app/assets/javascript/helpers/replace-placeholder.js
@@ -1,0 +1,4 @@
+export default function (text, placeholder, replacement) {
+  const regex = new RegExp(`({{\s*${placeholder}\s*}})`, 'ig');
+  return text.replace(regex, replacement);
+}
--- a/app/components/placeholder/placeholder.css
+++ b/app/components/placeholder/placeholder.css
@@ -1,15 +1,5 @@
 .Placeholder {
-  padding: 0.25em;
+  background-color: var(--color-blue-lightest);
   border-radius: var(--border-radius);
-  font-weight: 500;
-  color: var(--color-text-light);
-  background-color: rgba(0, 0, 0, 0.05);
-}
-
-.Placeholder::before {
-  content: '{{';
-}
-
-.Placeholder::after {
-  content: '}}';
+  color: var(--color-blue);
 }
--- a/app/components/plaintext_message/plaintext_message.rb
+++ b/app/components/plaintext_message/plaintext_message.rb
@@ -2,22 +2,31 @@
 
 module PlaintextMessage
   class PlaintextMessage < ApplicationComponent
-    def initialize(message: nil, **)
+    def initialize(message: nil, highlight_placeholders: false, **)
       super
 
       @message = message
+      @highlight_placeholders = highlight_placeholders
     end
 
     private
 
-    attr_reader :message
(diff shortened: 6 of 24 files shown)


app/components/textarea/textarea.html.erb
```diff
@@ -1,5 +1,23 @@
-<div <%= attrs.slice(:class) %>>
```
tillprochaska, line 1: The `Textarea` component is larger and more complex over time, including many features that might not be needed in all cases, for example:

* Autosizing depending on the length of the content
* Displaying shortcut hints
* Highlighting placeholders

I’m not sure if it’s a good idea to keep adding functionality to this component. On the other hand, it seemed to be the simplest solution to implement this.

drjakob:
Please review!

spec/features/requests/personalization_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    visit new_request_path(as: user)
+
+    fill_in 'Interner Titel', with: 'Personalizes request'
+    fill_in 'Was möchtest du wissen?', with: 'Hi {{FIRST_NAME}}, how are you?'
+
+    click_button 'Frage an die Community senden'
+
+    # TODO: This isn't really what you'd do in a feature test. However, it's an easy to
+    # understand solution, given that we don't display broadcasted messages in the UI
+    expect(Message.pluck(:text)).to contain_exactly('Hi Adam, how are you?', 'Hi Zora, how are you?')
```
mattwr18, line 27: I sympathize with this cause it's actually not easy for us to run e2e tests. Maybe, we could make these contributors threema/telegram/signal contributors and stub out the external service and expect they get called with the text we expect?

That would actually be one step further down the flow.
Also, can we make this a system spec instead of a feature spec? We talked about this after you wrote this, I guess.

app/components/plaintext_message/plaintext_message.html.erb
```diff
@@ -1 +1 @@
-<div <%= attrs %>><%= rendered if !empty? %></div>
+<div <%= attrs %>><%= rendered.html_safe if !empty? %></div>
```
mattwr18, line 1: Let's favor semantic HTML?

`rendered` is not HTML, from what I can tell, so this is not a wrapper div for some combination of allowed HTML tags, correct? If it is plain text, as is stored in `request.text`, then I guess the most semantic tag to use would be a `<p>`.

From [MDN docs on %div usage](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div#usage_notes):

> The %div element should be used only when no other semantic element (such as [%article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) or [%nav](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav)) is appropriate.

Whereas [MDN docs on %p](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p):

> The %p [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element represents a paragraph.

Since, we don't use an editor, there is always the possibility that the editor adds several paragraphs, instead of one, but it is still more semantic than a `<div>` tag.

mattwr18 (commented):
I have a few thoughts on this PR @tillprochaska ... See my comment :point_down: about the "feature" spec.

How is removing the hints part of this issue? It seems unrelated, which means I think it would be better as a separate issue/PR or if you could explain why we are removing them in relation to this ticket.

I saw several places where the `{{FIRST_NAME}}` placeholder is not replaced and is confusing. I can see that the message sent out did include the personalized text, but from the web UI, I could not. See screenshots :point_down:  

Also, we save the placeholder in the `Request.pluck(:text)`, and only replace the placeholder in the `Message`, is that intended?

As far as the improvements, I think that adding a button to add the placeholder might be a requirement before deploying to production. I think it would be much worse to send out a request with `{FIRST_NAME}}` than sending a generic one, so I'm concerned with user errors and not really having any validations that would prevent this. 

Screenshot from 2022-07-11 11-36-41
Screenshot from 2022-07-11 11-36-03
Screenshot from 2022-07-11 11-35-25

tillprochaska:
> How is removing the hints part of this issue? It seems unrelated, which means I think it would be better as a separate issue/PR or if you could explain why we are removing them in relation to this ticket.

I could have done it in a separate PR, but it would have likely created lots of merge conflicts and thus more work.

tillprochaska:
> Also, we save the placeholder in the Request.pluck(:text), and only replace the placeholder in the Message, is that intended?

Yes, that is intended, for two reasons:

1. Requests are associated to more than one contributor, thus it’s impossible to replace the placeholder with a concrete value.
2. Requests are stored as created/edited by an editor. Messages store the actual text that is sent to contributors. Storing the "raw" text for the request will allow us to implement other features in the future, e.g. creating draft or scheduled requests that can be created and edited later.

app/components/plaintext_message/plaintext_message.html.erb
tillprochaska: > rendered is not HTML, from what I can tell, so this is not a wrapper div for some combination of allowed HTML tags, correct?

It is! It will contain a `<p>` tag for every paragraph in the request text. For example, if you had a request like this:

```
Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa.

Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Donec quam felis, ultricies nec, pellentesque eu, pretium quis, sem.
```

It would be rendered like this:

```
<p>Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa.</p>

<p>Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Donec quam felis, ultricies nec, pellentesque eu, pretium quis, sem.</p>
```

Relevant part in the code base: https://github.com/tactilenews/100eyes/blob/main/app/components/plaintext_message/plaintext_message.rb#L20

spec/features/requests/personalization_spec.rb
tillprochaska: > I sympathize with this cause it's actually not easy for us to run e2e tests. Maybe, we could make these contributors threema/telegram/signal contributors and stub out the external service and expect they get called with the text we expect?

Sounds sensible, I’ll look into it and see if I can get it working within a reasonable amount of time.

> Also, can we make this a system spec instead of a feature spec? We talked about this after you wrote this, I guess.

As I said previously, I’m open to migrating from feature specs to system specs. However, I’d prefer to do that consistently instead of having a mix of both scattered across different directories. Also, I think it would be good to make sure we can run them in an actual headless browser as there isn’t much of a difference apart from that, right?

tillprochaska:
> I saw several places where the {{FIRST_NAME}} placeholder is not replaced and is confusing. I can see that the message sent out did include the personalized text, but from the web UI, I could not. See screenshots 👇

Did you make sure that the CSS and JS assets are built in your dev setup? Based on the screenshots you shared, I think this might be the issue. I just tried the exact same message and it works fine for me. 

image

spec/features/requests/personalization_spec.rb
mattwr18: > As I said previously, I’m open to migrating from feature specs to system specs. However, I’d prefer to do that consistently instead of having a mix of both scattered across different directories.

I agree that we should be consistent, but we also have finite resources and prioritization to consider. What I've done on a few teams in the past is that when we agree that we should change some established pattern, we hold each other accountable for new code so that we don't make migrating tasks bigger in the future. We then agree to carry out a reasonable amount of refactoring when we need to touch a file anyways. The hard part is finding the time to refactor all code, but in this case it's not much.

> Also, I think it would be good to make sure we can run them in an actual headless browser as there isn’t much of a difference apart from that, right?

I have a working example of running the tests with `selenium_chrome_headless` in https://github.com/tactilenews/100eyes/pull/1412. I guess this has a higher likelihood of getting merged first, so maybe we could just run them with `rack_test` till #1412 gets merged?

app/components/plaintext_message/plaintext_message.html.erb
mattwr18: Right, we use [simple_format](https://apidock.com/rails/v6.1.3.1/ActionView/Helpers/TextHelper/simple_format), I saw that, but forgot what it did.

mattwr18:
> I could have done it in a separate PR, but it would have likely created lots of merge conflicts and thus more work.

I don't like this as an argument for making unrelated changes. I would like to understand what motivates changes to the code base for our future selves. When we arbitrarily make changes without context, it makes it harder to maintain in the future.

I think an issue explaining why we want to remove hints would be appropriate. I am unaware of the motive and how we decided it should even be done.

mattwr18:
> Yes, that is intended, for two reasons:
> 
>     1. Requests are associated to more than one contributor, thus it’s impossible to replace the placeholder with a concrete value.
> 
>     2. Requests are stored as created/edited by an editor. Messages store the actual text that is sent to contributors. Storing the "raw" text for the request will allow us to implement other features in the future, e.g. creating draft or scheduled requests that can be created and edited later.

Great! Thanks for the explanation, which makes total sense

mattwr18:
> Did you make sure that the CSS and JS assets are built in your dev setup? Based on the screenshots you shared, I think this might be the issue. I just tried the exact same message and it works fine for me.

That sounds exactly like what happened. I did not spin up docker, but was just running the rails server. I'll test it out again, but I guess that must have been it.

tillprochaska:
>  I would like to understand what motivates changes to the code base for our future selves. When we arbitrarily make changes without context, it makes it harder to maintain in the future.

Just to be sure, have you seen the commit message? I try to put relevant information into commit messages for exactly that reasons - makes it easy to retrieve that info again in the future using `git blame`.

https://github.com/tactilenews/100eyes/pull/1326/commits/6acdecff4091c6290a71be6c50a0e29741bbb961

mattwr18:
> Just to be sure, have you seen the commit message? I try to put relevant information into commit messages for exactly that reasons - makes it easy to retrieve that info again in the future using git blame.

No, actually I haven't seen that commit message. I guess different developers look for this information in different ways. You wrote:

> We introduced a feature to append text blocks/canned messages (for some reason named “hints) to requests, but later found out that we didn’t have a good use case for it. We already removed it from the UI some time ago. I’m removing the backend implementation/specs now as well, because it’s unlikely we’ll need this feature again.

> We have considered adding a new feature that allows editors to quickly reply using canned replys/text blocks, but that would likely be implemented differently.

> Close https://github.com/tactilenews/100eyes/issues/389

When you say we removed it from the UI, could you link to the commit that removed it? I can see that this doesn't just remove backend implementation as we are removing code from the components themselves and also translated messages. 

It's great to know there is an actual issue that it's linked to and closes. Though, I would prefer to have the description of why we want a change to appear in the issue itself. 

In general, I don't have a problem with a PR closing more than one issue. Even though, it's not clear to me why you decided it should be closed in this PR or how it would cause merge conflicts/extra work, at least the having an explanation and issue linked to it conditions have been met.

tillprochaska:
> When you say we removed it from the UI, could you link to the commit that removed it? I can see that this doesn't just remove backend implementation as we are removing code from the components themselves and also translated messages.

Here is the PR that removed hints from the UI: https://github.com/tactilenews/100eyes/pull/189

mattwr18 (approved):
Ok, I double-checked that the issue was indeed the CSS and JS assets not being built. When running things with docker compose and all services running, the previews indeed work as expected and I was able to see the placeholder highlighting as well, which was a nice touch.

I still have concerns about the button not being part of v1, but talking to @FrauCsu, she asked that I prioritize that since it shouldn't be much work to add it and should help prevent some embarrasing messages being delivered because of incorrect synatx/placeholder text.

PR #1333, opened 2022-05-03
Add csv export functionality to admin contributor dashboard.

phoeinx:
This PR adds CSV download functionality on the admin dashboard for contributors:

image

It additionally adds `Data Processing Consented At` and `Additional Consent Given At` as displayed fields in the admin dashboard and therefore also in the downloaded CSV, as this information will be often needed in an export. The CSV file then contains all features visible in a single Contributor view of the dashboard (e.g. also `note`, `Created At` and `Updated At`).

This feature introduces a new dependency on the ['administrate-exportable'](https://github.com/SourceLabsLLC/administrate_exportable) gem. Using the gem made adding download functionality very easy. However, the gem is not widely used but appears at least somewhat maintained (after all it really has only one function and doesn't need to be updated often). Nevertheless, I had a short look through the code and did not find any apparent issues.

Adding the download functionality in the admin dashboard was the quickest way to approach #190. I am not sure if we can consider #190 closed, however, as this only allows admins to download the contributor data and not all users as described in #190. What is your opinion @FrauCsu / @drjakob ?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -80,3 +80,4 @@ gem 'sentry-ruby'
 
 # Admin
 gem 'administrate'
+gem 'administrate_exportable'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -91,6 +91,11 @@ GEM
       momentjs-rails (~> 2.8)
       sassc-rails (~> 2.1)
       selectize-rails (~> 0.6)
+    administrate_exportable (0.5.0)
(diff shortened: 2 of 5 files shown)


drjakob:
perfect

phoeinx:
Closes #190

tillprochaska (approved):

Issue #393, opened 2020-09-24
Rename `username` column


tillprochaska:
https://github.com/tactilenews/100eyes/pull/376 introduces Facebook messenger as a new channel. The `username` attribute/column was already quite ambiguous with Telegram as the only messenger implemented, but with the new Facebook related columns/attrs, I think we should rename that to something more clear, e.g. `telegram_username`.

PR #1335, opened 2022-05-06
Add contact information to contributor dashboard & CSV export.

phoeinx:
This PR adds contact information to the contributor dashboard and CSV export.

There are two beauty issues with the name of columns in the CSV export file:

image

The field holding the telegram username is only called `username` in our database instead of `telegram_username`. For the contributor view in the contributor dashboard, I was able to change the label to `Telegram Username` by specifying a label in the i18n file. The gem we use for the export (`administrate_exportable`) however doesn't use the label, leading to a somewhat unspecific `Username` column. 
If there is nothing speaking against a migration, I would propose migrating `username` to `telegram_username` as a follow-up. This would also improve clarity in our own schema and solve this issue.

That the threema ID column is not titelized is intentional of the implementation from `administrate_exportable` that treats `*_id` attributes special (not titelizing them). As I don't see a quick fix for that and don't think it is very relevant, I would propose living with it for the moment.

### DIFF
--- a/app/dashboards/contributor_dashboard.rb
+++ b/app/dashboards/contributor_dashboard.rb
@@ -9,6 +9,10 @@ class ContributorDashboard < Administrate::BaseDashboard
     last_name: Field::String,
     active: Field::Boolean,
     channels: Field::String.with_options(searchable: false),
+    email: Field::String,
+    username: Field::String,
+    signal_phone_number: Field::String,
+    threema_id: Field::String,
     note: Field::Text,
     created_at: Field::DateTime,
     updated_at: Field::DateTime,
@@ -33,6 +37,10 @@ class ContributorDashboard < Administrate::BaseDashboard
     last_name
     active
     channels
+    email
+    username
+    signal_phone_number
+    threema_id
     data_processing_consented_at
     additional_consent_given_at
     note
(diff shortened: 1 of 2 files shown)


app/dashboards/contributor_dashboard.rb
```diff
@@ -9,6 +9,10 @@ class ContributorDashboard < Administrate::BaseDashboard
     last_name: Field::String,
     active: Field::Boolean,
     channels: Field::String.with_options(searchable: false),
+    email: Field::String,
+    username: Field::String,
+    signal_phone_number: Field::String,
+    threema_id: Field::String,
```
tillprochaska, line 15: Could we also add the following attributes?

* `phone`
* `additional_email`
* `telegram_id`
* `zip_code`
* `city`

config/locales/de.yml
```diff
@@ -689,6 +689,9 @@ de:
       create: "%{model} erstellen"
       submit: "%{model} speichern"
       update: "%{model} aktualisieren"
+    label:
+      contributor:
+        username: Telegram Username
```
tillprochaska, line 694: > If there is nothing speaking against a migration, I would propose migrating username to telegram_username as a follow-up. This would also improve clarity in our own schema and solve this issue.

Yes 👍

tillprochaska: There already is an issue: #393

phoeinx: Alright, will do this as a follow up.

tillprochaska (approved):

PR #1342, opened 2022-05-19
Fix sending emails with commas in their sender name.

### LINKED ISSUE #1341: Emails fail to send if the project name / sender name contains a comma
See sentry 100EYES-3E.

A sender name containing a comma gets interpreted as a list of multiple sender names, which can be specified in the `from:` field in the email header according to [RFC 5322](https://www.rfc-editor.org/rfc/rfc5322#section-3.6.2).
To handle sender names with a comma in them correctly the comma in it needs to be escaped, e.g. by quoting the whole sender name.

phoeinx:
To comply with RFC 5322 (https://www.rfc-editor.org/rfc/rfc5322#section-3.6.2) where commas in the from: field specify a list of multiple senders, we need to escape commas in the sender name. This is necessary if a project specifies its project name and therefore the sender name, e.g. as 'Coffee, and you?'.

Closes #1341.

### DIFF
--- a/spec/adapters/postmark_adapter/outbound_spec.rb
+++ b/spec/adapters/postmark_adapter/outbound_spec.rb
@@ -74,6 +74,14 @@ RSpec.describe PostmarkAdapter::Outbound, type: :mailer do
 
         subject { message_email[:from].formatted }
         it { should eq(['TestingProject <100eyes-test-account@example.org>']) }
+
+        context 'with a comma / list separator in the project name' do
+          before do
+            allow(Setting).to receive(:project_name).and_return('TestingProject, with a comma!')
+          end
+
+          it { should eq(['"TestingProject, with a comma!" <100eyes-test-account@example.org>']) }
+        end
       end
 
       describe 'plaintext body' do
--- a/app/adapters/postmark_adapter/outbound.rb
+++ b/app/adapters/postmark_adapter/outbound.rb
@@ -62,7 +62,7 @@ module PostmarkAdapter
     end
 
     def default_from
-      "#{Setting.project_name} <#{Setting.email_from_address}>"
+      "\"#{Setting.project_name}\" <#{Setting.email_from_address}>"
     end
   end
 end


drjakob:
yes, lets test outbound_spec

phoeinx:
Thanks for the reminder! I added a regression test.

spec/adapters/postmark_adapter/outbound_spec.rb
```diff
@@ -74,6 +74,14 @@
 
         subject { message_email[:from].formatted }
         it { should eq(['TestingProject <100eyes-test-account@example.org>']) }
```
tillprochaska, line 76: I was wondering why this test doesn’t fail with that change? Shouldn’t this be `"TestingProject" <100eyes-test-account@example.org>` (with quotes)?

spec/adapters/postmark_adapter/outbound_spec.rb
```diff
@@ -74,6 +74,14 @@
...
+
+        context 'with a comma / list separator in the project name' do
+          before do
+            allow(Setting).to receive(:project_name).and_return('TestingProject, with a comma!')
+          end
```
tillprochaska, line 81: Not sure if the line break was mandated by rubocop, but in general I prefer the `{ … } ` block syntax for single-line blocks.

```suggestion
          before { allow(Setting).to receive(:project_name).and_return('TestingProject, with a comma!') }
```

spec/adapters/postmark_adapter/outbound_spec.rb
tillprochaska: Never mind, I didn’t notice that this doesn’t test for the raw `From` header, but for the `formatted` value, which parses, then converts back to a string again, which only adds quotes if needed.

https://github.com/mikel/mail/blob/22a7afc23f253319965bf9228a0a430eec94e06d/lib/mail/elements/address.rb#L47

tillprochaska (approved):

PR #1343, opened 2022-05-20
Update signal-cli-rest-api version.

phoeinx:
Since about two weeks ago registering new signal server phone numbers over the `signal-cli-rest-api` failed with a timeout.
These issues appear to have been caused by an outdated version, pushing the version from 0.42 to 0.59 (aka the latest version) resolved them.

However, this update also changed the structure of the signal message objects we receive. Mainly it adds new metainformation, e.g. the `sourceName` and a different `sticker` id format, which are not relevant for us. However, it also doesn't use empty attribute lists e.g. for `attachments: []` but instead doesn't include these attributes at all. This broke our code that depended on always being able to access an `attachments` attribute.
This PR fixes that.

Fixing the `signal-cli-rest-api` version to `0.59` as proposed in this PR would switch all current instances to the `0.59` version at the next deployment. I tested switching on staging, where also a signal phone number was already registered. This worked well. However, if we want to be cautious, we should only use the new version for new instances (as so far, no problems except for registering occured with the old version.) 

Edit: Regarding updating the `signal-cli-rest-api`: While updating to a higher version worked well, downgrading again to a lower version caused issues, which I could only resolve by completely reregistering the number by hand, meaning having to delete the phone number data directories by hand. (The downgrade happened on staging when I merged a security fix yesterday night, which lead to a new deploy on staging, again with the `0.42` `signal-cli-rest-api`).
Having to reregister a phone number on a production instance could potentially break those instances. As far as I experienced it on the `staging` system there were issues with sending messages to signal users who did not accept the number change in their client beforehand. 
Obviously, once we changed the version on `master` an accidental downgrade shouldn't be an issue. But the update does appear to be a bit brittle, what do you think?

### DIFF
--- a/app/adapters/signal_adapter/inbound.rb
+++ b/app/adapters/signal_adapter/inbound.rb
@@ -89,7 +89,7 @@ module SignalAdapter
 
     def initialize_files(signal_message)
       attachments = signal_message.dig(:envelope, :dataMessage, :attachments)
-      return [] unless attachments.any?
+      return [] unless attachments&.any?
 
       if attachments.any? { |attachment| SUPPORTED_ATTACHMENT_TYPES.exclude?(attachment[:contentType]) }
         @message.unknown_content = true
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,7 +26,7 @@ services:
       - postgres_data:/var/lib/postgresql/data
 
   signal:
-    image: bbernhard/signal-cli-rest-api:0.42
+    image: bbernhard/signal-cli-rest-api:0.59
     environment:
       - USE_NATIVE=0
         #- AUTO_RECEIVE_SCHEDULE=0 22 * * * #enable this parameter on demand (see description below)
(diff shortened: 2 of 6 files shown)


tillprochaska:
> Obviously, once we changed the version on master an accidental downgrade shouldn't be an issue. But the update does appear to be a bit brittle, what do you think?

I’m note sure if I understand what that should imply -- are there any other issues that would stop us from deploying the update?

tillprochaska:
Also this reminds me that we should probably include the `signal-cli-config` directory in our backup playbook so that we can restore old configuration without having to reregister a phone number?

phoeinx:
> Also this reminds me that we should probably include the `signal-cli-config` directory in our backup playbook so that we can restore old configuration without having to reregister a phone number?

You mean for restoring on a server/instance where the phone number was already registered, right? In this case, yes definitely!

phoeinx:
> > Obviously, once we changed the version on master an accidental downgrade shouldn't be an issue. But the update does appear to be a bit brittle, what do you think?
> 
> I’m note sure if I understand what that should imply -- are there any other issues that would stop us from deploying the update?

Just general caution as this can break things (as demonstrated on staging). I think updating should work - but I wanted to ask for your opinion because of that and maybe we try deploying on demo first.

PR #1345, opened 2022-05-22
Display raw message data in admin dashboard

### LINKED ISSUE #462: Display/Download raw message data from UI
We should have an option to show or download the raw message data for messages, similar to how most email clients allow you to open the raw email message. Right now, the only way to access raw message data is to ssh into the server, find the message record using the Rails console, copy and paste the raw data or download the message via scp or a similar tool. That takes quite a lot of time, can be down *only*  by devs and is error-prone.

Screenshot 2021-02-01 at 14 47 57

I think there are at least two use cases for this:

**1) Deciding wether unsupported contents are important, recovering unsupported contents**
We’ve had quite a few situations where actual contributors sent us emails containing unsupported contents. In many of those cases, the reasons for the "unsupported content" warning probably is caused by attached images in email signatures (e.g. company logos), contact cards (e.g. vcf files), email signatures (e.g. p7s) files.

However, in some cases, the attached files might actually be relevant to the message. While ideally 100eyes should support receiving any email attachment, for know it would already help if it was possible to open/download the raw message data. This would allow us to quickly evaluate what a problem with a message is and, if necessary, to decode the attachment from the raw email.

**2) Recovering parts of messages that have been incorrectly stripped out**
As email clients format HTML emails in lots of different ways, it happens from time to time that the logic that removes all HTML markup incorrectly also removes parts of the actual message content. In those cases, even tech-savvy editors could use that feature to quickly see the actual message contents.

tillprochaska:
This displays the raw message data in the admin dashboard in order to help debugging messages that contain unsupported content types etc.

We store the raw message data for all inbound messages as ActiveStorage attachments. administrate doesn’t support attachments out of the box. There are gems that add support, but in this case, all we need is to display a list of links to those attachments, which can be implemented without too much effort using a custom administrate field type.

Closes #462

### DIFF
--- a/app/dashboards/message_dashboard.rb
+++ b/app/dashboards/message_dashboard.rb
@@ -9,7 +9,8 @@ class MessageDashboard < Administrate::BaseDashboard
     recipient: Field::BelongsTo,
     request: Field::BelongsTo,
     created_at: Field::DateTime,
-    text: Field::Text
+    text: Field::Text,
+    raw_data: RawDataField
   }.freeze
 
   COLLECTION_ATTRIBUTES = %i[
--- a/app/fields/raw_data_field.rb
+++ b/app/fields/raw_data_field.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+require 'administrate/field/base'
+
+class RawDataField < Administrate::Field::Base; end
--- a/app/views/fields/raw_data_field/_form.html.erb
+++ b/app/views/fields/raw_data_field/_form.html.erb
@@ -1,0 +1,4 @@
+<div class="field-unit__label">
+  <%= f.label field.attribute %>
+</div>
+<div class="field-unit__field"></div>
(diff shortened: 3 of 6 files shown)


mattwr18 (approved):
Looks great @tillprochaska ... sorry it took so long to review this, I wanted to actually send and respond to requests via an actual channel rather than manually creating a message responding to a request.

I have created a new issue based on this onboarding step. See https://github.com/tactilenews/100eyes/issues/1381

I tested this by sending several different types of messages from Telegram. I don't think that I need to test it from multiple channels, but if there is something I am not thinking of, please let me know. I sent text messages, voice, gif, photo. For the gif, I received the unknown content, but could still see the raw data from the admin dashboard. For photo, there were four photos saved, each with their own file_unique_id, but the same file_id. They differ in their height/width, so I guess they relate to delivering responsive images.

tillprochaska:
> I don't think that I need to test it from multiple channels, but if there is something I am not thinking of, please let me know.

Yep, this is totally fine. There’s nothing channel-specific in this PR.

PR #1371, opened 2022-06-01
Enable contributors edit from admin dashboard

### LINKED ISSUE #1350: bugfix: do not edit contributors via admin dashboard
As Editor with Admin rights I want to manage my contributors data. Currently I see an edit button in the admin dashbboard: 
Bildschirmfoto 2022-05-30 um 12 16 40
which does not lead anywhere. 

Either I can really edit here or the button must go.

tillprochaska:
Editing contributors via the admin dashboard isn’t possible, but the edit form for contributors is accessible via an “Edit” button in the admin dashboard. This removes the respective `edit` route. As administrate automatically displays/hides action buttons (like the edit button) depending on wether a respective route is available, this will also hide the “Edit” button in the admin dashboard.

| Before | After |
| - | - |
| image | image |

Closes #1350

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -72,7 +72,7 @@ Rails.application.routes.draw do
       root to: 'users#index'
 
       resources :users
-      resources :contributors, only: %i[index show edit destroy] do
+      resources :contributors, only: %i[index show destroy] do
         get :export, on: :collection
       end
       resources :requests, only: %i[index show destroy]
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -72,7 +72,7 @@ Rails.application.routes.draw do
       root to: 'users#index'
 
       resources :users
-      resources :contributors, only: %i[index show destroy] do
+      resources :contributors, only: %i[index show edit update destroy] do
         get :export, on: :collection
       end
       resources :requests, only: %i[index show destroy]
--- a/spec/features/admin/contributors_spec.rb
+++ b/spec/features/admin/contributors_spec.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.feature 'Contributors', type: :feature do
+  context 'as admin' do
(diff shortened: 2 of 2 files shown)


mattwr18 (commented):
Updating a contributor from admin is not possible cause we did not configure the `update` action on line 75. Adding that, it is possible, so my question to you is why shouldn't an admin be able to edit a contributor? They can destroy them, should they also not be able to update them?

tillprochaska:
> Updating a contributor from admin is not possible cause we did not configure the update action on line 75. Adding that, it is possible, so my question to you is why shouldn't an admin be able to edit a contributor? They can destroy them, should they also not be able to update them?

All the relevant functionality to edit contributors is already exposed through the main application. When we introduced the admin dashboard, we added support for new models incrementally, whenever there was a need to do something that wasn’t possible yet in the main app.

That’s why we didn’t have that feature before (and why I simply removed the edit button).

I can imagine being able to update contributor data via the admin dashboard being more convenient in some situations though. But it requires some configuration (e.g. the setting up fields that should be editable etc.) and I just wanted to fix this issue.

tillprochaska:
If you'd like to, feel free to add this to the PR!

mattwr18 (approved):
Sorry @tillprochaska I paused on approving this because I saw that you added another feature spec and I would really rather that we don't add more work for ourselves to convert these to system specs later.

I've opened a `PR` to convert feature specs to system specs run with a headless browser using `cuprite` as you suggested. See https://github.com/tactilenews/100eyes/pull/1416/files

Please either approve that PR so I can merge it and update this PR before merging, or feel free to merge and I can rebase and convert it in https://github.com/tactilenews/100eyes/pull/1416/files

tillprochaska:
Thanks @mattwr18, I’ve seen your PR and skimmed the changes. I wanted to look at some of the changes you added comments about and try it out in my dev environment. Will do that before you return from vacation!

PR #1372, opened 2022-06-01
Display phone number in profiles of contributors with incomplete onboarding

### LINKED ISSUE #1349: Show Signal Number in case of failed registration
As an editor, I cannot see the signal number of the registered contributor if they have not completed the registration. That would be helpful to assist.

tillprochaska:
This displays the phone number used for Signal onboarding in the contributors profile, so editors can contact them to assist with finishing the onboarding process.

image

Fixes #1349

### DIFF
--- a/app/components/contributor_signal_settings/contributor_signal_settings.html.erb
+++ b/app/components/contributor_signal_settings/contributor_signal_settings.html.erb
@@ -18,6 +18,7 @@
             name: contributor.name,
             first_name: contributor.first_name,
             date: l(contributor.created_at.to_date),
+            phone_number: contributor.signal_phone_number.phony_formatted,
           ) %>
         </p>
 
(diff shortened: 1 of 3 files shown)


mattwr18 (approved):
Looks good! At first, I thought `phony_formatted` was a typo, but after looking into it, I can see that it comes from the `phony_rails` gem. 

I pulled it down and started the onboarding process with Signal, be left it incomplete. I then was able to see the message from the `/contributors/:id` endpoint.

The `.incomplete.text` is the same for `Signal` and `Telegram` besides the channel name. It might be nice to unify them. For example, why is it useful to know the number of someone who has an incomplete onboarding in Signal, but not from Telegram?

I guess we don't have the phone number for our Telegram contributors just their ids so we can't display them...

tillprochaska:
> The .incomplete.text is the same for Signal and Telegram besides the channel name. It might be nice to unify them. For example, why is it useful to know the number of someone who has an incomplete onboarding in Signal, but not from Telegram?
> 
> I guess we don't have the phone number for our Telegram contributors just their ids so we can't display them...

Yes, exactly!

PR #1373, opened 2022-06-01
Upcase Threema IDs

### LINKED ISSUE #1347: Threema IDs need to be all uppercase
Threema IDs consist of characters 0-9A-Z and key lookups will fail for Threema IDs with lowercase characters. 100eyes should handle those cases gracefully:

* Automatically uppercase Threema IDs before storing them into the database.
* Convert any existing Threema IDs to be all uppercase.

or

* Convert Threema IDs to uppercase before sending messages.
* Ensure contributor lookups by Threema ID (when receiving Threema messages) are case-insensitive.

Additionally, we might consider validating Threema IDs against Threema Gateway API during onboarding.

tillprochaska:
This implements case-insenstive Threema IDs using the second suggested solution. I went with this approach, as it doesn’t require data migrations and it seemed more robust to me and the channel specific requirements are implemented in the adapter (and for example not in a `before_save` callback in the `Contributor` model). Not sure though if it’s the *best* solution.

Fixes #1347

### DIFF
--- a/spec/adapters/threema_adapter/outbound_spec.rb
+++ b/spec/adapters/threema_adapter/outbound_spec.rb
@@ -4,18 +4,17 @@ require 'rails_helper'
 
 RSpec.describe ThreemaAdapter::Outbound do
   let(:adapter) { described_class.new }
-  let(:threema) { instance_double(Threema) }
-  let(:contributor) { create(:contributor, threema_id: 'V5EA564T', email: nil) }
+  let(:threema_id) { 'V5EA564T' }
+  let(:contributor) { create(:contributor, threema_id: threema_id, email: nil) }
   let(:message) { create(:message, recipient: contributor) }
-
-  before do
-    allow(Threema).to receive(:new).and_return(threema)
-  end
 
   describe '::send!' do
     before { message } # we don't count the extra ::send here
+
     subject { -> { described_class.send!(message) } }
+
     it { should enqueue_job(described_class) }
+
     context 'contributor has no threema_id' do
       let(:contributor) { create(:contributor, threema_id: nil, email: nil) }
       it { should_not enqueue_job(described_class) }
@@ -24,7 +23,9 @@ RSpec.describe ThreemaAdapter::Outbound do
 
   describe '::send_welcome_message!' do
     subject { -> { described_class.send_welcome_message!(contributor) } }
+
     it { should enqueue_job(described_class) }
+
     context 'contributor has no threema_id' do
       let(:contributor) { create(:contributor, threema_id: nil, email: nil) }
       it { should_not enqueue_job(described_class) }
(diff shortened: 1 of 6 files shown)


spec/adapters/threema_adapter/outbound_spec.rb
```diff
@@ -24,29 +23,40 @@
...
 
   describe '#perform' do
-    subject { adapter.perform(text: message.text, recipient: message.recipient) }
+    subject { -> { adapter.perform(text: message.text, recipient: message.recipient) } }
 
     it 'sends the message' do
-      expect(threema).to receive(:send).with({ type: :text, threema_id: contributor.threema_id, text: message.text })
-
-      subject
+      expect_any_instance_of(Threema).to receive(:send).with(type: :text, threema_id: 'V5EA564T', text: message.text)
```
tillprochaska, line 39: It’s recommended not to use `expect_any_instance_of` and `allow_any_instance_of` as it’s an indicator for design smells.

Previously, we worked around this by stubbing the `new` method to always return an instance double. However, for some reason this failed as soon as I tried to use this approach in multiple test cases with the following error message:

> Double was originally created in one example but has leaked into another example and can no longer be used. rspec-mocks' doubles are designed to only last for one example, and you need to create a new one in each example you wish to use it for.

So I did what the message recommended and initialized a new double for each test case, which didn’t change anything though.

As stubbing `new` doesn’t really solve the underlying design issue (the reason why using `*_any_instance_of` is discouraged), I decided to use `expect_any_instance_of` instead, which doesn’t lead to leaks.

I’d still be interested in cleaner solutions, though.

mattwr18: I also have read about how the use of `allow_any_instance_of` is discouraged. So, I would like to understand why this failure is happening and find a better solution.

I pulled down the code and did the same as you and ran into the same issue you did. I searched SO and found lots of ideas, one of them [rspec-mocks' doubles are designed to only last for one example](https://stackoverflow.com/questions/27088250/rspec-mocks-doubles-are-designed-to-only-last-for-one-example)](https://stackoverflow.com/questions/27088250/rspec-mocks-doubles-are-designed-to-only-last-for-one-example) suggested something that helped get to the bottom of the issue.

If you look at the [ThreemaAdapter](https://github.com/tactilenews/100eyes/blob/master/app/adapters/threema_adapter/outbound.rb#L7-L9), you can see that we are memoizing @threema_instance, so that when you call `self.class.threema_instance.send(...)` multiple times instead of using a new instance of `Threema`, it uses the memoized version. This might, or might not, be desirable for our application, but it makes testing it difficult.

It's not really clear to me what we gain from memoizing `Threema.new` in this way. The code was added while solving an issue that is unrelated and there is no explanation. See https://github.com/tactilenews/100eyes/pull/930/commits/8ad5c6f264c707262b90b5b4cecfca1ce6d55ba5.

I understand that what we want to achieve here is that for every threema recipient we don't set up a new instance of `Threema`, but I guess I'm doubtful if we gain a lot from that. Do you have any thoughts on that? @roschaefer you added this code, could you explain the benefit to me?

spec/adapters/threema_adapter/outbound_spec.rb
```diff
@@ -24,29 +23,40 @@
...
+
+  describe '#welcome_message' do
+    subject { described_class.welcome_message }
+
+    it 'strips whitespace to not break basic formatting' do
+      allow(Setting).to receive(:onboarding_success_heading).and_return(" \n text with leading and trailing whitespace \t \n ")
+      allow(Setting).to receive(:onboarding_success_text).and_return("\nSuccess text.\n")
+      is_expected.to eq("*text with leading and trailing whitespace*\n\nSuccess text.\n")
+    end
+  end
```
mattwr18, line 61: Is this test block unrelated? It seems to me that we didn't ask @roschaefer to add a test for this in the `PR` that introduced it so we are sneaking it in here?

If so, I generally prefer to avoid this and create a separate micro PR whose objective is to add missing tests, but am open to discussing this with the team.

mattwr18 (approved):
I'm going to approve this because I don't want to block the feature, but I had some thoughts about the `allow_any_instance_of` doubt you raised and the underlying code it tests. Please see my comments.

spec/adapters/threema_adapter/outbound_spec.rb
tillprochaska: Thanks for having a closer look into it! If we can find a closer solution to that without too much effort, I’m all in for it. If it turns out to be something that would take more time, I think we should balance that with the value we gain from it, especially as it’s something that we touch rarely.

mattwr18: A quick solution would be to revert the change that memoizes the instance of `Threema`. I am not sure we gain a lot from memoizing here since creating an instance of the class can't be an expensive action, but of course I might be missing the larger purpose @roschaefer wanted to achieve by memomizing in the first place.

spec/adapters/threema_adapter/outbound_spec.rb
mattwr18: Oh, sorry... When I apply `Hide whitespaces` I can see that these are not changes you added... :blush:

PR #1375, opened 2022-06-07
Add examples to create/migrate db during setup

mattwr18:
- [ ] add example to run the tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ For development purposes, you can set up `100eyes` using Docker. To perform the 
 
     ```bash
     $ cd ~/100eyes
-    $ docker-compose build
+    $ docker compose build
     ```
 
 3. #### Configuration
@@ -44,15 +44,29 @@ For development purposes, you can set up `100eyes` using Docker. To perform the 
     You can start all containers with a single command. `100eyes` can then be viewed in your browser at http://localhost:3000.
 
     ```bash
-    $ docker-compose up
+    $ docker compose up
     ```
 
-5. #### Seed database
+5. #### Create database
+
+   The database is not created in the steps above because it is a one-time task, whereas building and starting the services should be reproducible. Create the database with:
+
+   ```bash
+   $ docker exec 100eyes-app-1 rails db:create
+   ```
+
+6. #### Run the migrations
+
+   ```bash
+   $ docker exec 100eyes-app-1 rails db:migrate
+   ```
+
+7. #### Seed database
 
    One user has been created to use in development, since registration is disabled. You can create your own user in the `rails console` if you want, or you can run:
 
    ```bash
-   $ docker-compose exec app rails db:seed
+   $ docker exec 100eyes-app-1 rails db:seed
    ```
 
   ##### Webpack Development Server
@@ -60,7 +74,7 @@ For development purposes, you can set up `100eyes` using Docker. To perform the 
   With this command you can speed up the compilation of the assets for development purposes (e.g. for many changes to CSS files):
 
   ```bash
-  $ docker-compose exec app bin/webpack-dev-server
+  $ docker exec 100eyes-app-1 bin/webpack-dev-server
   ```
 
 #### Messengers
--- a/README.md
+++ b/README.md
@@ -40,42 +40,31 @@ For development purposes, you can set up `100eyes` using Docker. To perform the 
 
    Edit the new `.env` file and add API keys and other secrets there. This file `.env` is ignored by `git`.
 
-4. #### Start container
+
+4. #### Create, migrate, seed database
+
+   The database is not set up in the steps above because it is a one-time task, whereas building and starting the services should be reproducible. Setup the database with:
+
+   ```bash
+   $ docker compose run app rails db:setup
+   ```
+
+5. #### Start container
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,7 +28,7 @@ For development purposes, you can set up `100eyes` using Docker. To perform the
 
     ```bash
     $ cd ~/100eyes
-    $ docker-compose build
+    $ docker compose build
```
mattwr18, line 31: `docker-compose` is part of v1, which uses python, and has been deprecated. It has been migrated to Go in v2 and now is `docker compose`. See [docker docs](https://docs.docker.com/compose/#compose-v2-and-the-new-docker-compose-command) for more info.

README.md
```diff
@@ -44,23 +44,37 @@ For development purposes, you can set up `100eyes` using Docker. To perform the
...
+   $ docker exec 100eyes-app-1 rails db:migrate
+   ```
+
+7. #### Seed database
 
    One user has been created to use in development, since registration is disabled. You can create your own user in the `rails console` if you want, or you can run:
 
    ```bash
-   $ docker-compose exec app rails db:seed
+   $ docker exec 100eyes-app-1 rails db:seed
```
mattwr18, line 69: the services have been renamed at some point in the past and the docs have not been updated, resulting in an error 
```bash
Error: No such container: app
```

tillprochaska: I think you might have forgotten `compose` in there. You can use `docker compose exec` to run a command inside a compose service, while you’d have to use the container name or id with `docker exec`.

So in this case the following should work just fine:

```
docker compose exec app rails db:seed
```

README.md
```diff
@@ -44,23 +44,37 @@ For development purposes, you can set up `100eyes` using Docker. To perform the
...
+   $ docker exec 100eyes-app-1 rails db:create
+   ```
+
+6. #### Run the migrations
+
+   ```bash
+   $ docker exec 100eyes-app-1 rails db:migrate
+   ```
+
+7. #### Seed database
```
tillprochaska, line 64: I think creating the database, running migration and seeds would need to go before running `docker compose up`, i.e.

1. Clone repo
2. Building container
3. Configuration
4. Create, migrate, seed
5. Start containers

Instead of running `rails db:create`, `rails db:migrate`, and `rails db:seed`, you should also be able to run `rails db:setup`. It’s basically the same, but I think it doesn’t run all migration, instead loads the schema from `schema.rb`, so it’ll be a little faster when setting up a dev environment from scratch.

README.md
mattwr18: Cool, I hadn't realized this referenced the service. I'll update it.

README.md
mattwr18: I thought so as well, but when I attempted to do that, I got an error. I'll see if I can reproduce it, maybe it was due to running the command with `docker` instead of `docker compose`.

Cool, yeah I had forgotten about the `db:setup` command, but that's a nice shortcut.

tillprochaska: Okay, cool! If I find the time, I’ll also try going through the setup instructions with fresh containers/volumes.

mattwr18: the reason it was failing is because I ran the command with `exec`, but without the containers running it needs to be run with `run`. See 05a8ac68

README.md
mattwr18: Weird thing is that running `docker compose up SERVICE` starts all services, but `docker-compose up SERVICE` starts only the service you want to start. For example, you want to use your local db/rails server, but want to start the `mailserver`.... I checked the [docker compose up docs](https://docs.docker.com/engine/reference/commandline/compose_up) and it seems to indicate that it should be possible to start a single service, but it doesn't work for me :shrug:

README.md
```diff
@@ -40,28 +40,31 @@ For development purposes, you can set up `100eyes` using Docker. To perform the
...
+
+In Github Actions, we run the tests with the local environment because setting up the containers and running them there is more expensive in terms of time. You can run the tests in the containers, if you don't want to get set up locally with:
+
+```bash
+$ docker compose exec app bundle exec rspec
+```
 
-  ```bash
-  $ docker-compose exec app bin/webpack-dev-server
-  ```
```
mattwr18, line 64: Running this command results in the error:

> OCI runtime exec failed: exec failed: unable to start container process: exec: "bin/webpack-dev-server": stat bin/webpack-dev-server: no such file or directory: unknown

Looking into it a bit further, I notice that we completely removed webpacker in https://github.com/tactilenews/100eyes/pull/1081, but forgot to remove the docs

tillprochaska (approved):
Great, thanks! I think we could also remove the outdated "Production" section and refer to the Ansible playbooks instead.

mattwr18:
@tillprochaska sorry, I should have opened this as a draft PR cause there are still steps that I think need to be documented as part of getting set up with the project.... This will be a living PR for a little while, then I'll ask for another review.

mattwr18:
@tillprochaska I just merged as is. I went back and had a look at the production section and there is nothing outdated about it. It just says what ruby/postgres version we use.

I think adding a section linking to Ansible could be a good idea, especially since we have another README there already. However, I think we could invest a little time when we onboard the next person.

Issue #1287, opened 2022-03-29
Browser-Push-Notification


drjakob:
As an editor, I would like to have every new answers to a question displayed as a push message on the desktop in my browser. Then I would no longer miss anything and could react even more immediately.

tillprochaska:
I think we should consider splitting this into two separate issues:

1. Creating/storing notifications for specific events (e.g. new replies by contributors, completed contributor onboarding). [`noticed`](https://github.com/excid3/noticed) is quite popular and allows to store notifications in the database as well as to deliver them via multiple built-in channels (email, Slack, …) This would also allow us to display a feed of recent activity in the Dashboard etc.

   ![image](https://user-images.githubusercontent.com/1512805/161583609-e3029382-b655-4f36-9157-b3137292aebc.png)

3. Delivering notifications via [Web Push](https://developer.mozilla.org/en-US/docs/Web/API/Push_API). Web Push allows web apps to push e2e-encrypted information to web browsers, even if the web site is not currently open. The standard API is currently supported by Chrome and Firefox.

@roschaefer: Since you asked about it, this is how it works (basically):

1. The frontend requests permission from a user to subscribe to push messages via a web API. The web API returns a push endpoint operated by browser vendors and auth details. The endpoint and auth details are stored in the database.
2. The frontend registers a service worker that runs in the background (i.e. even if the app is closed).
4. In order to trigger a notification, the backend sends the message in an encrypted payload to the push endpoint.
5. The service worker handles the push event and displays a (native) notification.

There is a gem that implements the encryption/encoding details: [`webpush`](https://github.com/zaru/webpush)

drjakob:
gem looks good

mattwr18:
@tillprochaska since I've been assigned this issue and in very much interested in working on it, I've read what you prepared and had some thoughts/questions.

First, I'd like to say thanks for taking the time to write a detailed description of how we might tackle the issue/issues with links to more information to gain a really nice context.

I agree with you that this very well could/should be broken into separate issues. I can imagine that a newsroom might not want to have push notifications, but might very well prefer to have those notifications delivered via a feed on the site, email, slack, etc... This makes it seem useful for many different use cases.

[...]

PR #1378, opened 2022-06-09
Add ActivityNotifications

### LINKED ISSUE #1376: Activity Notifications
```
As an editor
I would like to see notifications on recent activity, ie replies, successful onboarding, etc..
So that I can have an overview of what's happening
```

### Dev Notes
This issue is related to a request for push notifications by editors. See https://github.com/tactilenews/100eyes/issues/1287 and its related discussion.

There, [@tillprochaska suggested breaking it into two separate issues](https://github.com/tactilenews/100eyes/issues/1287#issuecomment-1087742329), one to introduce the concept of notifications and another to handle push notifications.

It was also suggested to have a look at the [noticed gem](https://github.com/excid3/noticed) for setting up and delivering notifications.

mattwr18:
Closes #1376 

## Description

This PR introduces `ActivityNotification` using the [noticed gem](https://github.com/excid3/noticed). At the moment, we only deliver these notifications to the database to store them and display them in an activity feed on the newly redesigned dashboard, which this PR introduces the first step of. We log 3 types of events at the moment:

- OnboardingCompleted
- MessageReceived
- ChatMessageSent

These notifications are delivered to all users. This is the first step in our notifications flow. We have not decided on whether we want to support web push notifications. See [the issue and the conversation there](https://github.com/tactilenews/100eyes/issues/1287), or if we want to notify them in some other way, for example using our own adapters to allow the User to add their preferred channel of communication, ie Signal, Threema, Telegram, or Email.

We have deployed this feature for user testing to our lz.100ey.es instance and gotten feedback from them, which has been added to this PR. We have also seen some bugs that were fixed.

ActivityNotification types `MessageReceived` and `ChatMessageSent` are grouped to help keep the notifications list items down and to make it easier to keep track of.

We have decided to add the `*_id`s of the params passed in to the message type in the `activity_notifications` table to benefit from eager loading and avoid O(N+ 1) issue discussed in the [ How to solve N+1 when params are records? #123 ](https://github.com/excid3/noticed/issues/123). This also resolved some weird issue that we saw on demo.100ey.es where the params were not being serialized correctly and were saved as a String instead of a Hash. See [#monitoring channel issue and discussion](https://tactilenews.slack.com/archives/C01ACFHFTV2/p1661513451897009) and [continued findings](https://tactilenews.slack.com/archives/C01ACFHFTV2/p1661758731052099?thread_ts=1661758685.894009&cid=C01ACFHFTV2)

This PR does not introduce updating the feed via ActionCable so the page must be refreshed for new activities to come through. Maybe, we want to add this as an improvement?

See other comments :point_down: for questions and possible future improvements.


app/models/contributor.rb
```diff
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
+# rubocop:disable Metrics/ClassLength
```
mattwr18, line 3: I'd like to discuss what to do here @phoeinx @tillprochaska @roschaefer ... I don't like just disabling this, but addressing the issue `rubocop` brings up seems outside of the scope of this `PR`.

Also, one thing I don't like about our pre-push hooks is that I can't even push up the code unless I address linting issues. If it is just a matter of running `rubocop -a` then fine, but when they require more thought and action, I don't want to leave code un-pushed and I don't want to be late to dinner or work longer than I have for a day so that I can push my code.

tillprochaska: > Also, one thing I don't like about our pre-push hooks is that I can't even push up the code unless I address linting issues. If it is just a matter of running rubocop -a then fine, but when they require more thought and action, I don't want to leave code un-pushed and I don't want to be late to dinner or work longer than I have for a day so that I can push my code.

I personally have don’t have the git hooks set up because I find them annoying for exactly the same reason and suggested removing them some time ago. But I think Robert (?) that he actually finds them useful?

So if you’re annoyed by them, maybe simply disable/remove them for now?

app/models/message.rb
```diff
@@ -56,7 +59,17 @@ def chat_message_link
...
-      anchor: "chat-row-#{id}"
+      anchor: "message-#{id}"
     )
   end
+
+  private
+
+  def notify_recipient
+    if reply?
+      MessageReceived.with(contributor: sender, request: request, message: self).deliver_later(User.all)
```
tillprochaska, line 70: `deliver_later` enqueues a job. Jobs are written to the database. Once the job is executed, it will then delivery the notification, which in our case means writing a notification record to the database.

So maybe we could just use `deliver` instead of `deliver_later` here? Writing a job to the database isn’t going to be faster than writing the actual notification to the database.

mattwr18: Do you mean just for the `MessageRecieved` type?

I guess the benefit of enqueuing a job is not that it's faster, but that it happens in the background and the user does not need to wait for it to happen. In our case, the `MessageReceived` type wouldn't need to concern itself with this since there is no user interaction with the site, but for `OnboardingCompleted` and `ChatMessageSent` there would be. Or are you just not concerned with the time it takes to create a record for each User, which is typically quite few?

mattwr18: Note that making this change resulted in many errors popping up that were resolved when removing, rebuilding, and deploying an image without this proposed change. See [monitoring channel discussion](https://tactilenews.slack.com/archives/C01ACFHFTV2/p1662381100797989).

It doesn't make sense to me how this would result in a `ActiveModel::UnknownAttributeError` for `contributor_id`, but I tested the error/fix flow on multiple instances, staging and demo and reliably experienced the error/saw the fix get rid of the error. When I manually created the notifications in the rails console I did not get an error.

app/models/contributor.rb
mattwr18: That would be a solution. I don't know how to disable them. I quickly searched for it after I read your comment and didn't find anything.

I would personally be in favor of removing them for the reasons stated. The downside would be that we would use more minutes if we push up code that fails linting in our actions.

phoeinx: I personally like the hooks exactly because I otherwise tend to forget linting the code and it then fails in our actions, which means I have to refocus on the code again even though I probably started something new already.

tillprochaska: > That would be a solution. I don't know how to disable them. I quickly searched for it after I read your comment and didn't find anything.

You can run `rm -rf ./.git/hooks` to remove all installed Git hooks. I think you might need to do that again when you (re-)install lefthook from npm.

app/activity_notifications/chat_message_sent.rb
```diff
@@ -0,0 +1,63 @@
...
+  def pluralization_key(count)
+    case count
+    when 1
+      :one
+    when 2
+      :two
+    else
+      :other
+    end
+  end
```
mattwr18, line 58: There is duplication with this method and the one in `MessageReceived` class. I considered if it was worth it to add these to a helper, but as it only appears twice, didn't think it was worth it. If the events grow and we want to maintain different German for one, two, and other than I think it would be worth it. What do others think?

app/assets/stylesheets/application.css
```diff
@@ -1,2 +1,3 @@
 @import './global/*.css';
 @import '../../components/**/*.css';
+@import '../../views/**/*.css';
```
mattwr18, line 3: @phoeinx @tillprochaska and I started a discussion on this, but didn't finish it.  My question was

> Component-only views - I see we have a PageHeader component, which I originally used, but it has a gray background and padding that I don't want. I added my own HTML, but that meant adding a dashboard/index.css file and importing it. From looking at the code, I think you don't do this anywhere, but favor a component-only solution. So, should I refactor the PageHeader to accept a gray/white background, with/without padding, etc?

He responded, and I responded, but then the conversation died. See the [Display notifications PR](https://github.com/tactilenews/100eyes/pull/1412), which was merged into this one.

I feel like we can discuss and add this as an improvement after this has been merged and deployed.

app/components/application_component.rb
```diff
@@ -6,19 +6,20 @@ class ApplicationComponent < ViewComponent::Base
   include SvgHelper
   include ColorHelper
 
-  def initialize(styles: [], style: nil, **attrs)
```
mattwr18, line 9: I was surprised that the `Card` component didn't have any padding. I noticed that we only use it in 3 places at the moment and it always has a nested `box` component whose only job is to add padding and/or background colors.

I do not like the idea of adding html elements for the purpose of styling as this can be done with CSS and scoping. I would like the ability to use the Card component, or any other as the parent element in my component and scope it to that particular component. Giving frontenders the ability to pass in a unique class name to scope the CSS is a nice addition to our component library.

I originally updated the Card component. See [`a9b23ec` (#1378)](https://github.com/tactilenews/100eyes/pull/1378/commits/a9b23ecb1f63eb0a1001b8fb5a45b8c84c8e268e) where I removed the code as it changed several components which have nothing to do with this feature.

I don't think there is any use case where we would want to use a card without padding and being able to configure its background color sound like something that would be super useful. I plan to put in another PR to reintroduce these changes, but I still feel like having the ability to pass down a class from one component to another is super useful. One of the things that drives me crazy about popular component libraries is their inflexibility in giving the ability to do so when you want to apply CSS to change some styling.

app/models/contributor.rb
mattwr18: No one commented on the fact that we just disable the cop. I guess we might want to see if there is a way we can make this model skinnier?

This would be another ticket then, I believe. Or, are people not concerned with this cop in this case?

app/models/message.rb
```diff
@@ -59,4 +62,15 @@ def chat_message_link
...
   end
+
+  private
+
+  def notify_recipient
+    if reply?
+      MessageReceived.with(contributor_id: sender.id, request_id: request.id, message_id: id).deliver_later(User.all)
+    elsif !broadcasted?
+      ChatMessageSent.with(contributor_id: recipient.id, request_id: request.id, user_id: Current.user.id,
+                           message_id: id).deliver_later(User.all)
```
mattwr18, line 73: Also commented on a [PR adding some improvements](https://github.com/tactilenews/100eyes/pull/1431#discussion_r958347111), which has been merged into this one:

> @tillprochaska maybe we can discuss this? I found it quite surprising that we don't save any information on which person in the newsroom has sent a "chat messge" to a contributor.

> I understand that we save everything as a Message and then we do a slightly complicated way of deciding if it's a broadcasted?,reply? or a manually_created? message, but we don't save any information on which User sent the message.

> I guess we could not save it in the ActivityNotification as well, then we could completely remove the ActiveSupport::CurrentAttributes. At present, we use this information to display which User sent out the chat message.

Haven't received a response on this issue.

app/models/user.rb
```diff
@@ -3,10 +3,19 @@
 class User < ApplicationRecord
   include Clearance::User
 
+  has_many :activity_notifications,
+           as: :recipient,
+           dependent: :destroy
+  has_many :notifications, class_name: 'ActivityNotification', dependent: :destroy
```
mattwr18, line 9: maybe we should consider renaming these to make it easier to understand?

like `user.notifications_as_recipient` and `user/message/request/contributor.notifications_as_mentioned` or something similar?

app/models/user.rb
```diff
@@ -3,10 +3,19 @@
...
   has_one_time_password
   validates :password, length: { in: 8..128 }, unless: :skip_password_validation?
 
   def name
     "#{first_name} #{last_name}"
   end
+
+  def avatar?
+    false
+  end
```
mattwr18, line 20: related to the change to the Avatar component mentioned :point_up:

app/views/dashboard/index.html.erb
```diff
@@ -1,38 +1,14 @@
...
-        <%= c 'heading', tag: :h2 do %>
-          <%= I18n.t 'dashboard.how_it_works' %>
-        <% end %>
-        
-        <%= c 'icon_list',
-          elements: I18n.t('dashboard.how_it_works_list'),
-          styles: [:vertical, :alternating]
-        %>
-
+<div class='Dashboard'>
```
mattwr18, line 1: I received feedback from @tillprochaska that we should change the background-color of the dashboard page so I added this wrapping `div` and gave it styles, but this violates the pattern of component-only pages and introduces import views css and I wrote `div` that this bigger issue for me is the fact that we are not using semantic HTML.

> I added a commit for this, but I'm not happy about the way I achieved it. I added a div to wrap the entire page and added a class/background-color to it.
It's actually one of my pet peeves, that I would like to discuss with you. I really do not like the patter of using the main tag in the layout. I understand that it means that we cannot forget to have a main tag on a page, but it means that if I have the need of a tag to wrap the entire page, I am left with using a non-semantic HTML tag, div , when in reality the semantic choice is main as it contains the main content of the page.

I could open a PR that introduces a component, which is the main content of the page and would move the styles out of the views into this component and also update the existing page views, which aren't many. However, I think we can do this in an improvement PR.

phoeinx: yeah I agree here too that introducing a new dashboard component as the main content of the page seems to be the way to go. Would be great if you could open a PR! 👍

phoeinx (approved):

app/activity_notifications/onboarding_completed.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def record_for_avatar
+    record.contributor
+  end
+
+  # rubocop:disable Rails/OutputSafety
+  def group_message(notifications:)
+    t('.text_html',
+      contributor_name: notifications.first.contributor.name,
+      contributor_channel: record.contributor.channels.first.to_s.capitalize).html_safe
```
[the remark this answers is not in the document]
mattwr18: I like that suggestion.

app/assets/stylesheets/application.css
mattwr18: See https://github.com/tactilenews/100eyes/issues/1454

app/components/activity_notifications/activity_notifications.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <ul role='list' class='ActivityNotifications-list'>
+        <% activity_notifications.each do |notification_hash| %>
+          <li class='ActivityNotifications-listItem' >
+            <%= c 'avatar', record: notification_hash[:record] %>
+            <header class='ActivityNotifications-listItemHeader'>
+              <p>
+                <%= notification_hash[:group_message] %>
+              </p>
+              <time class='ActivityNotifications-listItemRelativeTime'>
+                <%= "vor #{time_ago_in_words(notification_hash[:created_at])}" %>
```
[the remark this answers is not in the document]
mattwr18: I did that at one point, but reverted it because I don't know enough German to know if we would ever use `etwa eine Stunde` for any reason without the `vor`. Also, it should be noted that these translations come from Rails, they were not added by us, so that made me think there might be instances where we would. The final consideration that helped me make the decision in favor of leaving it here in the template was that it would be repeated ~20x in the `de.yml` vs only once here.

For now, I'll just leave it for those reasons :point_up:, but I don't have a strong feeling about it so let me know if these are things you already considered and would still prefer it in the translation yml. One strong argument would be that if we use another language, this would not work, for example.

app/models/message.rb
tillprochaska: This is for historical reasons, as the prototype didn't have multi-user auth. Storing the user makes sense!

app/components/application_component.rb
mattwr18: > I see your point with the nested box component. On the other hand, using a box component for padding and/or background colours also is very versatile and removes the need to add padding and background parameters to every component. I assume that was/is the idea behind the box component.

I come from the school that says that we HTML elements are for structure and CSS to apply styling. We add HTML elements in this case only for the purpose of adding some styling, ie padding and background color when the same can be achieved with CSS. In the case that we want to dry something out, we can use @mixins to create a style and apply it widely. One way this affects us is that we have a lot of nested elements, some of which could be avoided by following this principle, and it makes it more difficult to navigate the HTML tree. Considering this example:

Screenshot from 2022-09-21 10-50-44

There are 11 nested elements! Maybe others don't mind this. I know that @tillprochaska has reasons for setting it up like this. I don't think one is wrong and the other right, I just think they are different ways of looking at frontend code and how it's organized. I know that he sent me a link explaining how he looks at it, but I can't seem to find it. Anyways, I think there is space for doing things in slightly different ways.

Thankfully, @tillprochaska already provided a way to inject classes and other HTML attributes. A PR I helped review and had forgotten about. See  [Add helpers to add HTML attributes to component #452 
](https://github.com/tactilenews/100eyes/pull/452). This means that we don't need to make any changes, so no follow-up PR for this.

app/components/card/card.html.erb
```diff
@@ -1,3 +1,3 @@
-<div class="<%= class_attr %>">
+<article class="<%= class_attr %>">
```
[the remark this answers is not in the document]
mattwr18: I would say that is the definition of a Card component, but maybe that is just how I look at it. One potential issue might be that the request row uses an article tag to wrap the card, which is also an article tag in this case. I will address this in another PR.

app/helpers/date_time_helper.rb
```diff
@@ -8,4 +8,23 @@ def date_time(date, format: nil)
 
     I18n.l(date, format: :short)
   end
+
+  def time_based_heading
```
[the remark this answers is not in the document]
mattwr18: They are one second off, but since this is the last when statement, it wouldn't matter cause it would never reach this part of the statement if it is midnight or later. Since it removes a line of code, I like it! thanks :wink:

app/models/contributor.rb
mattwr18: I would like to, to be honest cause I am interested in this type of refactoring, I just don't know if it provides much business value and so don't know if it can be prioritized. Just didn't want to not mention the fact that I am just disabling the cop so that we don't have to address the underlying issue.

app/models/message.rb
mattwr18: See https://github.com/tactilenews/100eyes/issues/1456

app/models/user.rb
mattwr18: I don't know if I like the semantics of `notifications_as_sender`... In the case the `user` has their id saved to the `activity_notifications`  table, there would also be a `contributor_id`, `message_id`, and `request_id` since this only happens for `ChatMessageSent` types. In this case, they would all be the "sender"?

app/views/dashboard/index.html.erb
mattwr18: I've created an issue, which will have a follow-up PR See [Create MainComponent](https://github.com/tactilenews/100eyes/issues/1457)

spec/spec_helper.rb
```diff
@@ -101,4 +101,8 @@
   config.after(:each) do
     Setting.clear_cache
   end
+
+  config.before(:each, type: :system) do
```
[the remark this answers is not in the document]
mattwr18: I was the one who placed it in the `rails_helper` in https://github.com/tactilenews/100eyes/commit/bd9f94e9a983228acafd2d85aa7f0be95caf93bd, but it seems to me that most of our before/after config is in the spec_helper, so just trying to be consistent. The only exception now is `config.after { Telegram.bots.each_value(&:reset) }`.

I don't know if there is a best practice around this, or not.

spec/factories/messages.rb
```diff
@@ -4,6 +4,7 @@
   factory :message do
     created_at { Time.zone.now }
     unknown_content { false }
+    text { 'Am I an incoming message?' }
```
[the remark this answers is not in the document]
mattwr18: Actually, I changed it to just be a `Faker::Lorem.sentence` since if we override the trait `with_sender` in the test then it would not be an incoming message, but an outgoing one.
I still like never having a `nil` text. A little surprised we don't validate that this cannot happen.

PR #1380, opened 2022-06-13
Add signal-cli-config to backup created by ansible backup script.

phoeinx:


mattwr18 (commented):
I guess there is not an issue related to this task? It makes sense to me why we might want to also back up this config and restore it if needed, but maybe since there is no issue, you could write a brief PR description?

I tested the backup/restore_locally from this branch and it doesn't appear to have broken anything. One thing I noticed is that we don't add a step to the `local_restore` to unarchive the compressed signal-cli-config folder like we do with the storage backup/restore.

We probably want to do that, no? Also, I noticed that we are missing instructions on how to verify a backup/restore was successful. I noticed you added some notes in 1password, I also did the same, but I think we used different ways of verifying its integrity. Also, I didn't do anything to verify the storage was backed up and restored successfully. Maybe, we could work together to write some docs on what we think is sufficient to verfiy its success?

Once I restore the signal-cli-config, what can I do to verfiy it worked?

FrauCsu:
Hi @phoeinx are you on to this? Would be great to get this done asap.

mattwr18:
> I’d be careful restoring signal-cli configuration in a development environment. If you forget that you restored a configuration or if you run certain signal-cli commands, this can have an impact on the Signal account and consequently also on the production instance the account belongs to.

That's a good point I hadn't thought of.

> Basically, my opinion on this is that we should probably not verify the integrity of the signal-cli backups by restoring locally. We should probably verify in regular intervals that we can restore signal-cli backups, but I’d do that using a staging backup.

Maybe, we should be testing out our backups on the staging server instead of locally since there are some key differences.

phoeinx:
Hi,

As a short explanation - we had been talking for quite some time about also backing up the signal-cli-config in case of that something went wrong. In June it then actually did (caused by me) and we had to restore a Signal setup from a signal-cli-config and could verify that it works on a production instance. Therefore that was my follow-up PR to include storing the signal-cli-config in all our back-ups.

Otherwise, I would agree with Till about not restoring signal-cli-configs into local development setups. I think in regular periods testing if we can restore the `staging` signal-cli-config on `staging` should be a good method to verify the ability to use those backups 👍 .

mattwr18 (approved):
Great! thanks for walking me through these changes and pointing out the potential issues to trying to test restoring the backup locally.

PR #1385, opened 2022-06-17
Fix bug where outbound adapters triggered from incoming messages

mattwr18:


spec/models/message_spec.rb
```diff
@@ -113,5 +113,17 @@
         end
       end
     end
+
+    describe 'incoming message' do
+      let(:message) { create(:message, sender: create(:contributor), recipient: nil) }
+
+      it 'does not send outbound messages to any channels' do
+        [PostmarkAdapter::Outbound, SignalAdapter::Outbound, TelegramAdapter::Outbound, ThreemaAdapter::Outbound].each do |adapter|
+          expect(adapter).not_to receive(:send!)
```
roschaefer, line 122: You could also test the happy path here, so we can see that `recipient: nil` makes the difference.

roschaefer (approved):
nailed it

app/models/message.rb
```diff
@@ -59,4 +55,12 @@ def chat_message_link
...
     )
   end
+
+  private
+
+  def send_outbound_messages
+    [PostmarkAdapter::Outbound, SignalAdapter::Outbound, TelegramAdapter::Outbound, ThreemaAdapter::Outbound].each do |adapter|
+      adapter.send!(self)
+    end
+  end
```
tillprochaska, line 67: I have two suggestions:

1) Could we move the condition from `after_create_commit` into the method (as a guard clause). This ensures that no matter how this method is called (i.e., not by a lifecycle hook), it will only send outbound messages. Also, I’d prefer to have all the logic in one place.

2) Can we rename this method to `send_if_outbound` or something similar? I think I see where the current name comes from (with “messages” referring to the actual email or Signal/Threema/Telegram message instead of the `Message` record in our database). However, we already use the term “message” for our `Message` model, so I’d prefer if we didn’t use it for two things with slightly different meanings.

tillprochaska (changes requested):

app/models/message.rb
tillprochaska: @mattwr18 Just saw that you changed that literally minutes ago!

app/models/message.rb
```diff
@@ -59,4 +55,14 @@ def chat_message_link
       anchor: "chat-row-#{id}"
     )
   end
+
+  private
+
+  def send_outbound_messages
+    return if manually_created? || reply?
```
mattwr18, line 62: I moved the conditional to the private method because there was a failing test `but if contributor blocked the telegram bot` where the method wasn't being called to send outbound messages when the message was created.

Also, I read that it is a best practice to have the conditionals in the method call to allow them to become more complex and not be defined in the line that creates the `after_commit`. Note that the same place I read this actually suggests to have the guard clause in the worker itself because even though it will enqueue more jobs, they are no ops. See [Rails Callbacks: 5 Best Practices Used at Gusto](https://engineering.gusto.com/the-rails-callbacks-best-practices-used-at-gusto/)

I still however believe that we don't even want to try to send an outgoing message after receiving an incoming one for reasons stated in the issue.

app/models/message.rb
mattwr18: Yes, I also just commented on why I changed it. It was pretty similar to the reason you gave, except for with the caveat that it was buggy with the inline conditional/proc and a test was failing. Thanks for tests!

2. I don't know if I understand your concern. It is referring to the `Message` model as this is what we send with our adapters. The Outbound adapters accept `self` as a param, which is the instance of `Message`, they then call `perform_later` with `message` and/or `message.text`

tillprochaska: > I guess send_if_outbound doesn't seem to address the manually_created? part of the condition.

A message that’s created manually is always an inbound message.

tillprochaska: > so the method is actually responsible for sending out messages.

Yep, and that’s the point. I’d prefer if we didn’t use the term *Message* for both the `Message` record type and the Signal/Threema/Telegram/email message. Even though they are obviously related, they are different and it introduces ambiguity.

I’d like to think of this as "sending out a single message via multiple channels".

(Also in practice, we never send out a message via multiple channels because contributors onboard using a single channel so the fact that we loop over multiple adapters is more of an implementation detail.)

mattwr18: > I’d prefer if we didn’t use the term Message for both the Message record type and the Signal/Threema/Telegram/email message.

I am not using it to refer to the Signal/Threema/Telegram/email message, I am using it to refer to the instance of the `Message` class, which is being sent by the channel.

> and it introduces ambiguity.

It doesn't for me cause it's clear to me that message refers to the instance of the `Message` class that is being sent. Also, `Telegram` for example already uses `send_message` as the name of their method so the ambiguity, if there is any, already exists to some degree.

All this being said, I don't have an issue with your proposed naming, I just want to make sure that we are not asking for changes to code for personal code preferences, which we all have and are different from each other sometimes.

PR #1391, opened 2022-06-24
Add ErrorNotifier class and add context to Signal errors

mattwr18:


app/adapters/signal_adapter/outbound.rb
```diff
@@ -35,6 +37,13 @@ def perform(text:, recipient:)
         http.request(req)
       end
       res.value # may raise exception
+    rescue Net::HTTPServerException => e
+      ErrorNotifier.report(SignalError.new, context: {
+                             code: e.response.code,
+                             message: e.response.message,
+                             headers: e.response.each_header(&:header),
```
tillprochaska: Does this work? There was [an error](https://sentry.io/organizations/tactilenews/issues/3381440601/?project=5686596&query=is%3Aunresolved) in Sentry… I think `e.response.to_h` might be an easier way to get a hash.

mattwr18: when I attempted this locally it threw an error... I have since updated the code to fix the error

app/adapters/signal_adapter/outbound.rb
```diff
@@ -38,7 +38,7 @@ def perform(text:, recipient:)
       end
       res.value # may raise exception
     rescue Net::HTTPServerException => e
-      ErrorNotifier.report(SignalError.new, context: {
+      ErrorNotifier.report(ServerException.new, context: {
                              code: e.response.code,
```
tillprochaska, line 42: Why do you create a new exception object here instead of simply reporting the rescued exception? We lose the stack trace of the original exception… :(

```suggestion
      ErrorNotifier.report(e, context: {
```

app/adapters/signal_adapter/outbound.rb
tillprochaska: Ah, it seems to be `to_hash` instead of `to_h` -- I always get confused about when to use `to_h` vs. `to_hash`

app/adapters/signal_adapter/outbound.rb
mattwr18: Ok, I tested this out and I think it's not actually true that it is easier to see where the error occurred because with the original exception you also have the class name on the high level view. See screenshot

Screenshot from 2022-06-28 09-38-53

That being said, I don't think we actually lose anything as you suggested. When I reported the original exception, it was just a string `400 "Bad Request"`, which we have in the code/message.

Compare the two screenshots:


Screenshot from 2022-06-28 09-38-17
Screenshot from 2022-06-28 09-21-36

tillprochaska: Ah, sorry, I was actually talking about the stack trace which is empty, but I can see it’s there for the recent reported events! :)

image

spec/adapters/signal_adapter/outbound_spec.rb
```diff
@@ -77,8 +78,13 @@
       end
 
       describe 'on error' do
-        before { stub_request(:post, 'http://signal:8080/v2/send').to_return(status: 400) }
-        it { should raise_error(Net::HTTPServerException) }
+        before(:each) { stub_request(:post, 'http://signal:8080/v2/send').to_return(status: 400) }
+
+        it 'ErrorNotifier' do
+          expect(Sentry).to receive(:capture_exception).with(Net::HTTPServerException)
```
tillprochaska, line 84: It would be nice to replace the stubs of `Sentry`, and stub `ErrorNotifier` instead, as we now have that abstraction in place (I like that!). (That way, should we ever move away from Sentry, we can simply adjust the `ErrorNotifier` implementation without having to adjust all the tests. Doesn’t need to be done before merging though.

tillprochaska (approved):
Nice, I like it!

spec/adapters/signal_adapter/outbound_spec.rb
```diff
@@ -77,8 +78,13 @@
       end
 
       describe 'on error' do
-        before { stub_request(:post, 'http://signal:8080/v2/send').to_return(status: 400) }
-        it { should raise_error(Net::HTTPServerException) }
+        before(:each) { stub_request(:post, 'http://signal:8080/v2/send').to_return(status: 400) }
+
+        it 'ErrorNotifier' do
```
tillprochaska, line 83: Oh, maybe change this to something that sounds a little more natural?

```suggestion
        it 'reports the error' do
```

PR #1402, opened 2022-06-28
Bump signal-cli-rest-api and add tasks to set global trust_mode config

mattwr18:
Built and deployed this on staging. Ran the play locally and got these results:

### POST:
Screenshot from 2022-06-28 16-53-40

### GET:
Screenshot from 2022-06-28 16-54-06

Sending still failed to the `UNTRUSTED` contributor, manually sending to that contributor with the `--trust-new-identities=always` flag also fails with:

```
Failed to send (some) messages:
+4915735994012: Unregistered user "+4915735994012"
```

### DIFF
--- a/ansible/roles/signal/tasks/main.yml
+++ b/ansible/roles/signal/tasks/main.yml
@@ -13,47 +13,44 @@
   set_fact:
     signal_phone_number_not_registered: "{{ response['status'] != 200 and response['json']['error'] == error_message }}"
 
-- name: Ask for captcha result
-  pause:
-    prompt: "{{ lookup('template', 'enter_captcha.j2') | trim }}"
-  register: captcha_prompt_result
-  when: signal_phone_number_not_registered
+- name: Register, verify phone number
+  block:
+    - name: Ask for captcha result
+      pause:
+        prompt: "{{ lookup('template', 'enter_captcha.j2') | trim }}"
+      register: captcha_prompt_result
 
-- name: Set captcha result
-  set_fact:
-    captcha: "{{ captcha_prompt_result.user_input }}"
-  when: signal_phone_number_not_registered
+    - name: Set captcha result
+      set_fact:
+        captcha: "{{ captcha_prompt_result.user_input }}"
 
-- name: Register phone number
-  uri:
-    url: "{{ signal_cli_rest_api_endpoint }}/v1/register/{{ signal_server_phone_number }}"
-    method: POST
-    status_code: 201
-    body_format: json
-    body:
-      captcha: "{{ captcha }}"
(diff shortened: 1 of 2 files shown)


mattwr18:
I was able to test this on staging, by:

1. Registering the Sportschau number locally.
2. Going through the onboarding process.
3. Verifying I was listed as `TRUSTED_UNVERIFIED` by:
	- sshing into the staging server
	- bashing into the signal service
	- running  `signal-cli --config /home/.local/share/signal-cli -a +4915735993163 listIdentities -n +4915735994012`
4. Verifying I could receive messages.
5. Unregistering the account locally.
6. Placing the data in `~/.local/share/signal-cli/data` into another directory
7. Re-registering number locally
8. Sending out another message, this seems to have marked the number as `UNTRUSTED` in signal-cli
9. Verify number was `UNTRUSTED` with :point_up_2: command.
10. Sending out another message, which was received.

I also added a check to the `register_signal_phone_number.yml` playbook to see if we are able to receive signal messages on the server using our `signal_server_phone_number`, which returns a 400 with `User {{ signal_server_phone_number }} is not registered.\n` if the number is either not registered and/or not verified. This can then be used to skip the registration and verification steps of this playbook and make it idempotent.

ansible/roles/signal/tasks/main.yml
```diff
@@ -1,10 +1,28 @@
...
+  pause:
     prompt: "{{ lookup('template', 'enter_captcha.j2') | trim }}"
   register: captcha_prompt_result
+  when: signal_phone_number_not_registered
 
-- set_fact:
+- name: Set captcha result
+  set_fact:
     captcha: "{{ captcha_prompt_result.user_input }}"
+  when: signal_phone_number_not_registered
```
tillprochaska, line 25: Something I noticed while having a quick look at this: You can group multiple tasks in [blocks](https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html) if the tasks share certain properties. For example, in this case, you could group all tasks related to registering/verifying and use a single `when` condition on that block (instead of setting the `when` condition on every step individually.)

I’ll take a closer look in the following days.

mattwr18: That's cool! I've tested it out in e7676114 and it seems to work.

FrauCsu: @tillprochaska Have you had the chance to have a look?

tillprochaska (approved):
I tested both cases:

* Running the playbook when the phone number is already set up
* Running the playbook when the phone number hasn’t yet been set up.

Both worked as expected! 👍

Issue #1409, opened 2022-07-12
Investigate why we don’t get Dependabot updates for our Docker base image


tillprochaska:
We currently use `ruby:3.0.2-alpine3.12`. There are newer versions of both Alpine and Ruby available. Dependabot is configured to update our Docker dependencies and has run recently, but no PRs to update the image have been opened.

mattwr18:
@roschaefer do you think you might get to this this quarter or should we move it to the future?

tillprochaska:
By the way, it seems Dependabot has fixed the underlying issue (https://github.com/dependabot/dependabot-core/issues/3929), so maybe maybe it’ll work again soon?

PR #1410, opened 2022-07-12
Downgrade postcss-cli

tillprochaska:
I merged a Dependabot update, because the only breaking change was dropped support for Node 12. Turns out we’re still on version 12, even though the current LTS version is already at 16. As a short-term fix, I’ve reverted the upgrade.

I have also created #1409, as we should have updated our Docker base image to be on a newer version of Alpine (and Node).


mattwr18 (approved):
Cool, I like this flow to fix the issue. First revert the change that introduced it, upgrade our node version, then bump the package again.

PR #1412, opened 2022-07-13
Display activity notifications

mattwr18:
- [ ] Add bell icon to link to a user's notifications on the dashboard
- [ ] Explore idea of marking read
- [ ] Add inside turbo_tag to facilitate display in dropdown in menu
- [ ] Group MessageReceived notifications


mattwr18:
@tillprochaska Some questions I have about the frontend in general and this PR include:

- Font-size, the mockups seem to use much bigger text than we currently have set for h1, h2. Should we always respect these font-sizes to keep consistency in the project, or should I try to make it most similar to the design?

- Spacing - same question. --spacing-unit-xl does not appear to be enough spacing at the top of the page. Should I be adding calc(1.5 x ...)or should we only be using pre-set variables/adding new ones.

- Component-only views - I see we have a PageHeader component, which I originally used, but it has a gray background and padding that I don't want. I added my own HTML, but that meant adding a `dashboard/index.css` file and importing it. From looking at the code, I think you don't do this anywhere, but favor a component-only solution. So, should I refactor the `PageHeader` to accept a gray/white background, with/without padding, etc?

app/activity_notifications/message_received.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  # Define helper methods to make rendering easier.
+  #
+  # rubocop:disable Rails/OutputSafety
+  def message
+    t('.message_html',
+      contributor_name: contributor.name,
+      request_title: params[:request].title).html_safe
+  end
+  # rubocop:enable Rails/OutputSafety
```
mattwr18, line 28: Don't really like this. I have seen from the rails docs:

> [Keys with a '_html' suffix and keys named 'html' are marked as HTML safe. When you use them in views the HTML will not be escaped.](https://api.rubyonrails.org/v6.1.5/classes/ActionView/Helpers/TranslationHelper.html#method-i-translate)

But, when I've attempted to do this, it just displays the `<b>` as plain text.

tillprochaska:
> Font-size, the mockups seem to use much bigger text than we currently have set for h1, h2. Should we always respect these font-sizes to keep consistency in the project, or should I try to make it most similar to the design?

The screendesign/mockups aren’t meant to be implemented pixel perfect and they might not be 100% consistent/systematic in themselves. Use different sizing and spacing if necessary/where you see fit. In general, I’d always prefer consistency and a systematic.

> Spacing - same question. --spacing-unit-xl does not appear to be enough spacing at the top of the page. Should I be adding calc(1.5 x ...)or should we only be using pre-set variables/adding new ones.

I tend to use multiples of the base spacing unit if the predefined are not enough. But you can also add `--spacing-unit-xxl`, `--spacing-unit-xxs` etc. if you’d prefer!

> I added my own HTML, but that meant adding a dashboard/index.css file and importing it. From looking at the code, I think you don't do this anywhere, but favor a component-only solution.

Yes, I’d prefer scoping all styles to specific components, as in my experience, it helps reducing unintended side effects.

> So, should I refactor the PageHeader to accept a gray/white background, with/without padding, etc?

The questions refers to this header, right?

image

I’d probably not try to adjust the `PageHeader` component to fit this case. Maybe you could just use the `Flex` utility component? Something like this might work, although I haven’t tested it:

```erb
<header>
  <%= c('flex', style: :spaceBetween) do %>
    <%= c('heading', tag: :h1) { 'Lorem Ipsum' } %>
    <%= c('button', …) %>
  <% end %>
</header>
```

tillprochaska:
@mattwr18 Enhancements that have been prioritized:

* Change the link "Zur Frage" to "Zur Antwort" (or "Zu den Antworten" once the messages are grouped). We already support anchor links for individual messages, i.e. we’d only need to add the respective hash to the link, e.g. `/requests/123#message-456`.

* There are two main problems: Editors are overwhelmed by the amount of notifications and they’d like to see at a glance whether another editor has already replied. As this is kind of a difficult problem. We’d like to implement grouping of notifications first, get more feedback, and see if how to solve the second problem from there.

Regarding grouping, I was thinking if adding a separate notification type for replies *by editors* and grouping only successive notifications of the same type might solve this issue. But I haven’t done any testing around it.

**Letzte Aktivität**

* **Contributor C** and **Contributor D** have replied to **Question A**.

* **Editor B** has replied to **Contributor A** and three others.

* **Contributor A** and three others have replied to **Question A**.

Note how the first and the second notification are both about Question A and have the same type (new reply by contributor), but they aren’t grouped as a single item because something else happened in between. This might make it easier for people to understand the chronological order, and might give them enough insight to answer the question "Did someone else from my team already take care of these new messages."

PR #1416, opened 2022-07-22
Convert feature specs to system specs

mattwr18:


spec/system/admin/raw_data_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  it 'admin debugs raw message data' do
+    visit admin_message_path(id: message.id, as: user)
+    click_on 'text.json'
+
+    switch_to_window(windows.last)
+
+    expect(page.body).to have_text('{"text":"Hello"}')
+    # expect(page.response_headers['Content-Type']).to eq('application/json')
+    # expect(page.response_headers['Content-Disposition']).to start_with('inline;')
```
mattwr18, line 17: These expectations are failing now... Not exactly sure why, but some differences I noted is that with a real browser, we must call this `switch_to_window(windows.last)` and then there are no `response_headers`, so it throws an error `NoMethod [] for NilClass`... If you refresh the page, there are `response_headers` and I could see the `Content-Disposition` did start_with `inline`, but could not see the `Content-Type`

spec/system/auth/sign_in_spec.rb
```diff
@@ -2,29 +2,29 @@
...
+    it 'editor signs in' do
       visit sign_in_path
 
       # Enter incorrect credentials
       fill_in 'session[email]', with: 'zora@example.org'
       fill_in 'session[password]', with: 'abcdefgh'
       click_button 'Anmelden'
 
-      expect(page).to have_current_path(session_path)
+      expect(page).to have_current_path(sign_in_path)
```
mattwr18, line 27: This was weird that this was passing... `session_path` is a:

```
session POST   /session(.:format)                                                                                sessions#create
```

I think we do _not_ expect the user to be on this path

spec/system/auth/sign_in_spec.rb
```diff
@@ -37,18 +37,18 @@
...
```
mattwr18, line 51: same as :point_up_2:

spec/system/settings/image_uploads_spec.rb
```diff
@@ -2,22 +2,24 @@
...
   let(:user) { create(:user) }
   before(:each) { allow(Setting).to receive(:project_name).and_return('Die Lokal-Community!') }
   let(:jwt) { JsonWebToken.encode({ invite_code: 'ONBOARDING_TOKEN', action: 'onboarding' }) }
 
-  scenario 'Upload new onboarding logo' do
+  it 'Upload new onboarding logo' do
     visit onboarding_path(jwt: jwt)
-    expect(page).to have_css('header', text: 'Die Lokal-Community!')
+
+    expect(page).to have_css('header', text: 'DIE LOKAL-COMMUNITY!')
```
mattwr18, line 13: Change was made because we transform the header to uppercase it and in a real browser the test fails

spec/components/onboarding_signal_link_spec.rb
```diff
@@ -8,7 +8,5 @@
   before(:each) { allow(Setting).to receive(:signal_server_phone_number).and_return('+4915112345678') }
 
   it { should have_css('.OnboardingSignalLink') }
-
-  # \u00a0 is a non-breaking space
-  it { should have_css('strong', text: "0151\u00a01234\u00a05678") }
+  it { should have_css('strong', text: '015112345678'.phony_formatted(normalize: :DE, spaces: ' ')) }
```
mattwr18, line 11: This test was kinda of inexplicably failing as I didn't think adding a test runner for our system specs should affect our component specs. Having a closer look, the configuration I borrowed from [EvilMartians blog on System Tests](https://evilmartians.com/chronicles/system-of-a-test-setting-up-end-to-end-rails-testing), configures `:better_cuprite` as the `Capybara.default_driver`.  This causes slight differences as it's actually run in a headless browser.

I considered changing this and using `rack_test` for all tests which are not system tests, but the test suite ran in CI in 2min 45s, so I don't think speed is an argument for favoring rack_test.

tillprochaska: I actually really like the ability to run component tests in a real browser as well!

spec/system/admin/raw_data_spec.rb
mattwr18: I've removed these expectations as they don't fit with the reality of using a real browser. See fdb29750
We could remove the `target="_blank"` from the link in [app/views/fields/raw_data_field/_show.html.rb]() and we would get the `page.response_headers` to be able to make assertions, but I'm not sure it's that important to the test.

There are other potential benefits to not opening a new window. It is an `a11y` issue, for example. Also, some believe it to be bad UX, but this is debatable. For now, I have decided to leave it as is because some admin might already be familiar with how it works now and not invite change.

spec/system/activity_notifications_spec.rb
```diff
@@ -165,7 +168,8 @@
 
       click_link('Zur Antwort', href: request_path(request, anchor: "message-#{reply_by_same_contributor.id}"))
       expect(page).to have_text("I'm a reply from #{contributor_two.name}")
-      click_link('nachfragen', href: contributor_request_path(contributor_two, request))
+      find("a[href='#{contributor_request_path(contributor_two, request)}']", text: 'nachfragen').trigger('click')
+
```
mattwr18, line 172: was getting a `Exception: Capybara::Cuprite::MouseEventFailed` for overlapping elements. In this case, I think this is just because we are running in headless and I know that the navbar is not overlapping this link in a real browser. Using `trigger('click')` ignores potentially overlapping elements.

README.md
```diff
@@ -75,6 +75,15 @@ bundle config local.threema .
...
 See this [blog post](https://rossta.net/blog/how-to-specify-local-ruby-gems-in-your-gemfile.html) for more information.
 
+### Testing
+
+To run tests in a Docker container, you can run:
+
+```bash
+$ docker compose exec -it app bash
+$ RAILS_ENV=test rspec
+```
```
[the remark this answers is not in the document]
tillprochaska: `docker compose exec app bin/rspec` should be enough!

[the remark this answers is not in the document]
mattwr18: Great! I tested that out. When I was bashing into the container and running it without the `RAILS_ENV=test` it was running in development mode.

PR #1419, opened 2022-07-27
Add simple toolbar with button to insert placeholder

### LINKED ISSUE #1407: Button inserting {first_name}

mattwr18:
Fixes #1407

### DIFF
--- a/app/components/request_form/request_form.css
+++ b/app/components/request_form/request_form.css
@@ -10,6 +10,19 @@
   top: calc(3.5 * var(--spacing-unit));
 }
 
+.RequestForm-messageTextAreaToolbar {
+  width: 100%;
+  border: 2px solid transparent;
+  border-radius: 5px;
+  padding: 6px;
+  background-color: var(--color-gray-light);
+  margin: var(--spacing-unit-s) 0;
+}
+
+.RequestForm-insertPlaceholderButton {
+  background-color: #fff;
+}
+
 [data-request-form-target='preview'] {
   white-space: pre-wrap;
 }
--- a/app/components/request_form/request_form.css
+++ b/app/components/request_form/request_form.css
@@ -10,18 +10,9 @@
   top: calc(3.5 * var(--spacing-unit));
 }
 
-.RequestForm-messageTextAreaToolbar {
-  width: 100%;
-  border: 2px solid transparent;
-  border-radius: 5px;
-  padding: 6px;
-  background-color: var(--color-gray-light);
-  margin: var(--spacing-unit-s) 0;
-}
-
 .RequestForm-insertPlaceholderButton {
   background-color: #fff;
-  font-size: 1rem;
+  font-size: var(--font-size-s);
 }
 
 [data-request-form-target='preview'] {
--- a/app/components/request_form/request_form.css
+++ b/app/components/request_form/request_form.css
@@ -21,6 +21,7 @@
 
 .RequestForm-insertPlaceholderButton {
   background-color: #fff;
+  font-size: 1rem;
 }
 
 [data-request-form-target='preview'] {
--- a/app/components/request_form/request_form.html.erb
+++ b/app/components/request_form/request_form.html.erb
@@ -17,8 +17,8 @@
           <div
             class='RequestForm-messageTextAreaToolbar'
             role='toolbar'
-            aria-label='Text Formatting'
-            aria-control='message-textarea'
+            aria-label='Symbolleiste'
+            aria-control='request[text]'
           >
             <%= c 'button',
               style: :secondary,
@@ -29,7 +29,6 @@
             %>
           </div>
           <%= c 'textarea', field.input_defaults.merge(
-            id: 'message-textarea',
             required: true,
             show_emoji_picker_hint: true,
             highlight_placeholders: true,
--- a/app/components/request_form/request_form.html.erb
+++ b/app/components/request_form/request_form.html.erb
@@ -24,7 +24,7 @@
               style: :secondary,
               class: 'RequestForm-insertPlaceholderButton',
               type: 'button',
-              label: 'Vornamen einfügen',
+              label: t('components.request_form.insert_placeholder_button'),
               data: { action: 'click->request-form#insertPlaceholderAtCursor' }
             %>
           </div>
--- a/app/components/request_form/request_form.html.erb
+++ b/app/components/request_form/request_form.html.erb
@@ -14,12 +14,22 @@
         <% end %>
 
         <%= c 'field', object: @request, attr: :text do |field| %>
-          <%= c 'button',
-            type: 'button',
-            label: 'Insert {{FIRST_NAME}}',
-            data: { action: 'click->request-form#insertPlaceholderAtCursor' }
-          %>
+          <div
+            class='RequestForm-messageTextAreaToolbar'
+            role='toolbar'
+            aria-label='Text Formatting'
+            aria-control='message-textarea'
+          >
+            <%= c 'button',
+              style: :secondary,
+              class: 'RequestForm-insertPlaceholderButton',
+              type: 'button',
+              label: 'Vornamen einfügen',
+              data: { action: 'click->request-form#insertPlaceholderAtCursor' }
+            %>
+          </div>
           <%= c 'textarea', field.input_defaults.merge(
+            id: 'message-textarea',
             required: true,
             show_emoji_picker_hint: true,
             highlight_placeholders: true,
--- a/app/components/request_form/request_form.html.erb
+++ b/app/components/request_form/request_form.html.erb
@@ -14,20 +14,15 @@
         <% end %>
 
         <%= c 'field', object: @request, attr: :text do |field| %>
-          <div
-            class='RequestForm-messageTextAreaToolbar'
-            role='toolbar'
-            aria-label='Symbolleiste'
-            aria-control='request[text]'
-          >
+          <%= c 'toolbar', 'aria-controls': 'request[text]' do %>
             <%= c 'button',
               style: :secondary,
               class: 'RequestForm-insertPlaceholderButton',
               type: 'button',
-              label: t('components.request_form.insert_placeholder_button'),
+              label: t('.insert_placeholder_button'),
               data: { action: 'click->request-form#insertPlaceholderAtCursor' }
             %>
-          </div>
+          <% end %>
           <%= c 'textarea', field.input_defaults.merge(
             required: true,
             show_emoji_picker_hint: true,
--- a/app/components/request_form/request_form.html.erb
+++ b/app/components/request_form/request_form.html.erb
@@ -14,6 +14,11 @@
         <% end %>
 
         <%= c 'field', object: @request, attr: :text do |field| %>
+          <%= c 'button',
+            type: 'button',
+            label: 'Insert {{FIRST_NAME}}',
+            data: { action: 'click->request-form#insertPlaceholderAtCursor' }
+          %>
           <%= c 'textarea', field.input_defaults.merge(
             required: true,
             show_emoji_picker_hint: true,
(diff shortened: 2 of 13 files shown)


app/components/request_form/request_form.js
```diff
@@ -15,11 +15,22 @@ export default class extends Controller {
     this.updateMembersCount();
   }
 
+  insertPlaceholderAtCursor() {
+    const placeholder = '{{VORNAME}}';
```
mattwr18, line 19: I favored adding the placeholder to the translations file like we discussed @phoeinx ...
At the moment we still have these two instances in JS land where they are hardcoded. I was looking into how to do this in a rails app and there is a gem [i18n-js](https://github.com/fnando/i18n-js) that we can use, but honestly, I'm not sure it's worth adding the complexity and another gem for a theoretical use case.

I really like that we have set up the project to make it easy to internationalize cause I have worked on a project where we need to find all hardcoded strings and add them to translation files and it was a lot of work, and not the fun type. However, to my knowledge we haven't ever had a use case where the application itself should be in another language and in this particular case it's not as easy as just moving the string to the translation file.

What do you think though?

app/components/request_form/request_form.js
```diff
@@ -15,11 +15,22 @@ export default class extends Controller {
...
+    this.messageTarget.dispatchEvent(event);
+    this.messageTarget.focus();
+  }
+
   updatePreview() {
     let message = sanitize(this.messageTarget.value);
     message = message || this.previewFallbackValue;
-    message = replacePlaceholder(message, 'FIRST_NAME', 'Max');
-
+    message = replacePlaceholder(message, 'VORNAME', 'Max');
```
phoeinx, line 33: maybe refactor 'VORNAME' into a placeholder variable like in `insertPlaceholderAtCursor`

app/components/request_form/request_form.html.erb
```diff
@@ -14,6 +14,20 @@
         <% end %>
 
         <%= c 'field', object: @request, attr: :text do |field| %>
+          <div
+            class='RequestForm-messageTextAreaToolbar'
+            role='toolbar'
+            aria-label='Symbolleiste'
```
phoeinx, line 20: not sure if Toolbar might be easier to understand or if that is just my missing knowledge :D

app/components/request_form/request_form.html.erb
```diff
@@ -14,6 +14,20 @@
         <% end %>
 
         <%= c 'field', object: @request, attr: :text do |field| %>
+          <div
```
phoeinx, line 17: I feel like this is something that could also be a component (e.g. `toolbar`). If I remember it correctly you and @tillprochaska had some discussions on that and I am not sure what the result was. 
Anyhow nice accessibility features!

app/components/request_form/request_form.css
```diff
@@ -10,6 +10,20 @@
...
+  border: 2px solid transparent;
+  border-radius: 5px;
+  padding: 6px;
+  background-color: var(--color-gray-light);
+  margin: var(--spacing-unit-s) 0;
+}
+
+.RequestForm-insertPlaceholderButton {
+  background-color: #fff;
+  font-size: 1rem;
```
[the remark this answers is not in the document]
mattwr18: I don't like the idea of text being smaller than `1rem/16px` because this is generally accepted as the minimum font-size for `a11y`. See [Minimum font size?](https://accessibleweb.com/question-answer/minimum-font-size/)

> While there is no official minimum font size for the web, it is generally agreed upon that 16px for body text is a good starting point.

I personally would like to scale up the font-size to meet this suggestion, but not sure how others feel.

I did make the font-size by adding CSS to decrease the font-size from our normal buttons.
One thing we talked about potentially doing is using an icon with a tooltip that contains the text.

app/components/request_form/request_form.css
```diff
@@ -10,6 +10,20 @@
   top: calc(3.5 * var(--spacing-unit));
 }
 
+.RequestForm-messageTextAreaToolbar {
+  width: 100%;
+  border: 2px solid transparent;
```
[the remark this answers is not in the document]
mattwr18: So, to be honest, this was a copy/:spaghetti: from [Toolbar Example](https://www.w3.org/WAI/ARIA/apg/example-index/toolbar/toolbar.html), which illustrates the[ roving tabindex method](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex). It is originally `transparent`, but the `border-color` is set on focus so that when one is navigating the page with tabs, it is clear where they are. You also jump to the first item in the toolbar with tab and move to the next element outside the toolbar with tab. If you want to use your keyboard to navigate the toolbar itself, you would then switch to the right/left arrows.

One, I did not set up the focus styles, which I should probably do since it is part of `a11y`. Two, I don't think we really want to add the complexity that comes with the `roving tabindex method` at this point. Especially since we only have one item in the toolbar, but it's something to keep in mind.

[Keyboard Accessibility](https://webaim.org/techniques/keyboard/)I guess is important for us to keep in mind while building our site, as it:

>  is one of the most important aspects of web accessibility.

app/components/request_form/request_form.js
mattwr18: sure, good idea!

app/components/request_form/request_form.html.erb
mattwr18: I just translated it in Deepl cause I guess we want to make it a11y for German users? Or, we want to add this also to the translation files, I guess. Do you just call it toolbar in German?

app/components/request_form/request_form.html.erb
mattwr18: Also like that idea, maybe it can be used/improved upon

tillprochaska, line 17: Is there a specific reason for using single quotes in the HTML?

app/components/request_form/request_form.css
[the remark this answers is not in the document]
phoeinx: Alright, I see the accessibility point here. However, we currently have two system font sizes that are smaller than `1rem`:
`--font-size-s: 0.85em; 
--font-size-xs: 0.75em;`

The help text to whose font size I proposed matching the button font size e.g. has `--font-size-s`. I am definitely open to discuss increasing our font sizes to a minimum of `1rem` (I don't see any down sides), but I would like to do this in a consistent effort to avoid mismatching proportions.

I therefore adapted my suggestion to `font-size: var(--font-size-s)` and would propose also opening an issue about increasing our system font sizes to a minimum of `1rem`(luckily with the font-size variables that would be easy to change).

app/components/request_form/request_form.css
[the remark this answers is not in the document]
phoeinx: Hi, I see where that came from and I agree that keyboard accessibility is important and just plainly improves usability. 
I am attaching two screenshots of navigating our form via keyboard (the screenshots are one tab away). In our case the button within the toolbar gets into focus immediately instead of first focusing on the toolbar. Works well for me and I agree that the `roving tabindex method` currently appears overkill, even though its a good thing to keep in the back of our minds! I think the transparent border of the toolbar element, therefore, isn't needed currently but I also agree with you that we should set the `focus` styles of the button to match the input elements `focus` styles. I'll add some suggestions.

image
image

phoeinx, line 15:
```suggestion
```

app/components/request_form/request_form.css
```diff
@@ -10,6 +10,20 @@
...
+  padding: 6px;
+  background-color: var(--color-gray-light);
+  margin: var(--spacing-unit-s) 0;
+}
+
+.RequestForm-insertPlaceholderButton {
+  background-color: #fff;
+  font-size: 1rem;
+}
+
```
phoeinx, line 26:
```suggestion

.RequestForm-insertPlaceholderButton:focus {
  border-color: var(--color-focus);
  outline: none;
  box-shadow: var(--input-focus-shadow);
}

```

app/components/request_form/request_form.css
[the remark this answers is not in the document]
phoeinx: Added an issue: https://github.com/tactilenews/100eyes/issues/1444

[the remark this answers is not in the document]
mattwr18: I think that can be a nice compromise, though I am a bit hesitant cause in my experience there almost always is a lack of interest in prioritizing this kind of change.

app/components/request_form/request_form.css
tillprochaska: Why do we need this? If the focus styles don’t work for some reason, we might want to consider to update them for all of our buttons.

mattwr18: Turns out that we don't need this cause we use `	postcss-focus-within` to add the focus styles and we just needing to add a `tabindex='0'` to the `<div>`

app/components/request_form/request_form.html.erb
mattwr18: What do you mean by this? If I look at the `RequestForm` component, there is actually a mixture of double and single quotes. Just briefly glancing this is the case with most of our HTML. We don't enforce any style guide so this is probably what has led to the inconsistency.

We do enforce single quotes with `rubocop` in ruby land, so it makes sense, I think.

app/components/request_form/request_form.css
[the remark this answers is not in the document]
mattwr18: Addressed this in a comment :point_down:

app/components/request_form/request_form.css
```diff
@@ -10,6 +10,20 @@
   top: calc(3.5 * var(--spacing-unit));
 }
 
+.RequestForm-messageTextAreaToolbar {
+  width: 100%;
+  border: 2px solid transparent;
+  border-radius: 5px;
+  padding: 6px;
```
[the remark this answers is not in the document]
mattwr18: good catch... sorry, I didn't add this. I am very much in favor of variables over hardcoded values

app/components/request_form/request_form.js
mattwr18: Do you think we should add a `TODO:` to be able to find these easier in the future?

phoeinx: Good point. Imo as we don't really have it as a TODO on the horizon I think it's fine without a `TODO:`. Less clutter currently and if we only have to search in `.js` files for `i18n`able text I think that's an acceptable search space.

app/components/request_form/request_form.css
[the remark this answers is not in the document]
phoeinx: Let's start a discussion on the issue then. (I think its cleaner and more impactful to have this discussion for the whole application.)

app/components/request_form/request_form.html.erb
phoeinx: The way I understood it was double quotes for attributes in plain HTML and single quotes for "ruby enhanced HTML".

app/components/request_form/request_form.css
phoeinx: I agree that the focusing behaviour from the example you reference is nice for big toolbars. As long as our toolbar only has one or two or three elements I think tabbing through is probably fine.

Also in regards to the button focus behaviour, summing up Matts` and Tills' comments, I also think we should update the button focus behaviour.

It's currently defined as

```
.Button:focus {
  outline: none;
  box-shadow: var(--input-focus-shadow);
}
```

in `button.css`. And I would propose updating it with a focus-color border-color to:
```
.Button:focus {
  border-color: var(--color-focus);
  outline: none;
  box-shadow: var(--input-focus-shadow);
}
```

This makes focusing behaviour between input elements and buttons consistent.

phoeinx: I'll add a short PR with that.

phoeinx (approved):

PR #1421, opened 2022-07-28
Silence non-actionable 50X errors from Signal server

### LINKED ISSUE #1420: Silence non-actionable 502 Bad Response errors from Signal
We receive quite a few errors in Sentry, see [100EYES-3B](https://sentry.io/organizations/tactilenews/issues/3180081580/?project=5686596&referrer=slack) whenever the Signal servers are down. These are transient errors that normally go away quite quickly, but they spam our #monitoring channel is Slack and take up quota in our Sentry issues.

We already have Oh Dear monitoring that checks if we are not able to poll the receive in Signal so this is much more useful for us to pay attention to as this is indicates that we might need to search for an explanation that would have a longer disruption in our ability to offer the Signal service to our customers.

For this reason, we would like to rescue these specific errors and not report them in Sentry.

mattwr18:
This raises an error and then configures Sentry to ignore it since we have a monitoring task set up to alert us if there is an issue for more than 8 minutes

Fix #1420

### DIFF
--- a/app/jobs/signal_adapter/receive_polling_job.rb
+++ b/app/jobs/signal_adapter/receive_polling_job.rb
@@ -33,7 +33,12 @@ module SignalAdapter
       signal_messages.each do |raw_message|
         adapter.consume(raw_message) { |m| m.contributor.reply(adapter) }
       rescue StandardError => e
-        ErrorNotifier.report(e)
+        ErrorNotifier.report(e, context: {
+          code: e.response.code,
+          message: e.response.message,
+          headers: e.response.to_hash,
+          body: e.response.body
+        })
       end
 
       ping_monitoring_service && return
--- a/app/jobs/signal_adapter/receive_polling_job.rb
+++ b/app/jobs/signal_adapter/receive_polling_job.rb
@@ -44,7 +44,7 @@ module SignalAdapter
     def request_new_messages
       url = URI.parse("#{Setting.signal_cli_rest_api_endpoint}/v1/receive/#{Setting.signal_server_phone_number}")
       res = Net::HTTP.get_response(url)
-      raise SignalAdapter::ServerError if (500..509).include? res.code.to_i
+      raise SignalAdapter::ServerError if res.code.to_i.eql?(400)
 
       JSON.parse(res.body)
     end
--- a/app/jobs/signal_adapter/receive_polling_job.rb
+++ b/app/jobs/signal_adapter/receive_polling_job.rb
@@ -13,7 +13,7 @@ module SignalAdapter
     def perform(*_args)
       return if Setting.signal_server_phone_number.blank?
 
-      @signal_messages = request_new_messages
+      signal_messages = request_new_messages
       adapter = SignalAdapter::Inbound.new
 
       adapter.on(SignalAdapter::CONNECT) do |contributor|
@@ -30,13 +30,16 @@ module SignalAdapter
         SignalAdapter::Outbound.perform_later(recipient: contributor, text: Setting.signal_unknown_content_message)
       end
 
-      consume_signal_messages
+      signal_messages.each do |raw_message|
+        adapter.consume(raw_message) { |m| m.contributor.reply(adapter) }
+      rescue StandardError => e
+        ErrorNotifier.report(e)
+      end
+
       ping_monitoring_service && return
     end
 
     private
-
-    attr_reader :signal_messages
 
     def request_new_messages
       url = URI.parse("#{Setting.signal_cli_rest_api_endpoint}/v1/receive/#{Setting.signal_server_phone_number}")
@@ -55,19 +58,6 @@ module SignalAdapter
 
     def queue_empty?
       Delayed::Job.where(queue: queue_name, failed_at: nil).none?
-    end
-
-    def consume_signal_messages
-      signal_messages.each do |raw_message|
-        adapter.consume(raw_message) { |m| m.contributor.reply(adapter) }
-      rescue StandardError => e
-        ErrorNotifier.report(e, context: {
-                               code: e.response.code,
-                               message: e.response.message,
-                               headers: e.response.to_hash,
-                               body: e.response.body
-                             })
-      end
     end
   end
 end
--- a/app/jobs/signal_adapter/receive_polling_job.rb
+++ b/app/jobs/signal_adapter/receive_polling_job.rb
@@ -44,7 +44,7 @@ module SignalAdapter
     def request_new_messages
       url = URI.parse("#{Setting.signal_cli_rest_api_endpoint}/v1/receive/#{Setting.signal_server_phone_number}")
       res = Net::HTTP.get_response(url)
-      raise SignalAdapter::ServerError if res.code.to_i.eql?(400)
+      raise SignalAdapter::ServerError if res.instance_of?(Net::HTTPBadRequest)
 
       JSON.parse(res.body)
     end
--- a/app/jobs/signal_adapter/receive_polling_job.rb
+++ b/app/jobs/signal_adapter/receive_polling_job.rb
@@ -13,7 +13,7 @@ module SignalAdapter
     def perform(*_args)
       return if Setting.signal_server_phone_number.blank?
 
-      signal_messages = request_new_messages
+      @signal_messages = request_new_messages
       adapter = SignalAdapter::Inbound.new
 
       adapter.on(SignalAdapter::CONNECT) do |contributor|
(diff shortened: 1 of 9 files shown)


app/jobs/signal_adapter/receive_polling_job.rb
```diff
@@ -44,6 +44,8 @@ def perform(*_args)
     def request_new_messages
       url = URI.parse("#{Setting.signal_cli_rest_api_endpoint}/v1/receive/#{Setting.signal_server_phone_number}")
       res = Net::HTTP.get_response(url)
+      raise SignalAdapter::ServerError if (500..509).include? res.code.to_i
```
tillprochaska, line 47: I think this won’t work. `signal-cli-rest-api` doesn’t return a 5xx response in this case. Instead, `signal-cli-rest-api` returns an exception that indicates that the Signal service has responded with a 5xx status code.

I also think we could maybe handle this with less complexity? I’d suggest simply raising an exception in case of a non-successful response from `signal-cli-rest-api`.

```suggestion
      res.value # raises an exception if response code is not 2xx
```

… then rescuing from 4xx response to ensure the job completes successfully and no errors are reported.

```ruby
 rescue_from(Net::HTTPClientException) do |exception|
    # In case the request fails, it’s likely due to the Signal service being
    # unavailable for a short period. As the polling job is run repeatedly in
    # a short interval, we simply ignore this error. If receiving messages fails
    # persistently, we will be notified by our cron monitoring service. Note that 
    # `signal-cli-rest-api` returns a 400 error (even if a 500 error would be more
    # appropriate semantically).
  end
```

spec/jobs/signal_adapter/receive_polling_job_spec.rb
```diff
@@ -164,5 +165,31 @@
...
+      let(:error_message) do
+        [['error', "Error while checking account #{Setting.signal_server_phone_number}: [502] Bad response: 502 \n"]].to_json
+      end
+
+      before do
+        create(:request)
+
+        unless Setting.signal_server_phone_number
+          allow(Setting).to receive(:signal_server_phone_number).and_return('SIGNAL_SERVER_PHONE_NUMBER')
+        end
```
tillprochaska, line 179: Why do you stub the setting only if it’s not set?

spec/jobs/signal_adapter/receive_polling_job_spec.rb
```diff
@@ -164,5 +165,31 @@
...
+      it 'raises an SignalAdapter::ServerError' do
+        expect { subject.call }.to raise_error(SignalAdapter::ServerError)
+      end
+
+      it 'stops immediately as a server error occurred' do
+        subject.call
+      rescue SignalAdapter::ServerError
+        expect(SignalAdapter::Inbound).not_to receive(:new)
+      end
+    end
```
tillprochaska, line 193: We should also test that we do *not* ping the monitoring service in this case.

tillprochaska (changes requested):

spec/jobs/signal_adapter/receive_polling_job_spec.rb
mattwr18: It's a copy/:spaghetti: from the test setup on https://github.com/tactilenews/100eyes/blob/main/spec/jobs/signal_adapter/receive_polling_job_spec.rb#L44-L52, which was added by you. To be honest, I didn't really think too much about why you added it, I just moved the specs I added to their own block cause I didn't want to use the vcr, but I knew they were working with the test setup.

app/jobs/signal_adapter/receive_polling_job.rb
mattwr18: > I think this won’t work. signal-cli-rest-api doesn’t return a 5xx response in this case. Instead, signal-cli-rest-api returns an exception that indicates that the Signal service has responded with a 5xx status code.

Yeah, I remember that from the docs, now that you mention it.

> also think we could maybe handle this with less complexity? I’d suggest simply raising an exception in case of a non-successful response from signal-cli-rest-api.

I thought about just rescuing, but I favored the "more complex", but more explicit way of handling it. Note, that we could handle this specific error by just adding a guard clause to the `SignalAdapter::Inbound#perform` method that returns if the `signal_message` does not respond_to? :with_indifferent_access. This would be less complex than both options.

But again, I think setting up a class for `IgnoredErrors` is more explicit and I can imagine we might use it in other places where we want to ignore third party servers not being available for some period of time. I also don't like calling `res.value` because we know that it might raise an exception and then needed to add a comment to explain what we are trying to do. In general, if you feel the need to write a comment to yourself or other devs, maybe there is a more explicit way to handle the same thing.

spec/jobs/signal_adapter/receive_polling_job_spec.rb
mattwr18: I originally had that test, but I wanted to actually check that the next line after `request_new_messages` doesn't get called, but if you think it's important to be super sure that we don't ping our monitoring service since we rely on that for error reporting after 8 minutes, I can add it back.

spec/jobs/signal_adapter/receive_polling_job_spec.rb
tillprochaska: I think my name’s in git blame only due to some reformatting/naming changes. Seems like Robert added this originally. This shouldn’t block the PR from getting merged, but maybe you could enlighten us why that’s necessary, @roschaefer? Thanks! :)

app/jobs/signal_adapter/receive_polling_job.rb
tillprochaska: > I thought about just rescuing, but I favored the "more complex", but more explicit way of handling it.

Fine with me if you prefer that!

> and then needing to add a comment to explain what we are trying to do. In general, if you feel the need to write a comment to yourself or other devs, maybe there is a more explicit way to handle the same thing.

I thought it might be valuable to add a comment to explain *why* we ignore the error (and why we get a 4xx error when the Signal service is unavailable), because that’s something that isn’t intuitive. The reason for that lies in a third-party component, so no matter how well our own code is structured, it will still be confusing.

tillprochaska (approved):

mattwr18:
> I thought it might be valuable to add a comment to explain why we ignore the error (and why we get a 4xx error when the Signal service is unavailable), because that’s something that isn’t intuitive. The reason for that lies in a third-party component, so no matter how well our own code is structured, it will still be confusing.

Sorry, I didn't mean the comment in the rescue, I can understand why that would be helpful, after all we are rescuing and then doing nothing. The comment helps to explain why we aren't doing anything. I was talking about the comment that we have to explain that calling `res.value` on anything other than a `200` throws an error. See [this example](https://github.com/tactilenews/100eyes/blob/main/app/adapters/signal_adapter/outbound.rb#L37) of where we use a comment like this.

This to me obfuscates what is happening, it's not explicit that calling `res.value` would throw an error in case of anything other response code.

app/jobs/signal_adapter/receive_polling_job.rb
```diff
@@ -44,6 +44,8 @@ def perform(*_args)
     def request_new_messages
       url = URI.parse("#{Setting.signal_cli_rest_api_endpoint}/v1/receive/#{Setting.signal_server_phone_number}")
       res = Net::HTTP.get_response(url)
+      raise SignalAdapter::ServerError if res.code.to_i.eql?(400)
```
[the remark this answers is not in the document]
mattwr18: I like this suggestion, but I think `instance_of?` might be more appropriate here? `kind_of?` or `'is_a?` would return true if `Net::HTTPBadRequest` were the parent of the response, to my understanding, but `instance_of?` only if it were actually an instance of `Net::HTTPBadRequest`.

As far as I can tell, `Net::HTTPBadRequest` does not have any children, so I think we really want to check that the response is an instance of the error. Or is there some reason you would favor `kind_of?` in this case?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_03_25_112312) do
+ActiveRecord::Schema.define(version: 2022_05_19_185616) do
```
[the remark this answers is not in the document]
mattwr18: Kind of... When I try to run the tests locally, I am getting this error.

```
== 20220519185616 RemoveHintsFromRequest: migrating ===========================
-- remove_column(:requests, :hints)
ERROR:  column "hints" of relation "requests" does not exist
STATEMENT:  ALTER TABLE "requests" DROP COLUMN "hints"
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

PG::UndefinedColumn: ERROR:  column "hints" of relation "requests" does not exist
```
I have gotten past this error in 2 different ways:

1) by adding `if_exists: true` to the `db/migrate/20220519185616_remove_hints_from_request.rb` and running the migrations - doesn't feel like the best way cause the column did exist so it shouldn't throw this error.
2) running `bundle exec rails db:drop db:create db:migrate RAILS_ENV=test`

Both of these update the schema version. The schema version is `2022_03_25_112312`, which corresponds to the `AddAdditionalConsentToContributors` migration `/home/mattwr18/code/100eyes/db/migrate/20220325112312_add_additional_consent_to_contributors.rb`. The schema version was not updated when the `RemoveHintsFromRequest` was added. As you are the one that added this migration, was this an oversight? Or did you purposely not update the schema version for some reason?

I am also able to get past the error by running `bundle exec rails db:setup RAILS_ENV=test` and it does not update the schema version because it loads the schema instead of running the migrations, but I'm curious if/when we update our schema version. Looking at the history of the file, it seems that we are updating it with each new migration that is added. So, I assume this is a change that we actually want to keep, would you prefer it in a separate PR?


 

Every time I run the migrations, even if it is just for the test db, the `db/schema` updates and I typically don't commit it, but in this case I decided to just push it up cause I was tired of rolling back the change.

PR #1428, opened 2022-08-05
Update Signal onboarding workflow to handle sender identification by uuid

phoeinx:


app/jobs/signal_adapter/create_contact_job.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'net/http'
+
+module SignalAdapter
+  class CreateContactJob < ApplicationJob
+    queue_as :create_signal_contact
+
```
tillprochaska, line 8: Is there a particular reason for using a separate queue for these? If not, I think I’d simply use the default queue.

```suggestion
```

tillprochaska (approved):

PR #1431, opened 2022-08-17
Improvements to ActivityNotifications feed

mattwr18:
Based on user feedback from LZ


app/models/message.rb
```diff
@@ -58,15 +59,17 @@ def chat_message_link
...
   private
 
   def notify_recipient
-    return unless reply?
-
-    MessageReceived.with(contributor: sender, request: request).deliver_later(User.all)
+    if reply?
+      MessageReceived.with(contributor: sender, request: request, message: self).deliver_later(User.all)
+    elsif !broadcasted?
+      ChatMessageSent.with(contributor: recipient, request: request, user: Current.user, message: self).deliver_later(User.all)
```
mattwr18, line 72: @tillprochaska maybe we can discuss this? I found it quite surprising that we don't save any information on which person in the newsroom has sent a "chat messge" to a contributor.

I understand that we save everything as a `Message` and then we do a slightly complicated way of deciding if it's a `broadcasted?`,`reply?` or a `manually_created?` message, but we don't save any information on which User sent the message.

I guess we could not save it in the `ActivityNotification` as well, then we could completely remove the `ActiveSupport::CurrentAttributes`. At present, we use this information to display which User sent out the chat message.

PR #1432, opened 2022-08-17
Group MessageReceived notifications

mattwr18:
To reduce the noise of the notifications


tillprochaska:
Hi @mattwr18, so this is what I thought of to simplify the grouping.

For each notification class (e.g. `MessageReceived`), we could add a `group_key` method:

```ruby
class MessageReceived < Noticed::Base
  def group_key
    "message-received-#{request.id}
  end
end
```

For notifications that shouldn’t be grouped, you can simply define the `group_key` in a way that ensures that every notification has a unique key. For example, there will be exactly one `OnboardingCompleted` notification per contributor:

```ruby
class OnboardingCompleted < Noticed::Base
  def group_key
    "onboarding-completed-#{contributor.id}
  end
end
```

In the controller, you can then simply group notifications like this:

```ruby
groups = user.notifications.newest_first.group_by(&:group_key)
```

tillprochaska:
In order to format message nicely, we could define a class method on each notification class…

```ruby
class MessageReceived < Noticed::Base
  def self.group_message(notifications:)
    t('.message',
      request: notifications.first.request.title,
      contributor_one: notifications.first.contributor.name,
      contributor_two: notifications.second&.contributor.name
      count: notifications.size,
    )
  end
end
```

… and use it like this:

```ruby
class NotificationGroup
  def message
    notifications.first.class.message(notifications)
  end
end
```

Given translations as follows, I think we would be able to generate messages in a way that isn’t too complex and that works in a generic way, i.e. would work out of the box for new notificatoin types as well.

```yml
message_received:
  one: '%{contributor_one} has replied to request “%{request}”'
  two: '%{contributor_one} and %{contributor_two} have replied to request “%{request}”'
  other: '%{contributor_one} and %{count} others have replied to request ”%{request}”'
```

mattwr18:
Hey @tillprochaska thanks for writing down what you were thinking of. I have given your approach a go and have some thoughts.

- I really like the idea of adding a `group_key` to the individual `ActivityNotification` types' classes. This has indeed simplified the logic.

- As far as adding a `NotificationGroup`, I got it working that way, but realized that it could be simplified by removing it and just calling a helper method directly from the controller to the  individual `ActivityNotification` types' classes. This meant that I did not need to access current_user from the helper methods as the notifications where already scoped when I passed them into the `group_message` method.

- Regarding the benefit of the `NotificationGroup` in being able to add helper methods, I don't see that as necessary since we already have the  individual `ActivityNotification` types' classes.

- As far as formatting messages, I had not added the `:two` key and customized message, but another benefit of your idea is that we have access to the second contributor name. I have now added this new message and test cases.

Overrall, I am quite happy with your help! Please have a look and let me know what you think

See ef54d83b106995911aa1ccefbf0d21f15b654cbc and f8bc359c

PR #1433, opened 2022-08-19
Add basic support for multi-lingual onboarding

mattwr18:


mattwr18:
Closing this as it is really out of date and the `Riff` project has finished.
If we require support for multi-lingual onboarding in the future, we can re-open or revisit how we want to handle it.

Issue #56, opened 2020-05-18
Secondary-Email-Adress


drjakob:
As Lokalexperte
I want that all my e-mail-adresses work with 100eyes
In order to minimize error messages.

PR #1434, opened 2022-08-26
chore(deps): update alpine, keep postgres-12

### LINKED ISSUE #1411: Update Alpine version
We’re using an outdated version of Alpine.

For context, see #1410 and #1409.

roschaefer:
Install it from from the 3.12 repository according to this stackoverflow link:
https://stackoverflow.com/a/63976418/2069431

I don't know if this is considered bad practice though. But that should
fix our immediate problem.

close #1411

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:3.0.2-alpine3.12
+FROM ruby:3.0.2-alpine3.14
 
 ARG git_commit_sha
 ARG git_commit_date
@@ -10,14 +10,17 @@ RUN apk --update add \
     git \
     nodejs \
     npm \
-    postgresql-dev=~12 \
-    postgresql-client=~12 \
     tzdata \
     libxslt-dev \
     libxml2-dev \
     imagemagick \
     less \
     libsodium
+
+RUN apk --update add \
+    postgresql-dev=~12 \
+    postgresql-client=~12 \
+    --repository=http://dl-cdn.alpinelinux.org/alpine/v3.12/main
 
 COPY entrypoint.sh /usr/bin/
 RUN chmod +x /usr/bin/entrypoint.sh


roschaefer:
```
$ docker compose exec -it app node --version
v14.20.0
```

Dockerfile
```diff
@@ -10,15 +10,18 @@ RUN apk --update add \
...
     libxml2-dev \
     imagemagick \
     less \
     libsodium
 
+RUN apk --update add \
+    postgresql-dev=~12 \
+    postgresql-client=~12 \
+    --repository=http://dl-cdn.alpinelinux.org/alpine/v3.12/main
+
```
mattwr18, line 24:
```suggestion
FROM postgres:12.2-alpine
COPY --from=alpine . .
```
I think it would be a better practice to use an official postgres image rather than installing postgres from this hardcoded source.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:3.0.2-alpine3.12
+FROM ruby:3.0.2-alpine3.14
```
mattwr18, line 1:
```suggestion
FROM ruby:3.0.2-alpine3.14 AS alpine
```

mattwr18 (approved):
@roschaefer I'm not sure whether it's considered a bad practice to do this, but `alpine3.14` has postgres 13 and we also have a ticket  [Upgrade Postgres version to 13 #56 ](https://github.com/tactilenews/100eyes-closed/issues/56), so maybe we could kill two birds with one stone?

Though, this would involve taking backups of all instances and upgrading, then restoring them, so I think we need to consider the pros and cons of upgrading running instances' dbs.

I made a suggestion below that you might consider.

Dockerfile
phoeinx: I agree that using hardcoded sources isn't that pretty. I am not sure if I understand your suggestion correctly - is it possible to copy everything from one alpine image into another alpine image as a multi-stage build or would that overwrite things?

mattwr18: Yes, exactly, it would be a multi-stage build

phoeinx: yeah I know, but I was wondering if you can copy one complete alpine image into another? The example given for [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds) copies a build/an artifact into an alpine image. 
When we copy everything I was wondering if we might e.g. overwrite library folders and therefore overwrite the postgres installation. Have you tried it locally? Cuz if it works I am good with that.

mattwr18: Oh, right, sorry now I understand your concern. I'm not sure if we override anything, but I did test it locally and it worked as expected. First, what I did was to try out without the hardcoded line pulling in v12 of postgres to verify it did not work as reported by @roschaefer. Then, I attempted it both how it is in the `PR` and with the proposed change and both worked, as in no errors happened and I was able to access and update the db.

This was also not an original idea of mine, but came from the [same commenter on SO](https://stackoverflow.com/questions/63939670/install-older-package-version-in-alpine/63976418#63976418) that @roschaefer referenced. 

> However, since 12.3 doesn't live in the official Alpine repositories anymore, you could rely on an external Docker image, instead.

> Luckily, the [postgres](https://hub.docker.com/_/postgres) official images has version tags, and you can find the desired Alpine image for 12.3:
Therefore, you can use FROM:postgres:12.3-alpine to get the desired version from.

Note, that he does not mention anything about multi-stage builds.

I also checked the v inside the app container:

```
$ psql -V
psql (PostgreSQL) 12.2
```

with the PR, the version we use, is not even the same version as we declare in https://github.com/tactilenews/100eyes/blob/main/docker-compose.yml#L24

```
$ psql -V
psql (PostgreSQL) 12.10
```
Also, the image size with the PR's changes is `1.32GB` vs the multi-stage build `1.14GB`

phoeinx: ok, if it all works out let's go with the multi-stage build 👍

phoeinx: thanks for testing that out!

phoeinx: not sure if @roschaefer has some opinions on that but otherwise I think it would also be fine if you would add your suggestions in and merge this PR @mattwr18

tillprochaska: Hi @phoeinx @mattwr18, tbh I don’t really understand what the issue with installing packages from a custom repository is?

Overriding random files in an image on the other hand is something that can have unintended side effects…

roschaefer: I think I agree here with @tillprochaska and I'm going to merge this as it is.

I hope we can also update our postgres version soon and then revert the change to the Docker image.

mattwr18:
hey @roschaefer I just wanted to ping you because I ended up reverting this since it completely broke our staging instance when it was deployed. I guess we still want to do this, but we need to make sure that we can deploy it to production instances without crashing the docker containers.

PR #1443, opened 2022-09-08
Add job to attach signal contributors avatar

### LINKED ISSUE #1429: Fetch Signal profile pictures during onboarding
During Signal onboarding, new contributors have to send an initial Signal message to the instance’s 100eyes phone number. By doing this, they also explicitly grant us access to their Signal profile picture. We could/should fetch the profile picture to display it in the 100eyes contributor profile.

mattwr18:
Closes #1429


spec/jobs/signal_adapter/attach_contributors_avatar_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+RSpec.describe SignalAdapter::AttachContributorsAvatar do
+  describe '#perform_later(contributor)' do
+    subject { -> { described_class.new.perform(contributor) } }
+    let(:contributor) { create(:contributor, signal_phone_number: '+491212343434') }
+
+    context 'with avatar on file system' do
+      let(:avatar) { file_fixture('profile-+491212343434') }
+      before do
+        allow(File).to receive(:file?).with("/app/signal-cli-config/avatars/profile-#{contributor.signal_phone_number}").and_return(true)
```
tillprochaska, line 13: Why are you using string interpolation to construct the path in this line, but not in line 16?

spec/jobs/signal_adapter/receive_polling_job_spec.rb
```diff
@@ -97,6 +98,12 @@
           subject.call
           expect(contributor.reload.signal_onboarding_completed_at).to be_present
         end
+
+        it 'enqueues a job to attach contributors avatar' do
+          expect(SignalAdapter::AttachContributorsAvatar).to receive(:perform_later).with(contributor)
+
+          subject.call
+        end
```
tillprochaska, line 106: I think instead of this you can simply do:

```suggestion
        it { should have_enqueued_job(SignalAdapter::AttachContributorsAvatar).with(contributor) }
```

https://relishapp.com/rspec/rspec-rails/docs/matchers/have-enqueued-job-matcher

app/jobs/signal_adapter/attach_contributors_avatar.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module SignalAdapter
+  class AttachContributorsAvatar < ApplicationJob
+    queue_as :attach_signal_avatar
+
+    def perform(contributor)
+      avatar = "/app/signal-cli-config/avatars/profile-#{contributor.signal_phone_number}"
+      return unless File.file?(avatar)
+
+      contributor.avatar.attach(io: File.open(avatar), filename: "#{contributor.name.delete(' ').underscore}_profile_pic")
```
tillprochaska, line 11: Are you sure `name.delete(' ').underscore` is enough to ensure we always get valid file names. Although I don’t think that ActiveStorage uses the filename for the actual files written to disk, having invalid file names might cause unexpected problems when switching the storage service or downloading files.

You could use this method from ActiveStorage: https://api.rubyonrails.org/classes/ActiveStorage/Filename.html#method-i-sanitized

tillprochaska (approved):

app/jobs/signal_adapter/attach_contributors_avatar.rb
mattwr18: I am not sure that there is a universally accepted "valid" filename. For example, I saw a lot of people dealing with issues when there are umlauts as part of the filename. See [Rails 6: Files with special characters in the name crash app when accessing amazon S3 with active storage](https://stackoverflow.com/questions/60223440/rails-6-files-with-special-characters-in-the-name-crash-app-when-accessing-amaz), for example.

But looking at the `ActiveStorage.new(filename).sanitized` method you linked to, you might be more interested in characters like `$`, in which case I think it would be helpful to sanitize them as you suggested since we don't validate our first and last names so we cannot be sure that they will not contain any "illegal" characters.

I also used [ActiveSupport::Inflector.transliterate](https://apidock.com/rails/v6.0.0/ActiveSupport/Inflector/transliterate) to address special characters. Maybe this is overkill, or maybe we shouldn't try to use the contributor's name at all to save as a filename.

I considered using `File.basename(avatar)`, but this would include a `+` character, which is considered a [Characters to Avoid in Filenames and Directories](https://www.mtu.edu/umc/services/websites/writing/characters-avoid/) so then we would have to call `.phony_formatted(spaces: '-')`

Don't know which is better, to be honest. I don't really care, except trying to avoid as many issues as possible for our future selves.

tillprochaska: How about just using the contributor ID as the filename?

mattwr18: Yeah, that's much simpler hahaha ... Here I was making it more complicated than it need be. Thanks :blush:

PR #1445, opened 2022-09-09
Update button focus behaviour to match that of input elements

phoeinx:
As discussed on the Placeholder Button PR let's improve the button focus behaviour and let it match the input element focus behaviour. 

To compare this is the input element focus behaviour:

```
.Input > input:focus {
  border-color: var(--color-focus);
  outline: none;
  box-shadow: var(--input-focus-shadow);
}
```

### DIFF
--- a/app/components/button/button.css
+++ b/app/components/button/button.css
@@ -30,6 +30,7 @@
 }
 
 .Button:focus {
+  border-color: var(--color-focus);
   outline: none;
   box-shadow: var(--input-focus-shadow);
 }


mattwr18 (approved):
Great! thanks for following up on this task!!
I guess it probably makes sense to explicitly set these styles so we don't rely on the user's browser's focus styles, which are different depending on browser. So, for example for the links in the navbar, we don't set focus styles.

Also, for the toolbar component I added, it would either be better to set the styles or remove the tabindex.

phoeinx:
@mattwr18  thanks for the review! Would you like to merge this PR as it is currently or would you want your proposed changes to the toolbar component added first?

mattwr18:
> @mattwr18 thanks for the review! Would you like to merge this PR as it is currently or would you want your proposed changes to the toolbar component added first?

Oh sorry, I am happy for it to get merged as is. It was more a note to me to clean up my own code. I typically am not in the habit of merging other people's PRs, but if you'd like me to if I approve, I can.

phoeinx:
Na, totally fine by me to merge my own PRs, but good to know that you prefer it that way. Afaik Robert e.g. prefers having his PRs merged after approval.

PR #1447, opened 2022-09-13
Rescue Postmark::InactiveRecipientError and tag with support

### LINKED ISSUE #1382: Handle `Postmark::InactiveRecipientError` exceptions in mail delivery jobs
Currently, these exceptions aren’t handled explicitly. The exceptions get reported to Sentry (which is a good thing), but they end up in our main monitoring channel.

Instead, we should handle this exception and report it to Sentry with an additional tag (e.g. support). We can then configure Sentry to send notifications about issues tagged with "support" into a separate channel without involvement from developers.

mattwr18:
Closes #1382

- [x] Add custom tag to Postmark::InactiveRecipientError events
- [x] Edit Alert rule in Sentry to stop reporting events with the tag `support: yes`
- [x] Create #support channel in Slack and add @FrauCsu and @drjakob 
- [x] Create Alert rule in Sentry to report these events to the #support channel in Slack
- [x] Manually test on staging.100ey.es

### DIFF
--- a/app/adapters/postmark_adapter/outbound.rb
+++ b/app/adapters/postmark_adapter/outbound.rb
@@ -10,7 +10,11 @@ module PostmarkAdapter
     def self.send!(message)
       return unless message.recipient&.email
 
-      with(message: message).message_email.deliver_later
+      begin
+        with(message: message).message_email.deliver_later
+      rescue Postmark::InactiveRecipientError => e
+        ErrorNotifier.report(e, context: { recipients: e.recipients }, tags: { type: 'support' })
+      end
     end
 
     def self.send_welcome_message!(contributor)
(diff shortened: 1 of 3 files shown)


mattwr18:
I originally spent time adding a test case for this. See [`5027037` (#1447)](https://github.com/tactilenews/100eyes/pull/1447/commits/50270379227a67b018450129fac2938770091ed5#diff-6bcb1e74ba29e4314b9b769253ab2ca15c5e810d83e5000908e27a4401cda242R169-R207), but I got it passing in a way that failed when manually testing. I was trying to use a simple `begin/rescue` block, but it wasn't working because the error was originating from the job scheduled to send the mail. The way to rescue from errors with `ActionMailer` is to use its `rescue_from` method, which is [extensively tested at the rails level.](https://github.com/rails/rails/blob/v6.1.6.1/activesupport/test/rescuable_test.rb) 

It still would be nice to test our implementation, but I could not find anywhere how to trigger an error that is rescued in this `rescue_from` method in the mailers. There are [some examples](https://relishapp.com/rspec/rspec-rails/docs/controller-specs/bypass-rescue) when using `rescue_from` in a controller, but in our case this is not helpful.

PR #1459, opened 2022-09-23
Convert sender to polymorphic and add User as sender

### LINKED ISSUE #1456: Save User who sent ChatMessage as part of Message
At the moment, we don't save the User who sent a ChatMessage to a Contributor. This was quite surprising to me because one of the benefits of having individual user logins is to know who is doing what in the application.

As part of the ActivityNotification, we want to display who sent out the ChatMessage and to get this information that we don't store on the record itself, we added `ActiveSupport::CurrentAttributes` to set the current_user and save it on the notifications record. This technique has been criticized as adding global state can have unintended consequences and even the [Rails documentation](https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html) says to be cautious.

> A word of caution: It's easy to overdo a global singleton like Current and tangle your model as a result. Current should only be used for a few, top-level globals, like account, user, and request details. The attributes stuck in Current should be used by more or less all actions on all requests. If you start sticking controller-specific attributes in there, you're going to create a mess.

We could easily solve this, and buy other features, like displaying the User who sent out the ChatMessage from the `contributor_request_path` if we simply save this information on the `Message` record.

mattwr18:
Closes #1456 

This PR changes the `sender` of a Message to be a polymorphic association between Contributor and User. This seems like the most semantic way to deal with it, but not the least complex. I considered adding another attribute, like `user_id`, but I think the logic is not so clear. We need to look at the implementation to see that we save the `user_id` when a user sends out a Request/ChatMessage. Then we would still have `sender_id: nil` in a lot of cases. We could rename the `sender_id` to `contributor_id` and then have:

```
contributor_id
user_id
creator_id
```

We can still consider this if we like the simplicity of it vs it being less semantic. I don't really think that we would have other types of senders, in the future so maybe polymorphic associations are a bit overkill? One thing I still don't really understand, or keep forgetting is how we `manually_create` a `Message`. I know that when we do so we add the `creator_id`. I can see that the `MessagesController#update` and I can see we have a test that checks that if one sends a `PATCH` request to this method, it adds a creator_id, but I don't see where we actually do this from inside the application.

Looking at the data, it doesn't seem to be a very-used feature.

## NOTE to DEVS
When we deploy this, we need to run `lib/tasks/temporary/add_sender_type_to_messages.rake` to add the `sender_type` to existing Message records that have the `sender_id` already


mattwr18:
@tillprochaska I don't know that you will have time, but it would be nice to get your perspective on this since you made a comment when talking about the `activity_notifications` that you didn't think it was worth using polymorphic associations, at least for that case, because of their complexity.

PR #1472, opened 2022-09-27
Refactor PageHeader, add Main, Section components

### LINKED ISSUE #1454: Refactor PageHeader to be more flexible
The name of the component suggests it should be reusable as a generic page's header, but it's opinionated in that it has specific background color and padding. If we make this configurable, we can use it in many different places, like the new dashboard. Also, we might be able to consolidate some of the different headers we have, i.e `OnboardingHeader`, `ContributorHeader`

### LINKED ISSUE #1457: Create MainComponent
At the moment, for all pages that use the application layout, we wrap them in a `main` tag. In most cases, this is ok and has the benefit of not forgetting to add a main tag to the page. However, in some cases this inflexibility causes us to not follow semantic HTML and we wrap the entire page in a `div` to apply a background for example.

This means that we imported styles for the page view for the new dashboard, which we would like to avoid. Until now, we have a component-only style, so that a view is just a combination of one or more components, which are encapsulated and have their own templates, styles, etc.

We can solve both of these concerns by introducing a component that has a main tag and rendering the content passed into it.

mattwr18:
Closes #1454 
Closes #1457 

This PR makes the `PageHeader` component more flexible and uses it on every page.
It also introduces the `Main` component that can be used to set background color for a page, for example.
It introduces a `Section` component
It favors semantic HTML in components that were touched to add :point_up: components.

There are some doubts I have, it's not really the way I would do things if I were the only one making decisions. For every little variation to the `PageHeader` component, there are modifier classes added so that some have 4 or more classes. I am familiar with having a base component and if there is this level of customization, I typically just give it a class and add styles where it is used. Since we want to avoid stylesheets in the views and instead favor a component-only infrastructure, I think this is an ok compromise.


app/components/two_column_layout/two_column_layout.css
```diff
@@ -9,7 +9,7 @@
   overscroll-behavior: contain;
 
   width: 20rem;
-  height: calc(100vh - var(--nav-bar-height));
+  height: calc(100vh - 2 * var(--nav-bar-height));
```
mattwr18, line 12: this fixes a bug where the footer is not displayed on the contributor's show page because it only accounted for the height of the header, not the height of the footer.

PR #1477, opened 2022-09-29
Give ability to add additional info to data protection

### LINKED ISSUE #1476: Give ability to add additional info to data protection consent
```
As the maintainer of the community
I would like to be able to add additional information to the data protection consent
So that I am sure the contributor understands what they are consenting to without the need to follow the link
```

I am trying to infer what the client, ZVW, asked for. It could also be more generic, like

```
So that I can have control over the language in the data consent field
```

### LINKED ISSUE #1480: Hide parts of Settings from non-admin
```
As tactile.news
I would like to be able to configure parts of 100eyes instances, but not expose them to all users
So that we have more control over sensitive parts being changed that might affect us, ie data protection
```

mattwr18:
Closes #1476 
Closes #1480 

This PR restricts certain Settings field to admin users and introduces the ability to add more information to the onboarding data consent checkbox.

It also adds a system spec that checks that the permissions work properly.

### DIFF
--- a/app/components/markdown/markdown.html.erb
+++ b/app/components/markdown/markdown.html.erb
@@ -1,3 +1,3 @@
-<div <%= attrs %>>
+<div <%= attrs.merge(class: 'rich-content') %>>
   <%= rendered %>
 </div>
--- a/app/components/markdown/markdown.html.erb
+++ b/app/components/markdown/markdown.html.erb
@@ -1,3 +1,3 @@
-<div class="<%= class_attr %> rich-content">
+<div <%= attrs %>>
   <%= rendered %>
 </div>
--- a/app/components/onboarding_consent/onboarding_consent.html.erb
+++ b/app/components/onboarding_consent/onboarding_consent.html.erb
@@ -3,7 +3,8 @@
     object: contributor,
     attr: :data_processing_consent,
     style: :leftAligned,
-    help: data_processing_consent_help.html_safe
+    help: data_processing_consent_help,
+    help_md: Setting.onboarding_data_processing_consent_additional_info,
   ) do |field| %>
     <%= c 'checkbox', **field.input_defaults, required: true %>
   <% end %>
--- a/app/components/onboarding_consent/onboarding_consent.html.erb
+++ b/app/components/onboarding_consent/onboarding_consent.html.erb
@@ -3,8 +3,7 @@
     object: contributor,
     attr: :data_processing_consent,
     style: :leftAligned,
-    help: data_processing_consent_help.html_safe,
-    help_md: Setting.onboarding_data_processing_consent_additional_info,
+    help: data_processing_consent_help,
   ) do |field| %>
     <%= c 'checkbox', **field.input_defaults, required: true %>
   <% end %>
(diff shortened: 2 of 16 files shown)


app/views/onboarding/index.html.erb
```diff
@@ -18,7 +18,7 @@
         <% end %>
 
         <div class="text-serif">
-          <%= c 'markdown', raw: Setting.onboarding_page %>
+          <%= c 'markdown', raw: Setting.onboarding_page, class: 'rich-content' %>
```
mattwr18, line 21: @tillprochaska you probably know a better way to do this, but when I updated the `markdown` component to use the `attrs` instead of `class_attrs`, I could not set a default class as we had been doing. Since it is used only in 1, now 2, places, I just added the class in the call to render the component.

app/components/markdown/markdown.html.erb
```diff
@@ -1,3 +1,3 @@
-<div class="<%= class_attr %> rich-content">
+<div <%= attrs %>>
```
tillprochaska, line 1: This applies a class and still allows you to add more classes when you render the component.

```suggestion
<div <%= attrs.merge(class: 'rich-content') %>>
```

Take a look at the specs here to see what other methods are available: https://github.com/tactilenews/100eyes/blob/main/spec/components/attribute_bag_spec.rb

app/views/onboarding/index.html.erb
tillprochaska: Added a comment/suggestion above :)

app/components/onboarding_consent/onboarding_consent.html.erb
```diff
@@ -3,7 +3,8 @@
     object: contributor,
     attr: :data_processing_consent,
     style: :leftAligned,
-    help: data_processing_consent_help.html_safe
+    help: data_processing_consent_help,
```
tillprochaska, line 6: Have you tried something like this?

```ruby
help = [
  data_processing_consent_help,
  Setting.onboarding_data_processing_consent_additional_info,
].join("\n\n").html_safe
```

Then you could probably pass it directly to the field component, without having to adjust them like this:

```ruby
c('field',
  ...
  help: simple_format(help),
  ...
)
```

app/components/markdown/markdown.html.erb
mattwr18: Cool, thanks :green_heart:

app/components/onboarding_consent/onboarding_consent.html.erb
mattwr18: Yeah, I started off by doing that, but then I thought that the client might want to use multiple paragraphs, but I had forgotten about `simple_format`, so I think it should work. I'll give it a go. Thanks for the idea

app/components/markdown/markdown.html.erb
```diff
@@ -1,3 +1,3 @@
-<div class="<%= class_attr %> rich-content">
+<div <%= attrs.merge(class: 'rich-content') %>>
```
mattwr18, line 1: didn't end up using this component, in the end, but we added the AttributeBag quite a while ago and are refactoring as we touch the code. I originally used this, but later favored a suggestion by @tillprochaska and I don't think it makes sense to remove the change we want to introduce anyways.

PR #1479, opened 2022-09-29
Sanitize tag names in tags input

tillprochaska:
Previously, if a user created a tag like…

```
<script>alert('Hello World')</script>
<img src="https://example.org/ping" />
```

… that would actually get rendered. Thankfully, we have a very restrictive CSP in place that prevents execution of inline and third-party JS as well as loading third-party images. So while the markup would get rendered, no scripts and no images would actually be executed or loaded. Furthermore, only authenticated editors are able to assign and create tags.

### DIFF
--- a/app/components/tags_input/tags_input.js
+++ b/app/components/tags_input/tags_input.js
@@ -7,11 +7,17 @@ const tagColor = tagData => {
 };
 
 function tagTemplate(tagData) {
-  const color = tagColor(tagData);
-
   // This is a simplified copy of the default template that ships with
   // Tagify. The main difference is that tag names are sanitized before
   // being rendered.
+
+  const color = tagColor(tagData);
+
+  // `tagData.name` is an unsanitized tag name from our database.
+  // `tagData.value` is provided by Tagify and has already been sanitized.
+  // https://github.com/yairEO/tagify/blob/8a9efbd680dc05f3f398bf91025f949a69551fbf/src/tagify.js#L1443
+  const label = sanitize(tagData.name) || tagData.value;
+
   return `
     <tag
       tabIndex="-1"
@@ -25,9 +31,7 @@ function tagTemplate(tagData) {
         aria-label="remove tag"
       ></x>
       <div>
-        <span class="tagify__tag-text">${
-          sanitize(tagData.name) || sanitize(tagData.value)
-        }</span>
+        <span class="tagify__tag-text">${label}</span>
       </div>
     </tag>
   `;
--- a/app/components/tags_input/tags_input.js
+++ b/app/components/tags_input/tags_input.js
@@ -1,9 +1,37 @@
 import { Controller } from '@hotwired/stimulus';
 import Tagify from '@yaireo/tagify';
+import sanitize from '../../assets/javascript/helpers/sanitize.js';
 
 const tagColor = tagData => {
   return tagData.color ? tagData.color : 'var(--color-text)';
 };
+
+function tagTemplate(tagData) {
+  const color = tagColor(tagData);
+
+  // This is a simplified copy of the default template that ships with
+  // Tagify. The main difference is that tag names are sanitized before
+  // being rendered.
+  return `
+    <tag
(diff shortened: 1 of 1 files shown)


mattwr18 (commented):
Hey @tillprochaska thanks for looking into this and putting in a PR. I am a little confused though. Maybe you could help clarify some doubts? 

I can see that you are using the `sanitize` function that was already added some 2 years ago to sanitize the tags, but it is not doing what I expected it to do. I added a tag to my contributor and it still renders it exactly as if it had not been sanitized. See screen shots. 

After I add the tag, but before saving, it appears to be sanitized, but after saving it does not. if I console.log the output of `sanitize(tagData.name)`, it looks good, but what is displayed, what I can see in the dev console, and what is saved in the db are all unsanitized, it appears.

Screenshot from 2022-10-04 11-31-54
Screenshot from 2022-10-04 11-45-15
Screenshot from 2022-10-04 11-29-42

mattwr18:
Oh, and another thought I had, could we add some automated test for this?

tillprochaska:
Hey @mattwr18, thanks for paying attention! What happened here is that the tag name in its unsaved state (gray tag pill) has actually been escaped twice. I’ve pushed a fix.

Now it does sanitize all instances of the tag name correctly:

https://user-images.githubusercontent.com/1512805/201545689-92c461dd-76bc-4153-bb37-062263adfb06.mp4

**Unsaved tag in input:**
image

**Saved tag in input:**
image

**Tag in autocomplete dropdown:**
image

> After I add the tag, but before saving, it appears to be sanitized, but after saving it does not. if I console.log the output of sanitize(tagData.name), it looks good, but what is displayed, what I can see in the dev console, and what is saved in the db are all unsanitized, it appears.

I think displaying the tag name just as the user typed it actually is the intended behavior here. The relevant part here is that it must not be treated as raw HTML.

Previously, a tag `<script>...</script>` would be inserted as literal HTML `<script>...</script>`. Now special chars like `<`,`>` etc. will be encoded properly and the HTML that is inserted will be something like `&lt;script&gt;...&lt;/script&gt;`.

When the browser renders this, it will display the text "<script>...</script>", but it will not actually create a `script` node in the document.

tillprochaska:
> Oh, and another thought I had, could we add some automated test for this?

Unfortunately, I don’t think this is possible as the issue and fix exist in the client-side part of 100eyes.

mattwr18 (approved):
Hey @tillprochaska ... sorry for the super delayed look at this, I know you said it wasn't a super high priority, but still.

Ok, thanks for having another look and updating the code. Also, thanks for the further explanation.

I had another look and it is still different for me than it is in your screenshots. I tested on Firefox and Chrome, I can provide versions , if you want. I now see that the tag is as the user writes it and as you say that is the expected behavior, but if I inspect and look at the `<span>` where we add the sanitized label, it is not escaped.

I can see that we save it unsanitized in the db, which is what you explain in your comment from line 16 of the tags_input.js.

I logged the value of the constant `label` that we pass into the return from the function `tagTemplate` and it is sanitized there so not 100% why it appears in my inspector as unsanitized, but as a sanity check I changed our content security policy to allow `unsafe_inline` and it does not execute the script, so even if we were to change that for some reason, I guess it is being sanitized properly.

See screenshot:
Screenshot from 2023-01-30 11-01-47

tillprochaska:
> so not 100% why it appears in my inspector as unsanitized, but as a sanity check I changed our content security policy to allow unsafe_inline and it does not execute the script, so even if we were to change that for some reason, I guess it is being sanitized properly.

I think the difference between actual HTML and properly encoded HTML markup can be a bit difficult to spot in the Firefox dev tools. But if you look closely, you can see that the actual HTML markup is highlighted and uses a brighter color, while the encoded HTML (that is interpreted as plain text) uses a muted text color.

This becomes clearer when you right-click on the text and select "Edit as HTML". This should show something like `&lt;script&gt;`.

215451143-297a16ef-f091-4f22-bb54-09146c1a582d

PR #1481, opened 2022-09-30
Set timeout to avoid long stalled test runs

mattwr18:
I noticed that we had some test runs that went for `6h`, we should avoid this by setting a timeout so that we don't use all our minutes on stalled runs.

Actually, it seems that public repos don't have a limit of free minutes. Maybe, it is still a good idea to time out after some time though? We do have a limit on concurrent jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
 
     env:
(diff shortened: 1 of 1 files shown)


phoeinx (approved):
Hi, 
I also noticed this - thank you for putting in a PR! When re-run those PR test runs usually succeded and I hope that this is some current issue with Github Actions perhaps and not some transient bug. Let's keep an eye on it.
10 minutes as a timeout work well though, especially also resource-wise to not waste resources needlessly.

PR #1482, opened 2022-09-30
Test running tests in docker container

mattwr18:
- we've been getting some intermittent failing tests running them locally and it's more accurate to run them in the container since that is what we actually deploy


mattwr18:
closing this as there is a level of complexity that I hadn't anticipated before opening this

PR #1488, opened 2022-10-04
Add CommunityMetrics section/component/tests

mattwr18:


phoeinx:
Hi, as a first point because I just saw it on staging:
image
The received messages box appears highlighted because it's bigger than the other boxes. I think it would be nicer to have the boxes in the same size.

mattwr18:
> Hi, as a first point because I just saw it on staging: image The received messages box appears highlighted because it's bigger than the other boxes. I think it would be nicer to have the boxes in the same size.

I think it's bigger because the text goes to 2 lines so maybe we could shorten the German?

mattwr18:
What screen width were you using @phoeinx? The only way I was able to reproduce this was to use a screen width of less than 820px

We could change it to `Nachrichten insgesamt`, which is what we use for the request show page, but that only gives us till 750px width before it also creates the same UI issue.

I had a doubt about this anyways, cause I thought it might be better to keep the same text in both places, but I don't know German enough to know whether this is good for both. It translates in Deepl to total messages so it could be.

mattwr18:
I ended up updating the `CardMetrics` to add it's own styles, which means that we fix the same issue that occurs on the request show page for screen widths less than `670px`. It uses a grid layout and gives a minimum width to each card and when it needs wraps to a second/third row. This is a step in the direction of responsiveness.

Considering the feedback from our clients, we want to pay better attention to how new pages we build look on smaller devices.

PR #1491, opened 2022-10-04
Configure dependabot to ignore patch version updates.

phoeinx:
Currently, the maintenance overhead for merging dependabot updates is quite high, especially considering the few active maintainers. This PR proposes reducing maintenance overhead for merging in multiple patch updates per minor version, security updates are not affected by this change.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,18 +4,30 @@ updates:
   directory: "/"
   schedule:
     interval: monthly
+  ignore:
+    - dependency-name: "*"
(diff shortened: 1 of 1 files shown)


mattwr18 (approved):
Great! thanks for taking care of this. For our future selves, I had a look at the [dependabot docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore) to have a look at the options.

PR #1500, opened 2022-10-17
Fix identation error in dependabot.yml.

phoeinx:


mattwr18 (approved):

mattwr18:
Great! not sure why it didn't get caught before or how you noticed it now, but glad to get it fixed.

PR #1504, opened 2022-10-17
Update staging config for new Threema ID

### LINKED ISSUE #1503: Receiving Threema messages on staging is broken
## Steps to reproduce

- Go to staging.100ey.es
- Onboard as a Threema contributor
- Send out a request to this contributor
- reply from your Threema account
- Notice no reply is received

## Issue
At some point, we decided to use `*100EYES` as our id for demo and at some point created `*100STAG`, but did not update the staging config and did not have the URL for incoming messages correctly configured.

mattwr18:
Closes #1503


mattwr18:
Tested this by sending and receiving messages from staging.100ey.es through Threema.

PR #1507, opened 2022-10-18
Validate threema_id on save when present

### LINKED ISSUE #723: Validate Threema ID during onboarding/when saving the Threema ID
We could do a public key lookup every time a contributor onboards using Threema (or even every time a contributor record is updated) to ensure that that Threema ID is actually linked to an active, valid Threema account.

Checking on every change (e.g. also when an editor updates the Threema ID manually) has the added benefit that our Threema IDs are always guaranteed to be valid (at least at the time they are saved), however the additional delay (due to an additional HTTP request to the Threema API) every time a user record is updated might be annoying.

This might become slightly less important as soon as we have a UI to display delivery issues or we implement automated welcome or double opt-in messages. Although these solutions would be suboptimal in terms of UX compared to catching invalid IDs early in the process.

mattwr18:
Closes #723

### DIFF
--- a/app/components/base_field/base_field.html.erb
+++ b/app/components/base_field/base_field.html.erb
@@ -11,15 +11,15 @@
     <% if help %>
       <p class="BaseField-helpText"><%= help %></p>
     <% end %>
-
-    <% if errors.any? %>
-      <% errors.each do |error| %>
-        <strong class="BaseField-errorText">
-          <%= error %>
-        </strong>
-      <% end %>
-    <% end %>
   </div>
 
   <%= content %>
+
+  <% if errors.any? %>
+    <% errors.each do |error| %>
+      <strong class="BaseField-errorText">
+        <%= error %>
+      </strong>
+    <% end %>
+  <% end %>
 </div>
--- a/app/models/contributor.rb
+++ b/app/models/contributor.rb
@@ -3,6 +3,7 @@
 # rubocop:disable Metrics/ClassLength
 class Contributor < ApplicationRecord
   include PgSearch::Model
+  include ActiveModel::Validations
 
   attr_accessor :editor_guarantees_data_consent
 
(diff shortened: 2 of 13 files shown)


app/components/base_field/base_field.html.erb
```diff
@@ -11,15 +11,15 @@
...
 
   <%= content %>
+
+  <% if errors.any? %>
+    <% errors.each do |error| %>
+      <strong class="BaseField-errorText">
+        <%= error %>
+      </strong>
+    <% end %>
+  <% end %>
```
mattwr18, line 24: I made this change because the `hide_label` attribute was hiding any errors that occur and I don't think that was its intention. At least semantically that is not what it says it does. Also, I am more familiar with the errors on the bottom of the input vs. the top

app/models/contributor.rb
```diff
@@ -32,12 +33,14 @@ class Contributor < ApplicationRecord
   validates :data_processing_consent, acceptance: true, unless: proc { |c| c.editor_guarantees_data_consent }
 
   validates :email, uniqueness: { case_sensitive: false }, allow_nil: true, 'valid_email_2/email': true
-  validates :threema_id, uniqueness: { case_sensitive: false }, allow_blank: true, format: { with: /\A[A-Za-z0-9]+\z/ }, length: { is: 8 }
+  validates :threema_id, uniqueness: { case_sensitive: false }, allow_blank: true
```
mattwr18, line 36: we still want to validate that we don't have any contributors already with the threema_id, ignoring case

config/locales/de.yml
```diff
@@ -409,8 +409,9 @@ de:
         label: Handynummer
       threema_id:
         label: Threema ID
-        help: 6 Zeichen lang, besteht nur aus Zahlen und Großbuchstaben
+        help: 8 Zeichen lang, besteht nur aus Zahlen und Großbuchstaben
```
mattwr18, line 412: somehow this help was with the incorrect length for a valid threema_id

PR #1525, opened 2022-11-08
Avoid reliance on project name for legacy messages

mattwr18:
- At this moment, the majority of all instances outbound messages, we don't know who sent. I originally was relying on the project name to display something, but not all instances have project names, or some might be super long, so we just display the date


mattwr18:
Closing this as there are currently no longer any instances that do not have a project_name set and there has been no traction in changing the UI

PR #1537, opened 2022-11-21
Preserve a file's caption if received

### LINKED ISSUE #1527: [BUG] Inbound Threema images with caption text is lost
If a contributor sends us a message with text as a caption, we do not save the image's caption as the text of the message.
I haven't seen if it is part of the raw_data, but I guess this is an implementation detail in the threema-rb gem we use. Further investigation is needed.

mattwr18:
Fixes #1527

### DIFF
--- a/app/adapters/threema_adapter/inbound.rb
+++ b/app/adapters/threema_adapter/inbound.rb
@@ -31,9 +31,11 @@ module ThreemaAdapter
     end
 
     def initialize_text(decrypted_message)
-      return nil unless decrypted_message.instance_of? Threema::Receive::Text
-
-      decrypted_message.content
+      if decrypted_message.instance_of? Threema::Receive::Text
+        decrypted_message.content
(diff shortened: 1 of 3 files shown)


app/adapters/threema_adapter/inbound.rb
```diff
@@ -31,9 +31,11 @@ def delivery_receipt?(decrypted_message)
...
 
     def initialize_text(decrypted_message)
-      return nil unless decrypted_message.instance_of? Threema::Receive::Text
-
-      decrypted_message.content
+      if decrypted_message.instance_of? Threema::Receive::Text
+        decrypted_message.content
+      elsif decrypted_message.instance_of?(Threema::Receive::File) && decrypted_message.caption
+        decrypted_message.caption
+      end
```
phoeinx, line 38: perhaps add an explicit return value of nil if none of the cases match
```suggestion
      end
      return nil
```

phoeinx:
Looks good, but maybe some tests testing preserving the caption for a file message?

PR #1542, opened 2022-11-24
Upgrade signal-cli-rest-api, which includes bug fix

### LINKED ISSUE #1535: Update signal-cli-rest-api to latest v0.64
This version includes a more permanent fix to the Signal issue discussed in [Slack thread](https://tactilenews.slack.com/archives/C01ACFHFTV2/p1665037975807479)

mattwr18:
- v0.63 included a temp bug fix for https://github.com/bbernhard/signal-cli-rest-api/issues/286. v0.64 is a more permanent fix, but also upgrades signal-cli to by a major version so we should test properly before merging and deploying

Closes #1535

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,7 +26,7 @@ services:
       - postgres_data:/var/lib/postgresql/data
 
   signal:
-    image: bbernhard/signal-cli-rest-api:0.63
+    image: bbernhard/signal-cli-rest-api:0.64
     environment:
       - USE_NATIVE=0
         #- AUTO_RECEIVE_SCHEDULE=0 22 * * * #enable this parameter on demand (see description below)


mattwr18:
v0.64 also upgrades the `signal-cli` from `v0.10.9` to `v0.11.4`, so I looked at the releases from `signal-cli` and there is one breaking change from `v0.10.9` to `v0.10.10`:

```
Attention: Now requires native libsignal-client version 0.18.1
```

I checked and `signal-cli-rest-api`, put in a [commit to upgrade](https://github.com/bbernhard/signal-cli-rest-api/commit/bddc30fe5b16841077e95a01f5f3861dceb1f08b) the `libsignal-client` to `v0.20.0` as part of the pre-release.

I also built and deployed the image to staging.100ey.es and checked sending/receiving through Signal. 

I will take a snapshot of the production instances and a backup before deploying this tonight, just in case, but I think we are good to upgrade this.

PR #1555, opened 2022-12-09
Avoid creating Photo record for outbound attachments

### LINKED ISSUE #1554: BUG PosmarkAdapter::Inbound does not differentiate between attachments sent and received
This results in assuming the attachment was sent to us and creating a Photo object and associating it with the inbound Message.

mattwr18:
Fixes #1554


mattwr18:
closing this as it never got any traction

PR #1569, opened 2022-12-20
Add WhatsApp channel through Twilio API

### LINKED ISSUE #1568: Add WhatApp as a channel through Twilio API
```
As an organization
We would like to add WhatsApp as a channel
because many contributors already use it
```

mattwr18:
Closes #1568

### TODOs

- [x] Add unsubscribe feature
- [x] Separate permitted error params
- [x] Manually test error handling
- [ ] Handle incoming unsuccessfully delivered messages. We don't receive status reports for these, we can fetch recent messages and select undelivered/failed ones and raise an error if there are any.


drjakob:
A contributor onboarded with signal and (with new account/same number) whatsapp gets only messages on signal.
 small typo in Dashboard "Whats_app" should be "WhatsApp"
 missing WA number in contributors profile
 OLD "Interner Titel. Damit du zwischen all deinen Fragen den Überblick behältst. Der Titel wird nicht an die Community gesandt." NEW should be: "Titel Damit du zwischen all deinen Fragen den Überblick behältst. Der Titel wird auf manchen Kanälen zur Information an Community-Mitglieder mitgeschickt."
– icon & instruction on onboarding page
– [ ] send & receive images
 Limit for message length
 correct routing bug "2/1 haben geantwortet" (see image). Jakobs answer should NOT be routed to this question because he hasn't received this question at all and answered to another question
[ ]

mattwr18:
> A contributor onboarded with signal and (with new account/same number) whatsapp gets only messages on signal.

Can you give me more information? I don't think this should have anything to do with each other. I think you said that @FrauCsu was a Signal contributor and then onboarded as a WhatsApp contributor, correct? She then did not receive the welcome message? Did she also not receive any requests? Contributors have different attributes for signal/whats_app so there should be no reason why you couldn't use the same number. Although, this is more for our team than for contributors, or? Do we have some contributors outside the team that are onboarded to multiple channels?

I checked later and @FrauCsu did not have a contributor with either Signal or WhatsApp so presumably they were both deleted, then later someone used a number to onboard her again. If I assume this is the same number that was used eariler, I can see in the Twilio logs that there was a welcome message sent out to that number, which was reported by Twilio as delivered, at 12:34am CET. I assume this is the second attempt at onboarding? I can see that there were several outgoing/incoming messages from that number around that time so I think everything is ok.

There is also an earlier attempt with a number that at first glance seems to be identical, where the first welcome message was sent out at 8:45am and several other requests were sent out to and all were reported as delivered by Twilio. Looking closer, I can see there is one digit that is a 2 in the number that is presumably @FrauCsu, but a 3 in the first attempt. It seems to me that we have probably made a slight typo when onboarding and have been spamming some other whatsapp user. It doesn't look like @FrauCsu tried to send a message to our staging instance; otherwise, I expect that we would have sent/or attempted to send the unknown contributor message.

mattwr18:
>  icon & instruction on onboarding page

I need an icon from the design team. I noticed that it's missing, but I don't have the icon to put there. What do you mean by instruction?

mattwr18:
> send & receive images

Yes, I wanted to give it to you to manually test, but I stated several times that I haven't even started looking into sending/receiving other types of messages. I think we need to do that before we deploy the feature, but I'm not sure it's even the next step. I think before working on other types, we should probably work on verifying that the messages we receive on our webhook endpoint actually came from Twilio. See [Webhooks Security](https://www.twilio.com/docs/usage/webhooks/webhooks-security)

We also need to see what types of messages we want to support to receive. Sending we only have the feature to send images to that is straightforward, but for receiving I think it maybe varies by channel. As that was in place a while before sending images, I think that is where I would start.

mattwr18:
> Limit for message length

Have we decided that we will do this? This is something that would affect all instances and all channels, even if the instance does not opt to use Twilio. There are also some complicating factors, I think. For example, do we want to validate the length of the request, which would affect outgoing messages? Or the message, which would affect incoming and outgoing messages? If the latter, what should we do if we receive a message that fails this validation? Save the raw data and inform the client, or admin only?

Maybe we can validate the request vs the message so as not to affect incoming messages and set up an email alert when an error occurs for incoming messages that exceed the limit. See [Webhook & Email Triggers](https://console.twilio.com/frame?frameUrl=%2Fconsole%2Fdebugger%2Falert-triggers%3Fx-target-region%3Dus1)

mattwr18:
> correct routing bug "2/1 haben geantwortet" (see image). Jakobs answer should NOT be routed to this question because he hasn't received this question at all and answered to another question

I answered this in Slack, but here is what I found:

> Ok, I think I have an idea of what happened with the message Martin sent [staging.100ey.es](http://staging.100ey.es/) and it was assigned to the incorrect request.

> At 9:35am there was a scheduled request that went out to Jakob-Whatsapp and there is only one contributor with that tag, not surprisingly,  @drjakob 

> At 10:57am, [@Martin T.](https://tactilenews.slack.com/team/U03G4MQKQV7) sent a message to [staging.100ey.es](http://staging.100ey.es/) even though he had not received any message from them. The logic, which is not new to WhatsApp, looks for the latest received_message the contributor has received and if there is one it assigns the message to that request. If there is not one, it looks to see if there have ever been any requests from that instance and assigns it to that request.

> At 11:00 a request to [@Martin T.](https://tactilenews.slack.com/team/U03G4MQKQV7)
 was sent out. After that his responses have been correctly assigned.

> This is an old issue that is outside the scope of WhatsApp , unless there is more information than what I've been able to figure out or that I've been given with the BUG report. This is incidentally, why we have the move message feature so that the clients can correctly assign the message to the request it belongs to.

> One could argue that we should not randomly assign messages to requests if the contributor has not received any requests from us, but that we would need to have its own issue for and discuss further.

app/controllers/whats_app/webhook_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      adapter.consume(whats_app_webhook_params) { |message| message.contributor.reply(adapter) }
+    end
+
+    def errors
+      return unless webhook_params['Level'] == 'ERROR' && JSON.parse(params['Payload'])['error_code'].to_i == 21_617
+
+      payload = JSON.parse(webhook_params['Payload'])
+      more_info = payload['more_info']
+      parameters = payload['webhook']['request']['parameters']
+      error_text = more_info['Msg'].gsub(Setting.twilio_account_sid, '<ACCOUNT_SID>')
```
roschaefer, line 36: Since you seem to have the `Setting.twilio_account_sid` in the payload, you could implement a reverse proxy that forwards the errors (http requests, whatever) to the right 100eyes instance. This is a workaround for the fact that we have only one error URL at Twilio.

config/routes.rb
```diff
@@ -41,6 +44,12 @@
     post '/webhook', to: 'webhook#message'
   end
 
+  namespace :whats_app do
+    post '/webhook', to: 'webhook#message'
+    post '/errors', to: 'webhook#errors'
+    post '/status', to: 'webhook#status'
```
roschaefer, line 50: After talking to @mattwr18 I understand that Twilio sends read state etc. on this webhook.

app/controllers/whats_app/webhook_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+                    :SmsStatus, :Timestamp, :To, :WaId, *media_params)
+    end
+
+    def status_params
+      params.permit(:AccountSid, :ApiVersion, :ChannelInstallSid, :ChannelPrefix, :ChannelToAddress, :ErrorCode, :EventType,
+                    :From, :MessageSid, :MessageStatus, :SmsSid, :SmsStatus, :To)
+    end
+
+    def media_params
+      indeces_of_media = params[:NumMedia].to_i
```
roschaefer, line 64: I think plural of index is "indices".

app/controllers/whats_app/webhook_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+                                                                        whats_app_phone_number: status_params['To'].split('whatsapp:').last)
+      ErrorNotifier.report(exception, context: { message_sid: status_params['MessageSid'] })
+    end
+
+    private
+
+    def webhook_params
+      params.permit(:AccountSid, :ApiVersion, :Body, :From, :Level, :Latitude, :Longitude, :MessageSid, :NumMedia, :NumSegments,
+                    :ParentAccountSid, :Payload, :PayloadType, :ProfileName, :ReferralNumMedia, :Sid, :SmsMessageSid, :SmsSid,
+                    :SmsStatus, :Timestamp, :To, :WaId, *media_params)
```
roschaefer, line 55: Why is `media_params` in the list of strong params? Isn't that a method?

app/controllers/whats_app/webhook_controller.rb
roschaefer: For what is `media_params` being used? I don't see it.

config/locales/de.yml
```diff
@@ -342,6 +357,22 @@ de:
...
+        new_request_evening_2: "Guten Abend %{first_name}, zum Thema %{request_title} haben wir eine Frage. Antworten Sie uns darauf?"
+        new_request_evening_3: "Hallo %{first_name}, heute Abend möchte ich gern mehr zum Thema %{request_title} erfahren. Würden Sie meine Frage beantworten?"
+        new_request_night_1: "Hallo %{first_name}, heute Nacht ist mir etwas zum Thema %{request_title} eingefallen. Antworten Sie mir?"
+        new_request_night_2: "Hallo %{first_name}, ich weiß, es ist spät. Aber ich möchte gern mehr zum Thema %{request_title} erfahren. Möchten Sie auf meine Frage antworten?"
+        new_request_night_3: "Hallo %{first_name}, zum Thema %{request_title} möchte ich heute Nacht gern mehr erfahren. Beantworten Sie mir eine kurze Frage?"
+      unknown_content_template: "Tut mir leid %{first_name}, leider ist unsere Software kann derzeit Textnachrichten, Sprachnachrichten, Bilder und Reaktionen auf Nachrichten verarbeiten, jedoch noch keine weiteren Inhalte wie z.B. Sticker oder Kontakte. In der Zwischenzeit senden Sie bitte Nachrichten mit solchen Inhalten an %{contact_person}. Herzlichen Dank!"
```
roschaefer, line 375:
```suggestion
      unknown_content_template: "Tut mir leid %{first_name}, leider kann unsere Software derzeit nur Textnachrichten, Sprachnachrichten, Bilder und Reaktionen auf Nachrichten verarbeiten, jedoch noch keine weiteren Inhalte wie z.B. Sticker oder Kontakte. In der Zwischenzeit senden Sie bitte Nachrichten mit solchen Inhalten an %{contact_person}. Herzlichen Dank!"
```

config/locales/de.yml
```diff
@@ -342,6 +357,22 @@ de:
...
+        new_request_day_2: "Hallo %{first_name}, ich möchte gern mehr zum Thema %{request_title} erfahren. Möchten Sie auf die heutige Frage antworten?"
+        new_request_day_3: "Guten Tag %{first_name}, ich habe eine neue Frage zu %{request_title}. Möchten Sie antworten?"
+        new_request_evening_1: "Guten Abend %{first_name}, wir haben eine neue Frage zum Thema %{request_title}. Möchten Sie darauf antworten?"
+        new_request_evening_2: "Guten Abend %{first_name}, zum Thema %{request_title} haben wir eine Frage. Antworten Sie uns darauf?"
+        new_request_evening_3: "Hallo %{first_name}, heute Abend möchte ich gern mehr zum Thema %{request_title} erfahren. Würden Sie meine Frage beantworten?"
+        new_request_night_1: "Hallo %{first_name}, heute Nacht ist mir etwas zum Thema %{request_title} eingefallen. Antworten Sie mir?"
+        new_request_night_2: "Hallo %{first_name}, ich weiß, es ist spät. Aber ich möchte gern mehr zum Thema %{request_title} erfahren. Möchten Sie auf meine Frage antworten?"
+        new_request_night_3: "Hallo %{first_name}, zum Thema %{request_title} möchte ich heute Nacht gern mehr erfahren. Beantworten Sie mir eine kurze Frage?"
```
roschaefer, line 374:
```suggestion
        new_request_night_3: "Hallo %{first_name}, zum Thema %{request_title} möchte ich heute nacht gern mehr erfahren. Beantwortest du mir eine kurze Frage?"
```
All of these use formal German. Since we address our contributors by `first_name` I guess we should use informal German. This would need to be changed for all the above :point_up:

db/seeds.rb
```diff
@@ -6,9 +6,7 @@
 
 organization = Organization.create_or_find_by(name: '100eyes',
                                               upgrade_discount: 10,
-                                              business_plan: BusinessPlan.where.not(
-                                                valid_from: nil, valid_until: nil
-                                              ).first)
+                                              business_plan: BusinessPlan.first)
```
roschaefer, line 9: Is this related?

roschaefer (changes requested):
brain explodes

I think this is it for now. @mattwr18 and I agreed on doing proper testing in another pair-programming session together.

app/adapters/whats_app_adapter/outbound.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    def self.send_files(files, recipient, text)
+      files.each_with_index do |file, index|
+        WhatsAppAdapter::Outbound::File.perform_later(recipient: recipient, text: index.zero? ? text : '', file: file)
+      end
+    end
+
+    def self.send_text(recipient, text)
+      WhatsAppAdapter::Outbound::Text.perform_later(recipient: recipient, text: text)
+    end
```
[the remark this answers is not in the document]
mattwr18: @drjakob and I discussed this and decided it's not worth the time to change the implementation for all 5 channels and change the media preview in the frontend for this change.

app/controllers/whats_app/webhook_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      adapter.on(WhatsAppAdapter::UNKNOWN_CONTENT) do |contributor|
+        WhatsAppAdapter::Outbound.send_unknown_content_message!(contributor)
+      end
+
+      whats_app_webhook_params = webhook_params.to_h.transform_keys(&:underscore)
+      adapter.consume(whats_app_webhook_params) { |message| message.contributor.reply(adapter) }
+    end
+
+    def errors
+      return unless webhook_params['Level'] == 'ERROR' && JSON.parse(params['Payload'])['error_code'].to_i == 21_617
```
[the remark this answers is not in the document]
mattwr18: fixed in d1437b81

db/seeds.rb
mattwr18: No, it's code creep.
Been removed in 4d4b66e2

app/components/chat_message/chat_message.html.erb
```diff
@@ -85,6 +85,13 @@
       </header>
     <% end %>
 
+    <% if video? %>
+      <video width="320" height="240" controls>
+        <source src="<%= rails_blob_url(video) %>" type="video/mp4">
+        Your browser does not support the video tag.
+      </video>
+    <% end %>
```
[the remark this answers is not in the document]
mattwr18: Done in 4d4b66e2

app/controllers/whats_app/webhook_controller.rb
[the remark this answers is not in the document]
mattwr18: I refactored the error handling to report all errors originating from Twilio in d0e1f1f5

app/controllers/whats_app/webhook_controller.rb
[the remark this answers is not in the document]
mattwr18: removed as we discussed this and the implementation changed.

app/controllers/whats_app/webhook_controller.rb
mattwr18: > I think plural of index is "indices".

yeah, haha I guess as a native speaker, I should know that. I mixed `indexes` and `indices`

> For what is media_params being used? I don't see it.

It is used in the strong params to permit `MediaContentTypeX/MediaUrlX`, where `X` is the index of the file.

[Twilio docs](https://www.twilio.com/docs/whatsapp/tutorial/send-and-receive-media-messages-whatsapp-ruby#get-urls-to-the-media) say:

> Twilio sends form variables named MediaUrlX, where X is a zero-based index.

But, I actually made this more complex than it need be since it also says:

> WhatsApp messages will only contain one media file per incoming message, so you can access the file at MediaUrl0 on the incoming request from Twilio to your webhook URL.

This was also my experience with sending multiple images. I was already considering refactoring it to simplify the logic, but now that I read this closer, I think it's definitely worth it.

app/controllers/whats_app/webhook_controller.rb
mattwr18: Don't know how to set up a reverse proxy. I'd be interested in learning, but I'm not sure it's worth the time spent right now.

I find it a bit strange that we have this code deployed to all instances and are only able to configurte one webhook endpoint for errors in Twilio, but I'm not sure what benefit we'd gain by having this reverse proxy. If for some reason the instance we configure in Twilio happens to be unreachable when an error happens, we'd likely miss the reporting, but depening on the issue the reverse proxy might also be unavailable :shrug: 

Also, would the idea be that only one instance would have the reverse proxy?

I thought of maybe having a microservice, which exposes an endpoint whose job it is to report the errors for all instances, then the code would not live on any instance, but there are also questions with that approach. Anything other than what we have would take more time, so I think what we have is an acceptable solution for now.

config/routes.rb
mattwr18: Yes, see [Track Delivery Status of Messages in Ruby](https://www.twilio.com/docs/sms/tutorials/how-to-confirm-delivery-ruby).

At the moment, we are just using it to track `UNSUCCESSFUL_DELIVERY` and raising an error. I consider it "unsuccessful" if the read status is `undelivered` or `failed`. I saw this when trying to send a freeform message to a contributor who is unable to receive one because the have not sent us a message in more than 24 hours.

It would also happen if there is some difference between the approved templates in our WhatsApp page and what we have in the application. For example, a typo, or someone edits the WhatsApp template, but we don't update the application.

I noticed that we don't get an error if we try to send a freeform message when one is not permitted, the message just never gets delivered. I configured this so that we actually do raised an error and we can track them and fix them.

app/controllers/whats_app/webhook_controller.rb
mattwr18: I can't remember where I saw this trick, but it's a way of permitting multiple strong params returned by the `media_params` method as an array. The idea was that if we received three images, they would come as:

```
MediaUrl0 MediaContentType0
MediaUrl1 MediaContentType1
MediaUrl2 MediaContentType2
```
and we would want to permit all these params. In this scenario, we would also not know how many images we would be receiving per webhook.

Since then, I've re-read the documentation and based on my expercience, we only ever receive one "media" per webhook, so I've simplified the logic in 3424a1f3

config/locales/de.yml
mattwr18: I appreciate this. @FrauCsu had the same concern, but as we need to get these first approved by WhatsApp and have had some issues with this, they are going to stay as they are until we resolve the issues upstream

app/components/onboarding_channel_buttons/onboarding_channel_buttons.rb
```diff
@@ -4,14 +4,19 @@ module OnboardingChannelButtons
...
+      @channels = channels
     end
 
-    def show_signal?
-      Setting.signal_server_phone_number.present?
+    attr_reader :channels
+
+    def styles
+      return super if (channels.length % 3).zero?
+      return super + [:twoColumn] if channels.length.even?
```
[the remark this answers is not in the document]
mattwr18: Ok, I've simplified it in f33783fe
It's not perfect since the styling assumes up to 6 channels being configured. If we end up adding 7 channels, it won't look broken, but we might want to reconsider the styling.

app/components/onboarding_channel_buttons/onboarding_channel_buttons.rb
```diff
@@ -4,14 +4,19 @@ module OnboardingChannelButtons
...
+
+      super + [:threeColumn]
```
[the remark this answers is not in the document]
mattwr18: As discussed it was an unfortunate name for the modifier. It was more like `threeColumnWithTwoOrphans`.
I've simplified it and just added the styles to the base class. The styling is only applied if there are 5/8/11,etc. channels configured.

app/controllers/whats_app/webhook_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      adapter = WhatsAppAdapter::Inbound.new
+
+      adapter.on(WhatsAppAdapter::UNKNOWN_CONTRIBUTOR) do |whats_app_phone_number|
+        exception = WhatsAppAdapter::UnknownContributorError.new(whats_app_phone_number: whats_app_phone_number)
+        ErrorNotifier.report(exception)
+      end
+
+      adapter.on(WhatsAppAdapter::RESPONDING_TO_TEMPLATE_MESSAGE) do |contributor|
+        contributor.update!(whats_app_message_template_responded_at: Time.current, whats_app_template_message_sent_at: nil)
+        message = contributor.received_messages.first
```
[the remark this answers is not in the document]
mattwr18: Yeah, I think that when you added this association 3 years ago, it was just assumed that a created message would have been received by the contributor since we have an `after_create_commit` that calls our adapters `send!(message)` method to attempt to send the message out.

In this case, we don't automatically send out the message though, we first check to see if a freeform message is permitted by Twilio/WhatsApp, then either send a template and wait for a response giving us permission to send the message, or if it is permitted, we send the message out.

This results in some interesting semantics where we have a list of "received messages", which might have never been sent out if the contributor is not recently active and/or gives us permission to send the message out.

We also have the case where the contributor does not receive the message because the job to send it out fails for one reason or another. I know of cases where a contributor has never received a message from us because their channel id was invalid and yet you wouldn't know this by looking at the `contributor.received_messages`.

I would be in favor of changing the semantics of this if it also meant having another attribute that would be set when the message is actually successfully delivered to the contributor so we could distinguish between received messages and outbound messages.

app/controllers/whats_app/webhook_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+        WhatsAppAdapter::Outbound.send!(message)
```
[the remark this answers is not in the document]
mattwr18: I looked into this and there is no information that we receive that lets us know what message they are responding to even though it appears to the contributor like they are quote replying to a specific template.

This is really unfortunate.

app/components/chat_message/chat_message.html.erb
```diff
@@ -86,10 +86,7 @@
     <% end %>
 
     <% if video? %>
-      <video width="320" height="240" controls>
-        <source src="<%= rails_blob_url(video) %>" type="video/mp4">
-        Your browser does not support the video tag.
-      </video>
+      <%= c 'chat_message_video', videos: files %>
```
roschaefer: This could probably go into a separate PR that can be merged earlier.

app/components/settings_form/settings_form.html.erb
```diff
@@ -67,7 +67,9 @@
 
     <% if current_user.admin? %>
       <%= c 'stack' do %>
+        <%= c('heading', tag: :h2) { t('setting.form.groups.all') } %>
         <%= c('settings_field', type: :image_input, attr: :channel_image) %>
+        <%= c('settings_field', type: :textarea, attr: :about) %>
```
roschaefer, line 72: This also looks like unrelated code that could be merged in a separate PR.

Small and many PRs for the win!

app/controllers/whats_app/webhook_controller.rb
```diff
@@ -9,34 +9,42 @@ def message
...
       end
 
-      whats_app_webhook_params = webhook_params.to_h.transform_keys(&:underscore)
-      adapter.consume(whats_app_webhook_params) { |message| message.contributor.reply(adapter) }
+      whats_app_message_params = message_params.to_h.transform_keys(&:underscore)
+      adapter.consume(whats_app_message_params) { |message| message.contributor.reply(adapter) }
     end
 
     def errors
-      return unless webhook_params['Level'] == 'ERROR' && JSON.parse(params['Payload'])['error_code'].to_i == 21_617
```
roschaefer: Nice that magic number is gone.

config/locales/about.txt
```diff
@@ -0,0 +1 @@
+Du nimmst an einem Communityprojekt mit der Redaktion teil. Dazu nutzt die Redaktion die Software 100eyes: Du antwortest über Whatsapp, andere Teilnehmer:innen nutzen andere Kanäle. Alles läuft in einem Dashboard zusammen.   Die Redaktion kann so viele Eins-zu_Eins-Dialoge mit vielen Teilnehmer:innen in der Community gleichzeitig führen.
```
roschaefer, line 1:
```suggestion
Du nimmst an einem Communityprojekt mit der Redaktion teil. Dazu nutzt die Redaktion die Software 100eyes: Du antwortest über Whatsapp, andere Teilnehmer:innen nutzen andere Kanäle. Alles läuft in einem Dashboard zusammen. Die Redaktion kann so viele Eins-zu_Eins-Dialoge mit vielen Teilnehmer:innen in der Community gleichzeitig führen.
```

roschaefer: May_be, some stran:ge inter-Punctuation here.

spec/adapters/whats_app_adapter/outbound_spec.rb
```diff
@@ -106,4 +106,32 @@
...
+        before { create(:message, sender: contributor) }
+
+        it { is_expected.to eq(true) }
+      end
+
+      context 'has not been received' do
+        it { is_expected.to eq(false) }
+      end
+    end
+  end
```
roschaefer, line 136: Great! :+1:

db/seeds.rb
```diff
@@ -6,7 +6,9 @@
 
 organization = Organization.create_or_find_by(name: '100eyes',
                                               upgrade_discount: 10,
-                                              business_plan: BusinessPlan.first)
+                                              business_plan: BusinessPlan.where.not(
```
roschaefer: `BusinessPlan` looks like a model that I don't know of yet, because I review not often enough.

roschaefer (approved):
twilio

I don't see anything terrible why this shouldn't get merged.

True, it's a lot, but let's test it in production and await the incoming problems :slightly_smiling_face: 

Apart from that I think the PR is a little busy. I have adopted the habit of splitting my PRs into as many steps as possible that can be merged independently.

mattwr18:
> Apart from that I think the PR is a little busy. I have adopted the habit of splitting my PRs into as many steps as possible that can be merged independently.

I agree with that and always try to adopt that workflow. Recently, it's been a bit difficult with the lack of PR reviews and getting manual testing done by Jakob and Astrid.

PR #1632, opened 2023-03-16
Add community engagement metric to dashboard

### LINKED ISSUE #778: Community-Metrics
As an editor, I want community metrics at a glance: 

- [ ] Speed of response
- [ ] best-responding community members
- [ ] total number of responses to research
- [ ]  most important tags
- [ ] most successful questions
- [ ] From this, a 100-KPI score should be calculated to have a target value for projects.

mattwr18:
Closes #778


mattwr18:
@tillprochaska I added a commit with one of the attempts I have tried to get the standalone svg icons addesd to our icons.svg file. See [`f985d99` (#1632)](https://github.com/tactilenews/100eyes/pull/1632/commits/f985d9982318cdcf62d6b96391574a85dd86b130)

I navigated to localhost:3000/icons.svg and can see that the icon I added is discoverable.
Screenshot from 2023-03-16 10-15-00

I also check the inspector to see if I could notice anything different there and did not notice anything.
Screenshot from 2023-03-16 10-16-26

The standalone svg is also added in this PR if you want to look at the source.

Thanks for looking into this.

mattwr18:
@drjakob here is a screen shot of the changes I made after your feedback.

* Round to nearest integer and add `%` symbol:
Screenshot from 2023-06-13 11-12-48

* Basic responsiveness - reverts to the current design for metrics since the new design does not leave room for small screens:

Screenshot from 2023-06-13 11-31-02


Don't like the `%` being in the text in the case of smaller screens, but to have that screen width dependent, we would need to add js and would add a level of complication that I don't think would be worth it. If we continue our responsiveness refactorings, we could explore it in the future. For now, just interested in the layout not being completely broken.

drjakob:
Yes. For the mobile versions the icons left and then the text?

mattwr18:
> Yes. For the mobile versions the icons left and then the text?

We could maybe do that, it would mean that we have three different designs for the time being though. Remember that this design is borrowed from the request metrics.

image

PR #1639, opened 2023-04-18
Add request title/text to search resutls

### LINKED ISSUE #1583: Search including questions
As editor I want the search including my questions not only answers.

mattwr18:
Closes #1583 

### DEV Notes:

To extend the search results to include the new class/attributes, we must follow the instructions in [Rebuilding search documents for a given class](https://github.com/Casecommons/pg_search#rebuilding-search-documents-for-a-given-class)(emphasis mine):

> If you change the :against option on a class, **add multisearchable to a class that already has records in the database**, or remove multisearchable from a class in order to remove it from the index, you will find that the pg_search_documents table could become out-of-sync with the actual records in your other tables.


app/components/search_result/search_result.html.erb
```diff
@@ -3,6 +3,8 @@
     <%= c 'contributor_row', contributor: result %>
   <% when 'Message' %>
     <%= c 'excerpt', title: result.sender_name, text: result.text, link: result.chat_message_link, date: result.created_at %>
+  <% when 'Request' %>
+    <%= c 'excerpt', title: result.title, text: result.text, link: request_path(result), date: result.created_at %>
```
mattwr18, line 7:
```suggestion
    <%= c 'excerpt', title: result.title, text: result.text, link: request_path(result), date: result.broadcasted_at || result.created_at %>
```

Should we have some UI that distinguished planned/scheduled requests from broadcasted ones?

mattwr18:
@drjakob and I did some manual testing.

* Add avatars to search results

mattwr18:
@drjakob please have a look at the search results with avatar for the sender of the request/message:

image

it also takes into account a time where we did not keep track of the sender of a message/request and does not break.

image

PR #1644, opened 2023-05-26
Refactor to favor  a request's broadcasted_at over created_at

mattwr18:
Closes https://github.com/tactilenews/100eyes-closed/issues/217

Originally, created_at and broadcasted_at were virtually identical as we would create the request and call `#broadcast!` in an `after_create_commit`, but since the additiona of [Schedule requests to be delivered in future #1581](https://github.com/tactilenews/100eyes/pull/1581), we allow the ability to schedule requests and keep track of when they were actually broadcast.


mattwr18:
@drjakob and I had a look together. Can merge as is.

PR #1645, opened 2023-06-03
360Dialog WhatsApp

mattwr18:
### Client onboarding
- [x] Add integrated onboarding button (where should it live? Settings page?, Should we prefill the sign up form on 360dialog?)
- [x] Improve UX, favor modal over new tab
- [x] Consume redirect and schedule job to create api key
- [x] Successfully create api key through integrated onboarding process
- [ ] Way to set profile info/image by api?

### Configure instance for 360dialog WhatsApp
- [x] Configure webhook url
- [x] Create templates via API (currently dealing with error which has been reported to 360dialog support)

### Send/receive messages
- [x] Send template messages with api_key :point_up:
- [x] Receive message/handle them - request for more info, to unsubscribe, to re-subscribe, to receive latest message
- [ ] Add feature that allows contributor to receive previous requests
- [x] Send freeform text message
- [x] Send freeform file message
- [x]  Receive files

### Testing

- [x] Fix failing tests
- [x] Add automated tests

### Replies
- [ ] Support replies
         1. Process status webhook callback to save the whats_app_message_id `params[:statuses].first[:id]`
         2. When a contributor replies, we receive a context object with the message id, which can be used to lookup and/or display the message. `params[:messages].first[:context][:id]`


app/adapters/whats_app_adapter/outbound.rb
```diff
@@ -4,94 +4,25 @@ module WhatsAppAdapter
...
 
-        if files.blank?
-          WhatsAppAdapter::Outbound::Text.perform_later(recipient: recipient, text: message.text)
-        else
-          files.each_with_index do |file, index|
-            WhatsAppAdapter::Outbound::File.perform_later(recipient: recipient, text: index.zero? ? message.text : '', file: file)
-          end
-        end
+      def business_solution_provider
+        Setting.three_sixty_dialog_client_api_key.present? ? 'ThreeSixtyDialog' : 'Twilio'
```
roschaefer, line 25: Instead of returning a string here you could return a class

app/controllers/whats_app/three_sixty_dialog_webhook_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+                               { audio: %i[id mime_type sha256] },
+                               { errors: %i[code details title] },
+                               { document: %i[filename id mime_type sha256] },
+                               { location: %i[latitude longitude] },
+                               { contacts: [{ org: {} }, { addresses: [] }, { emails: [] }, { ims: [] },
+                                            { phones: %i[phone type wa_id] }, { urls: [] },
+                                            { name: %i[first_name formatted_name last_name] }] }])
+    end
+
+    def create_api_key_params
```
roschaefer, line 84: Could we maybe use a different controller for messaging and setup?

docker-compose.yml
```diff
@@ -8,6 +8,8 @@ x-defaults: &x-defaults
     POSTGRES_PORT: "${POSTGRES_PORT:-5432}"
     SIGNAL_CLI_REST_API_ENDPOINT: "http://signal:8080"
     SIGNAL_CLI_REST_API_ATTACHMENT_PATH: "signal-cli-config/attachments/"
+    THREE_SIXTY_DIALOG_PARTNER_REST_API_ENDPOINT: "https://hub.360dialog.io/api/v2"
+    THREE_SIXTY_DIALOG_WHATS_APP_REST_API_ENDPOINT: "https://waba.360dialog.io/v1"
```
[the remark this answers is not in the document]
mattwr18: I'll hold off on this change because you have a PR that would consolidate the setup jobs into one and that will affect this suggestion. Maybe you could even add it there?

app/controllers/whats_app/three_sixty_dialog_webhook_controller.rb
mattwr18: Yeah, originally I added this in a separate controller, see [`9468a86` (#1645)](https://github.com/tactilenews/100eyes/pull/1645/commits/9468a8650a33ac58c61b6133365a81fd6e450c27)

I put it here because I wanted to differentiate between client/contributor onboarding, but maybe we could separate it again and rename it to `WhatsApp::SetupController`, or something similar?

app/adapters/whats_app_adapter/outbound.rb
mattwr18: I like this suggestion and have followed it in 176fbf31

PR #1684, opened 2023-08-09
fix: seeds

roschaefer:
## Description

I'm getting into this project again and the first thing that pops up is
that in `db/seeds.rb` the records are not persisted.

 ## How to test

1. `bin/rails db:drop db:create db:migrate db:seed`
2. Does not crash anymore


drjakob:
Welcome back.

PR #1685, opened 2023-08-17
ci: check for unstaged files

roschaefer:
When you do the following
```
$ asdf install
$ npm --version
6.14.16
```

and run:

```
npm install
```

~then you get a changed `package-lock.json`.~

It is good practice to always check in lock files into version control.
On CI we run `npm install` (among other things). If we make sure that we
do not leave any updates files, this ensures we always commit our lock
files.

EDIT: You don't, my fault. But it does not hurt to merge this PR anyways.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,3 +64,9 @@ jobs:
 
       - name: Run tests
         run: bin/rake spec
+
+      - name: Check unstaged changes
+        run: |
+          git --no-pager diff --color
+          git update-index --refresh
+          git diff-index --quiet HEAD


.tool-versions
```diff
@@ -1,3 +1,3 @@
 ruby 3.0.0
-nodejs lts-erbium
+nodejs 12.22.12
```
roschaefer, line 2: So this was necessary for the Github action `marocchino/tool-versions-action` to work. However it is now considered bad practice to set codename versions: https://github.com/asdf-vm/asdf-nodejs#partial-and-codename-versions

roschaefer:
```
$ asdf nodejs resolve lts-erbium --latest-available
12.22.12
```

mattwr18 (commented):
I have to admit I don't understand what the purpose of these changes are.
Could you please update the description as it is now outdated and it seems has nothing to do with these changes.

PR #1687, opened 2023-08-29
fix(signal-cli-rest-api): native mode

roschaefer:
## Changes

The `USE_NATIVE` flag is deprecated, see: https://github.com/search?q=repo%3Abbernhard%2Fsignal-cli-rest-api%20USE_NATIVE&type=code

As you can see in the search results, the `USE_NATIVE` is ignored. So
I guess, we were using normal mode all the time. In this case the change
will have an effect on the behaviour of `signal-cli-rest-api`.

 ## How to test

1. `docker compose up signal`
2. Deprecation warning is gone

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,7 +28,7 @@ services:
   signal:
     image: bbernhard/signal-cli-rest-api:0.67
     environment:
-      - USE_NATIVE=0
+      - MODE=native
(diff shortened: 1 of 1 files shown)


mattwr18 (approved):
I noticed this before, but never put in a PR for it. Thanks for doing that @roschaefer

PR #1693, opened 2023-09-05
ci: use asdf versions, avoid codenames

roschaefer:
Motivation
----------
Follow-up of #1685 @mattwr18 mentioned it's doing too many things.

How to test
-----------
1. Check if CI fails


mattwr18 (approved):
Looks good! Thanks @roschaefer 
I like the idea of ensuring that we run our ci with the same versions we have configured for development. We just need to make sure that if we change these in production, we also update the `.tool-versions`

PR #1694, opened 2023-09-05
fix: handle messages without text or caption

roschaefer:
Motivation
----------
If our webhook controller responds with 500 then Telegram API will try to send the message over and over.

We have a particular message that we did not expect: It's a message without text (which can happen) and we receive it before the user has onboarded herself. E.g you could simulate this by sending a sticker to `MAZ_Potsdam_Bot`.

How to test
-----------
1. `bin/rspec ./spec/adapters/telegram_adapter/inbound_spec.rb`
2. See:
```
  #consume
    |message| block argument
      ...
      given the payload does not have a caption nor text
        is expected to be nil
        when UNKNOWN_CONTRIBUTOR is registered
          is expected to yield control
```
image

### DIFF
--- a/app/adapters/telegram_adapter/inbound.rb
+++ b/app/adapters/telegram_adapter/inbound.rb
@@ -74,8 +74,14 @@ module TelegramAdapter
         return sender
       end
 
+      unless text
+        trigger(UNKNOWN_CONTRIBUTOR)
+        return nil
+      end
+
       telegram_onboarding_token = text.delete_prefix('/start').strip.upcase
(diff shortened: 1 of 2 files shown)


roschaefer:
I can actually see myself on Sentry :heart: [dced1fe97cda4e889da25b654a6b7aab](https://tactilenews.sentry.io/issues/4211656626/events/dced1fe97cda4e889da25b654a6b7aab/)

mattwr18:
Sorry, I forgot to give you an official review @roschaefer ... I had some question about whether we should send the same unknown contributor message, but after translating it, I think it works well. Thanks for quickly fixing this.

PR #1695, opened 2023-09-06
Ignore bot was kicked from chat message from Telegram webhook

mattwr18:


mattwr18:
Closing this for now as we've only ever seen the need to rescue this `TelegramBot::Forbidden` error from the webhook controller once and it has now resolved itself

PR #1696, opened 2023-09-12
Update signal-cli-rest-api with more verbose error messages and report them

mattwr18:
This changes how we report errors with Signal. Previously we just called `response.value`, which we know will raise an exception if the response is not successful. We then rescued the exception and provided as much context as we could.

- If this is part of the error message, we want to parse out the challenge token and tag the error with this token so that we can set up a Sentry filter to send this error to a specific channel whose only responsibility will be to alert a dev they need to manually solve a captcha a run signal-cli submitRateLimitChallenge --challenge-token <TOKEN> --captcha <CAPTCHA>

### DIFF
--- a/app/adapters/signal_adapter/api.rb
+++ b/app/adapters/signal_adapter/api.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module SignalAdapter
+  class Api
+    class << self
+      def perform_request(request)
+        uri = request.uri
+        response = Net::HTTP.start(uri.host, uri.port) do |http|
+          http.request(request)
+        end
+        case response
+        when Net::HTTPSuccess
+          yield response if block_given?
+        else
+          error_message = JSON.parse(response.body)['error']
+          exception = SignalAdapter::BadRequestError.new(error_code: response.code, message: error_message)
+          context = {
+            code: response.code,
+            message: response.message,
+            headers: response.to_hash,
+            body: error_message
+          }
+          ErrorNotifier.report(exception, context: context)
+        end
+      end
+    end
+  end
+end
--- a/app/adapters/signal_adapter/outbound/file.rb
+++ b/app/adapters/signal_adapter/outbound/file.rb
@@ -9,16 +9,17 @@ module SignalAdapter
 
       def perform(message:)
         @message = message
-        url = URI.parse("#{Setting.signal_cli_rest_api_endpoint}/v2/send")
-        request = Net::HTTP::Post.new(url.to_s, {
+        uri = URI.parse("#{Setting.signal_cli_rest_api_endpoint}/v2/send")
+        request = Net::HTTP::Post.new(uri, {
                                         Accept: 'application/json',
                                         'Content-Type': 'application/json'
                                       })
         request.body = data.to_json
-        response = Net::HTTP.start(url.host, url.port) do |http|
-          http.request(request)
+        SignalAdapter::Api.perform_request(request) do
+          # TODO: Do something on success. For example, mark the message as delivered?
+          # Or should we use deliver receipts as the source of truth.
+          Rails.logger.debug 'Great!'
         end
-        handle_response(response)
       end
 
       def data
@@ -31,25 +32,6 @@ module SignalAdapter
           message: message.text,
           base64_attachments: base64_files
         }
-      end
-
-      def handle_response(response)
-        case response.code.to_i
-        when 200
-          # TODO: Do something on success. For example, mark the message as delivered?
-          # Or should we use deliver receipts as the source of truth.
-          Rails.logger.debug 'Great!'
-        when 400..599
-          error_message = JSON.parse(response.body)['error']
-          exception = SignalAdapter::BadRequestError.new(error_code: response.code, message: error_message)
-          context = {
-            code: response.code,
-            message: response.message,
-            headers: response.to_hash,
-            body: error_message
-          }
-          ErrorNotifier.report(exception, context: context)
-        end
       end
     end
   end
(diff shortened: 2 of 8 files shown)


docker-compose.yml
```diff
@@ -26,7 +26,7 @@ services:
       - postgres_data:/var/lib/postgresql/data
 
   signal:
-    image: bbernhard/signal-cli-rest-api:0.67
+    image: tactilenews/signal-rest-api:full_error
```
roschaefer, line 29:
```suggestion
    image: bbernhard/signal-cli-rest-api:0.119-dev
```

app/adapters/signal_adapter/outbound/text.rb
```diff
@@ -36,6 +29,32 @@ def data
...
+            body: error_message
+          }
+          tags = if error_message.match?(/with challenge token /)
+                   regex = /"(.*?)"/
+                   challenge_token = error_message.split('with challenge token ').last.match(regex).to_s.split(/"/).last
+                   { challenge_token: challenge_token }
+                 else
+                   {}
+                 end
+          ErrorNotifier.report(exception, context: context, tags: tags)
```
roschaefer, line 55: I would include a default here for every other response code.

app/adapters/signal_adapter/outbound/text.rb
```diff
@@ -19,14 +19,7 @@ def perform(recipient:, text:)
         response = Net::HTTP.start(url.host, url.port) do |http|
           http.request(request)
         end
-        response.value # may raise exception
-      rescue Net::HTTPClientException => e
```
roschaefer, line 23: If you remove the `rescue` then you will not even get to `handle_response` if an exception is raised.

roschaefer: Ok, I see from the comment now that `response.value` may raise the error and it was also removed, nevermind. But I find the behaviour surprising.

roschaefer (changes requested):
The PR looks good! Only thing is I would expect a unit test here.

app/adapters/signal_adapter/outbound/text.rb
mattwr18: Yeah, I find the behavior as it exists surprising, to be honest. We call `response.value` because if it is an unsuccessful response calling `.value` on it raises an exception, we even explain that the only purpose for line `22` is that it may raise an exception.

I think a better way to handle it is to either check the class of the response, or check its code to determine whether the response was successful, or not.

app/adapters/signal_adapter/outbound/text.rb
mattwr18: To be honest, as I wrote to you in Slack, I think this is overkill and makes it harder to maintain.
The error message now, since you added the `stdout` to the output can contain several different clues as to why the message failed to send and I don't like the idea of parsing the error messages for several different reasons and doing different things based on that information.

For example, we now have the full message that a number is not registered with Signal any longer if a contributor unregistered their number.

I think it is enough to send the entire message to sentry, as I did:

```
error_message = JSON.parse(response.body)['error']
exception = SignalAdapter::BadRequestError.new(error_code: response.code, message: error_message)
```

and continue sending the errors to the #monitoring channel. This channel is for actionable errors that a dev should look into/do something about.

We have another channel #support for non-actionable errors that might require some support from a non-dev.

mattwr18:
> The PR looks good! Only thing is I would expect a unit test here.

There were actually already unit tests for the sad path, which were failing after the refactor, I've updated them. See 9ad81b39

app/adapters/signal_adapter/outbound/file.rb
```diff
@@ -9,23 +9,17 @@ class File < ApplicationJob
 
       def perform(message:)
         @message = message
-        url = URI.parse("#{Setting.signal_cli_rest_api_endpoint}/v2/send")
-        request = Net::HTTP::Post.new(url.to_s, {
+        uri = URI.parse("#{Setting.signal_cli_rest_api_endpoint}/v2/send")
+        request = Net::HTTP::Post.new(uri, {
```
roschaefer, line 13: It matters that we pass `URI` not string.

app/adapters/signal_adapter/api.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module SignalAdapter
+  class Api
+    class << self
```
roschaefer, line 5: I guess a class with a static method will do. `ErrorNotifier` is built the same way`.

docker-compose.yml
```diff
@@ -26,7 +26,7 @@ services:
       - postgres_data:/var/lib/postgresql/data
 
   signal:
-    image: bbernhard/signal-cli-rest-api:0.67
+    image: bbernhard/signal-cli-rest-api:0.119-dev
```
roschaefer, line 29: @mattwr18 Are we going to roll our own image or not?

roschaefer (approved):
Code-wise, this can be merged. The only critical issue is which image we're going to deploy. Yesterday I was convinced by @mattwr18 to use our own image, because upstream has also unrelated changes, including newer version of `signal-cli` which causes issues when downgrading.

app/adapters/signal_adapter/outbound/file.rb
mattwr18: Ok, this is just a copy/paste that survived from code you added https://github.com/tactilenews/100eyes/commit/d521f28e919b89b50cc7977eeca941e740cca88c#diff-2e66abeab283cf6d1ae8e78dbcd46410ae0b8eb01361bc44708e1426c6f89f28R31

I don't know why it matters. Maybe, you could further explain

docker-compose.yml
mattwr18: Don't think so. There is already a pre-release https://github.com/bbernhard/signal-cli-rest-api/releases/tag/0.68-pre

There, he advises to use this image for testing things out.

app/adapters/signal_adapter/api.rb
mattwr18: Sure, don't know why I didn't think of that hehe

PR #1704, opened 2023-09-28
Lock signal-cli to 0.11.11/maintain native mode

### LINKED ISSUE #1703: Lock signal-cli at 0.11.11 while maintaining `native` mode
### Motive:

We actually inadvertently changed the mode of `signal-cli` from `normal` mode to `native` mode in [fix(signal-cli-rest-api): native mode #1687](https://github.com/tactilenews/100eyes/pull/1687). This, in combination with upgrading `signal-cli-rest-api` to `0.68`, which upgrades `signal-cli` to `0.12.1` and `libsignal` to `0.30.0` breaks the implementation of `Signal`.

We could simply change back to using `normal` mode, which we have done since we introduced Signal, but we could benefit from `native` mode since it is much faster and uses less memory. We have periodically dealt with issues related to lack of `RAM` and have subsequently upgraded all our droplets, since they all use Signal. See https://tactilenews.slack.com/archives/C010QDQLCKW/p1670574356400719

The idea is to lock the `signal-cli` version to `0.11.11` and still make use of recent changes to error handling, introduced by @roschaefer in [fix: include $stdin in server response on error #404](https://github.com/bbernhard/signal-cli-rest-api/pull/404). We will then monitor the status of the issue with upgrading and decide either to upgrade after the issue is resolved upstream, or go back to `normal` mode.

mattwr18:
- upgrading to >= v0.12.0 breaks the implementation while in native mode

Closes #1703

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,7 +26,7 @@ services:
       - postgres_data:/var/lib/postgresql/data
 
   signal:
-    image: bbernhard/signal-cli-rest-api:0.119-dev
+    image: tactilenews/signal-cli-rest-api:latest
     environment:
       - MODE=native
         #- AUTO_RECEIVE_SCHEDULE=0 22 * * * #enable this parameter on demand (see description below)


mattwr18:
Superseded by https://github.com/tactilenews/100eyes/pull/1714

PR #1728, opened 2023-11-23
refactor: conditional routing

roschaefer:
refactor: conditional routing

Motivation
----------
It took me a while to debug the 404 error when clicking on the link to:
```
http://localhost:3000/onboarding/whats-app?jwt=eyJhbGc
```

I was expecting to see a problem in `config/routes.rb` but couldn't find anything. It was surprising to me that we programmatically throw a 404 in one of the controllers.

How to test
-----------
1. `bin/rspec ./spec/adapters/signal_adapter/api_spec.rb`
2. Tests pass

Requested changes
-----------------
@mattwr18 asked me to implement this consistently across our codebase. I have seen this only once, here's the commit that introduced it: 9a541d69fd8b41ce269f67eb11dc2fbccf9f5160.
Thankfully there is already a test (that still fails if you remove conditional routing).

In the tests there was:
```
unless Setting.signal_server_phone_number
  allow(Setting).to receive(:signal_server_phone_number).and_return('SIGNAL_SERVER_PHONE_NUMBER')
  # ..
end
```
from 07694a6fe58e1b7b47abf0e5eda1e90663f0a519.

@mattwr18 this looks intentional, why was that? Did I miss anything?

### DIFF
--- a/spec/jobs/signal_adapter/receive_polling_job_spec.rb
+++ b/spec/jobs/signal_adapter/receive_polling_job_spec.rb
@@ -45,10 +45,7 @@ RSpec.describe SignalAdapter::ReceivePollingJob, type: :job do
       before do
         create(:request)
 
-        unless Setting.signal_server_phone_number
-          allow(Setting).to receive(:signal_server_phone_number).and_return('SIGNAL_SERVER_PHONE_NUMBER')
-        end
-
+        allow(Setting).to receive(:signal_server_phone_number).and_return('SIGNAL_SERVER_PHONE_NUMBER')
         allow(job).to receive(:ping_monitoring_service).and_return(nil)
       end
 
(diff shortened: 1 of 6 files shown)


spec/jobs/signal_adapter/receive_polling_job_spec.rb
```diff
@@ -187,10 +182,8 @@
     describe 'given a contributor who has unsubscribed and requests to resubscribe',
              vcr: { cassette_name: :receive_signal_message_to_resubscribe } do
       before do
-        unless Setting.signal_server_phone_number
```
roschaefer, line 190: This looks intentional, why was this put in place? Did I miss sth. @mattwr18 ?

mattwr18: @roschaefer we already had a conversation about this code in another PR; unfortunately, I can't find it. It was actually you that added this code in https://github.com/tactilenews/100eyes/commit/5845dc4ecefd1bc0ea6538733aa7ddee9a52b4d5... You did not go into great detail why you added it, but the commit message states that you were fixing ci. It then has been refactored and copy/pasted throughout the spec file.

I can say for my part that I find the code useful the first time I record the cassette so that I can get some actual data back when I run the receive messages rake task. After the cassette is recorded, I think we can and probably should delete the code.

I think just having the line:

```
allow(Setting).to receive(:signal_server_phone_number).and_return('SIGNAL_SERVER_PHONE_NUMBER')
```

was enough to get the ci to pass 2 years ago, but I think you maybe wanted to continue using what you had configured in `.env` :shrug:

mattwr18 (approved):
Looks good... thanks :)

spec/jobs/signal_adapter/receive_polling_job_spec.rb
roschaefer: As discussed in our pairing, it's too confusing, so let's just remove it. Maybe we can check if we're in `record` mode instead of checking the configuration.

PR #1729, opened 2023-11-24
fix: broken invitation link

roschaefer:
merge after #1728

Motivation
----------
So this is a follow up for #1728.

It will prevent rendering the whatsapp link unless activated.

We still have an inconsistency, because the channel links depend on
`Setting.channels` whereas the routing depends on
`Setting.whatsapp_configured?`. Ie. `Setting.channels` can be activated
for whatsapp although `Setting.whatsapp_configured?` is false.

How to test
-----------
1. `bin/rspec spec/system/onboarding/index.spec.rb`
2. See:
```
Onboarding
  visit /onboarding/
    if WhatsApp was explicitly activated or activated by configuration
      renders invitation link for WhatsApp
    but if WhatsApp is deactivated
      renders no invitation link for WhatsApp
```

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -12,6 +12,11 @@ THREEMARB_API_IDENTITY=
 THREEMARB_API_SECRET=
 THREEMARB_PRIVATE=
 
+TWILIO_ACCOUNT_SID=
+TWILIO_API_KEY_SID=
+TWILIO_API_KEY_SECRET=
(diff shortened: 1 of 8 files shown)


mattwr18:
Closing this PR as some of the changes were introduced in #1728 and the other changes will be introduced in https://github.com/tactilenews/100eyes/pull/1818 along with expanding the changes to include all channels.

PR #1730, opened 2023-11-24
Improve new instance generate config

mattwr18:


ansible/generate_config.sh
```diff
@@ -84,6 +84,12 @@ rails:
     api_key:
       sid: ""
       secret: ""
+  three_sixty_dialog:
```
roschaefer, line 87: Yesterday in our pairing @mattwr18 told me that the motivation for this change is to avoid missing keys. Turns out, ansible's `default` handles missing keys:
https://github.com/ansible/ansible/issues/18102#issuecomment-504472825
https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.8.html#jinja-undefined-values

But I'm not opposed to listing all possible values in the config!

@mattwr18 should we add the `default('')` to all the environment variables in `ansible/roles/installation/tasks/main.yml`? This would handle missing keys in e.g. outdated inventories.

It's annoying that `ansible-vault` encrypts the entire file, not just the values.

roschaefer (approved):
I've nothing against merging this PR as it is. However, see my comments below.

below

ansible/generate_config.sh
mattwr18: I updated all the production instances with the key/value pairs to avoid the playbook erroring out already. The generate config script also adds the config for Twilio and 360dialog even though only one should be used. Still I have added the defaults so that if/when we change implementation the older instances would not need to have their config updated.

PR #1762, opened 2023-12-09
[Feature] Add ability to delete planned requests

### LINKED ISSUE #1753: Add ability to delete planned/scheduled requests
```
As an editor
I would like to be able to delete a planned request
If I no longer want it to go out
```

mattwr18:
Closes #1753


spec/system/requests/scheduling_requests_spec.rb
```diff
@@ -15,6 +15,7 @@
     it 'schedules a future job' do
       visit new_request_path(as: user)
 
+      expect(page).not_to have_link(I18n.t('components.request_form.planned_request.destroy'))
```
roschaefer, line 18: Checking for some text not to present on a website is an assertion that quickly turns into a false-negative. It's too easy for some random text not to be present on a website.

My suggestion is to either delete the assertion or to have a test case where the same link is in fact asserted to be present.

roschaefer (approved):
This one is a very solid PR! I haven't tested it but code-wise it looks very straight-forward and readable.

nailed it

spec/system/requests/scheduling_requests_spec.rb
mattwr18: I agree that it can quickly turn into a false negative. For example, if the button text changes, the test could still pass even though the button is actually visible. I tried to mitigate this by using the `I18n` translation so that if we do change the text of the button, it would also change the test assertion.

I do have a test where the link is asserted to be present. This line, https://github.com/tactilenews/100eyes/pull/1762/files#diff-1ca2f0f20382968b26e4448455a8a781d0a7f9a02acbb1e54c39b24cc355f51dR40

```rb
click_on I18n.t('components.request_form.planned_request.destroy')
```
will fail if there is no button to click.

The reason I added the test is that we conditionally render the delete button for requests that have been created and are able to be edited. Maybe, I could write a component spec instead?

mattwr18:
> This one is a very solid PR! I haven't tested it but code-wise it looks very straight-forward and readable.



@roschaefer it turns out that I broke the schedule request feature in commit https://github.com/tactilenews/100eyes/pull/1762/commits/b213f6fd3e26c88f50fa995462767cd8e9475e74

I fixed it in b84d258b and attempted to explain why it was broken. See if it makes sense to you.

spec/system/requests/scheduling_requests_spec.rb
mattwr18: See https://github.com/tactilenews/100eyes/pull/1762/commits/e87fb99d82d9fa6d029ac35d2ffa6a300ea2d8e6

PR #1781, opened 2024-02-09
Add rake tasks to create mock data in database.

phoeinx:
<details>
<summary>Previous PR description</summary>

This PR includes a rake task to create mock data in the database. (e.g. to evaluate UX of a system with 500 contributors.)

In its current state the script still requires some manual changes on the code base of the instance it's applied to (e.g. to disable validations for elements that are a bit more complicated to mock). I found this a valid trade-off time-wise (as I assumed that we would not use this script regularly). However, if we end up using it a bit more, it might make sense to automate/mock those parts better.

The necessary changes are documented at the top of the file.

Additionally the script currently only takes the number of contributors as a parameter, always creates 14 requests and for each request approx. messages from 50% of the contributors. (By basically just creating 14 * 1/2 * #contributor many messages.)

Basic usage is as follows:

```
rake fakeout:clean
rake fakeout:medium RAILS_ENV=development
``` 

where medium generates 500 contributors.
</details>

I changed the seed script substantially. I would suggest that we never run it on any deployed instance. Instead, let's run it locally and then upload and restore a local backup.

Usage:
```
bin/rails db:seed:fill
```


roschaefer:
@phoeinx I changed the code substantially and wanted to ask you for a review but I can't because you're the PR author on Github.

roschaefer:
I cannot simulate the same performance issues locally. I used `git checkout 7dc533155d8217d753910518711bf77e73836b24`.

phoeinx:
@roschaefer thanks for taking over! I don't have time to review this week, so if it's time critical please ask somebody else. Otherwise I'll have a look next week.

Regarding the unrepeatability - have you tested it with a similar setup (e.g. 2GB RAM) as on the production server? Sadly, cannot go back more than 2 weeks in the load data of the server to see if that was the issue, but that might be worth a try.

db/seeds/fill.rb
```diff
@@ -0,0 +1,61 @@
...
+    unknown_content { false }
+    broadcasted { false }
+    sender { contributors.sample }
+    recipient { nil }
+    request { requests.sample }
+  end
+end
+
+Rails.logger.debug 'Seeding messages..'
+FactoryBot.create_list(:message, message_count)
```
mattwr18, line 61: I tried this out a few weeks ago, before I saw Robert's comments about the code changing and the motive behind it. I also like the idea of seeding the db locally and restoring it to a production instance.

I mentioned this to @roschaefer already, but I wanted to review it. I ran into an issue when I ran the rake task because I had some of my own contributors already in the db, and the creation of new messages resulted in our after_create_commit being run:

```
after_create_commit :send_if_outbound
```

I killed all jobs as quickly as possible, but not before several messages went out to WhatsApp and Threema contributors, which are paid messages. I also commented out  the after_create_commit and re-ran the script, but ideally one wouldn't need to remember doing this.

@roschaefer do you have any experience with skipping callbacks? I saw there are a few ways to do it. This issue on factory_bot discussing one way that we could probably easily implement here See https://github.com/thoughtbot/factory_bot/issues/931#issuecomment-307542965

Other than that, I think it's really nice to have this. I would like to re-seed my db to test out the performance improvements I am working on at the moment

mattwr18 (changes requested):

spec/factories/contributors.rb
```diff
@@ -5,9 +5,7 @@
     first_name { 'John' }
     last_name { 'Doe' }
     data_processing_consent { true }
-    sequence :email do |n|
-      "contributor#{n}@example.org"
-    end
+    email { Faker::Internet.email }
```
mattwr18, line 8: this change is related to this PR since the seed rake task fails when trying to create simply sequential emails.

app/models/request.rb
```diff
@@ -24,7 +24,7 @@ class Request < ApplicationRecord
 
   acts_as_taggable_on :tags
 
-  after_create { Request.broadcast!(self) }
+  after_create :broadcast_request
```
mattwr18, line 27: I didn't see a way to easily skip this callback without turning it into a method that can be skipped in the seeds file.

db/seeds/fill.rb
mattwr18: I went ahead and added the code I was suggesting here. See 2d4015f2.

soey:
What is the status of this? Can I help somehow?

soey:
Good to have this. I add my findings here ..

To be changed: 

- [ ] The factory creating the contributors creates an unneeded organizations for (1 for each contributor)

db/seeds/fill.rb
soey: Well it changes the class itself instead of the instance. But I think that won't have any side effects as we want this behavior for the whole seeding script. Seems to work. Fine for me

spec/factories/contributors.rb
```diff
@@ -18,6 +16,8 @@
       end
     end
 
-    organization
+    trait :with_new_organization do
+      organization
+    end
```
mattwr18, line 21: First of all, I wanted to say that I am a huge fan of this improvement!
We definitely benefit from not needlessly creating an organization for **every** contributor!! :heart_eyes: 

I think my idea when I added the organization to the contributors factory was that it should by default assign a contributor to _an_ organization because that is the most used case, where we have a single organization and almost all contributors belong to the organization. It is optional because we want to be able to have contributors, like myself, who would not count in the total number of contributors allowed for an organizations business plan. I must have overlooked that it would create a new organization to assign the contributor to.

I would be in favor of removing this trait, as it is currently not being used anywhere.
I know there are places where we specify in the tests that we want a contributor assigned to a specific organization. If we find the need to create an organization in this way, we can always add back the trait when we actually use it.

What do you think? I'll approve the PR and leave you to remove it unless you want to explain why you prefer to leave it.

mattwr18 (approved):

PR #1791, opened 2024-02-26
Optimize requests show page

### LINKED ISSUE #1783: Optimize db queries for requests show page
Related to [[Tracking issue] Optimize db queries #1778](https://github.com/tactilenews/100eyes/issues/1778)

mattwr18:
Closes #1783


app/controllers/requests_controller.rb
```diff
@@ -9,8 +9,8 @@ class RequestsController < ApplicationController
 
   def index
     @filter = filter_param
-    @sent_requests_count = Request.include_associations.broadcasted.count
-    @planned_requests_count = Request.include_associations.planned.count
```
mattwr18, line 13: I don't know why I added this scope here, but it doesn't make any sense now. We simply want the count so we don't need to include, preload, or eager load anything.

app/controllers/requests_controller.rb
```diff
@@ -128,9 +128,9 @@ def filter_param
 
   def filtered_requests
     if @filter == :planned
-      Request.planned.reorder(schedule_send_for: :desc).include_associations
+      Request.planned.reorder(schedule_send_for: :desc).includes(:tags)
     else
-      Request.broadcasted.include_associations
+      Request.broadcasted.includes(:tags)
```
mattwr18, line 133: On the request index page, we don't use anything that needs to be eager loaded other than the tags.

app/models/request.rb
```diff
@@ -37,9 +38,9 @@ def personalized_text(contributor)
   def stats
     {
       counts: {
-        recipients: messages.map(&:recipient_id).compact.uniq.size,
-        contributors: messages.select(&:reply?).map(&:sender_id).compact.uniq.size,
-        photos: replies.pluck(:photos_count).sum,
+        recipients: outbound.select(:recipient_id).distinct.count,
+        contributors: replies.select(:sender_id).distinct.count,
+        photos: replies.sum(:photos_count),
```
mattwr18, line 43: Favor returning as little as possible from the db and handling count/sum with postgres instead of ruby.

spec/models/request_spec.rb
```diff
@@ -244,22 +245,22 @@
...
-
-      describe 'iterating through a list' do
-        subject { -> { Request.find_each.map(&:stats) } }
-        it { should make_database_queries(count: 35) }
-
-        describe '::include_associations' do
-          subject { -> { Request.include_associations.find_each.map(&:stats) } }
-          it { should make_database_queries(count: 7) } # better
-        end
-      end
```
mattwr18, line 262: After the changes in this PR, there are actually less db calls without eager loading than with. The last counts from these tests where that the db queries went from the 35 expected to 10 and from 7 with `::include_associations` to 13.
Since using `::include_associations` actually adds three db calls, it has been removed.

PR #1795, opened 2024-02-27
refactor: merge Photo model with Message::File

### LINKED ISSUE #1387: Refactor attachment handling
We currently handle only some types of attachments explicitly (images, audio messages). We store attachments There are a couple of problems with that:

* We frequently receive messages that contain other types of attachments (PDF files, vCards, plaintext attachments, …). It’s difficult to add support for new content types and it’s an annoying experience for newsrooms, because they have to wait for us to manually retrieve the attachments if we don’t have support for the respective attachment type.

* We currently store attachments inconsistently. We have a generic `Message::File` model, but we also have a `Photo` model. This leads to bugs, e.g. #813.

I suggest the following:

1. Refactor our codebase to store all attachments we can currently process using the generic `Message::File` model (or maybe even without that intermediate model, simply using ActiveStorage directly with the `Message` model.

2. The only difference between attachment types would then be their presentation, i.e. images are displayed as thumbnails, audio files are rendered with an audio player. But the data model and storage should be the same for all types.

3. Inbound channel adapters should then store attachments of any type. For types that we don’t support explicitly, we would simply display download links or similar. We should consider security in this case and probably display a warning message to make sure editors are aware of the implications of downloading random files.

roschaefer:
close #1387

### DIFF
--- a/app/adapters/postmark_adapter/inbound.rb
+++ b/app/adapters/postmark_adapter/inbound.rb
@@ -30,7 +30,7 @@ module PostmarkAdapter
       @photos, @unknown_content = initialize_photos_and_unknown_content(mail)
       @message.unknown_content = unknown_content
       @photos.each do |photo|
-        @message.association(:photos).add_to_target(photo)
+        @message.association(:files).add_to_target(photo)
       end
     end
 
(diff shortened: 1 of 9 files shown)


soey:
@roschaefer : WDYT should we rebase this and move it forward until being mergeable?

roschaefer:
> @roschaefer : WDYT should we rebase this and move it forward until being mergeable? 

If it's a priority, we can!

mattwr18 (approved):
Thanks @roschaefer, I rebased and fixed some failing tests by removing some associations that are now superfluous.
Also, I ran a script to transfer any photos as files and removing the photos. In a second step, I would like to drop the photos table.

PR #1798, opened 2024-03-11
Eager load contributor avatars on dashboard

### LINKED ISSUE #1796: Optimize db queries on dashboard
Related to #1778

mattwr18:
Closes #1796

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -14,7 +14,7 @@ class DashboardController < ApplicationController
   def activity_notifications
     grouped = current_user.notifications_as_recipient
                           .newest_first
-                          .includes(:contributor, :request, :message, :user)
+                          .includes({ contributor: { avatar_attachment: :blob } }, :request, :message, :user)
                           .last_four_weeks
                           .limit(100)
                           .group_by { |notification| notification.to_notification.group_key }


mattwr18:
@soey @roschaefer I wanted to get your feedback on this. 

### Steps to reproduce
* Create ActivityNotifications of each type. Currently, we have:

```
ChatMessageSent
ContributorMarkedInactive
ContributorSubscribed
MessageReceived
OnboardingCompleted
RequestScheduled
```

For me, the easiest way to create these notifications is to search for each of these in the repo, you will see the api for creating a notification of each of these types. For example:

```
MessageReceived.with(contributor_id: sender_id, request_id: request.id, message_id: id).deliver_later(User.all)
```
* tail the bullet logs `tail -f log/bullet.log`
* Visit the dashboard at `localhost:3000/dashboard`
* Notice that the suggestions are contrary to each other.

Here is the issue, as I understand it. We eager load resources that might be needed by the notification in the [dashboard_controller.rb](https://github.com/tactilenews/100eyes/blob/main/app/controllers/dashboard_controller.rb#L17). For some, we do not need to eager load the messages, or the user so we should avoid eager loading; however, we grab all notifications, irrespective of type and then group them to be displayed on the dashboard. We could silence these suggestions by making separate db calls for each notification type, only eager loading the necessary resources, but I think this would result in more db calls.

Thoughts?

soey:
Second: Maybe we do a quick pairing on this so that I can get the idea without digging in.. My feeling is that an additional eager loading is not that costly as a missing one.

soey:
> First thought: What about to add some seed data including notifications etc. so that we have a prepopulated db for local development so that we don't have to create needed data individually..?

Ah.. just stumbled upon: https://github.com/tactilenews/100eyes/pull/1781 What about this PR? This should do sth. like I suggest..

mattwr18:
> > First thought: What about to add some seed data including notifications etc. so that we have a prepopulated db for local development so that we don't have to create needed data individually..?
> 
> Ah.. just stumbled upon: #1781 What about this PR? This should do sth. like I suggest..

Yes, I want to get that PR in, and for the most part it would create the notifications because I purposely did not skip them. I am also keen for a pairing session.

PR #1802, opened 2024-03-11
Feature: Add ability to deactivate user

### LINKED ISSUE #1777: Deactivate editors
As admin
I would like to deactivate editors.
in order to use my paid seats properly.

mattwr18:
- When deactivated user attempts to sign in, redirect with alert message

Fixes #1777

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -15,6 +15,8 @@ class User < ApplicationRecord
   validates :password, length: { in: 8..128 }, unless: :skip_password_validation?
 
   scope :admin, ->(boolean = true) { where(admin: boolean) }
+  scope :active, -> { where(deactivated_at: nil) }
+  scope :inactive, -> { where.not(deactivated_at: nil) }
 
   after_create_commit :notify_admin
   after_update_commit :reset_otp
@@ -25,6 +27,22 @@ class User < ApplicationRecord
 
   def avatar?
     false
+  end
+
+  def active?
+    deactivated_at.nil?
+  end
+
+  alias active active?
+
+  def inactive?
+    !active?
+  end
+
+  alias inactive inactive?
+
+  def active=(value)
+    self.deactivated_at = ActiveModel::Type::Boolean.new.cast(value) ? nil : Time.current
   end
 
   private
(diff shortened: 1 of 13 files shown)


config/locales/de.yml
```diff
@@ -680,6 +680,7 @@ de:
   sessions:
     errors:
       otp_incorrect: "Der 6-stellige Anmeldecode ist nicht korrekt."
+      deactivated_user: Dein Konto ist derzeit nicht aktiv, bitte kontaktierst du den Support unter support@tactile.news.
```
mattwr18, line 683: @drjakob can you ok this is what you want to say?

drjakob: "Dein Konto ist derzeit nicht aktiv, bitte kontaktiere den Support unter support@tactile.news"

spec/support/shared_examples/activity_notifications.rb
```diff
@@ -10,8 +10,8 @@
 
     it 'for each user' do
       run_action(subject)
```
soey, line 12: will an inactive user get notifications and what is expected in that regard?

soey:
Well this is one of those features that goes through the whole application ... you have to review all collection retrievals of users and have to consider which scope to be used. I've yet no overview on if there are more occurences somewhere..

app/models/user.rb
```diff
@@ -27,6 +29,22 @@ def avatar?
     false
   end
 
+  def active?
+    deactivated_at.nil?
+  end
+
+  alias active active?
```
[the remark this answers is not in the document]
mattwr18: Good question. To be honest, I copied and pasted the implementation for deactivated contributors for the most part, but even for contributors this these aliases don't seem to be used. There is a unit test for `contributor.active`, but it's not used in the application code. I prefer using predicate methods when I expect a boolean returned.

spec/support/shared_examples/activity_notifications.rb
mattwr18: Yes, good point. I'm not sure what the behavior should be in that case. As we discussed in our pairing session, the notifications are only internal to the site and are not sent by any messenger or email so they would not be able to see them unless someone were to reactivate them. I'm not sure exactly what the client would expect to occur in that case. I could see reasons for them to want the user to see, or not see, the notifications.

I'll ask @drjakob about it when I next speak with him.

PR #1803, opened 2024-03-11
Don't attempt to display contact person if blank

mattwr18:
- sometimes we set up an instance before we are giving the information on who the contact person is. Ideally, this would not result in an error if someone visits the profile page

### DIFF
--- a/app/components/profile_header/profile_header.html.erb
+++ b/app/components/profile_header/profile_header.html.erb
@@ -3,8 +3,10 @@
     <%= t('.main_heading', business_plan_name: organization.business_plan.name) %>
   <% end %>
 
-  <%= c 'heading', style: :gamma do %>
-    <%= t('.sub_heading.contact_person', contact_person_name: organization.contact_person.name, contact_person_email: organization.contact_person.email) %>
+  <% if organization.contact_person.present? %>
+    <%= c 'heading', style: :gamma do %>
+      <%= t('.sub_heading.contact_person', contact_person_name: organization.contact_person.name, contact_person_email: organization.contact_person.email) %>
+    <% end %>
   <% end %>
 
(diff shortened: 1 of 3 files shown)


soey:
lgtm

app/components/profile_header/profile_header.html.erb
```diff
@@ -3,8 +3,10 @@
     <%= t('.main_heading', business_plan_name: organization.business_plan.name) %>
   <% end %>
 
-  <%= c 'heading', style: :gamma do %>
-    <%= t('.sub_heading.contact_person', contact_person_name: organization.contact_person.name, contact_person_email: organization.contact_person.email) %>
+  <% if organization.contact_person.present? %>
```
roschaefer, line 6: I, too, give it a :+1: although I'd liked to see a little unit test.

roschaefer (approved):

app/components/profile_header/profile_header.html.erb
mattwr18: Ok, so I added a component spec for the ProfileHeader component where this conditional is.
I also decided I might as well add a test for each conditional in the component.
This lead me to realize I was misusing the Heading component being explicit and passing in the name of the tag I wanted to be used. See  [[a11y] Check that there are not duplicate h1 tags on pages #1812 ](https://github.com/tactilenews/100eyes/issues/1812)

I also introduced the idea of test id's to test the presence or absence of certain elements instead of testing the presence or absence of some text, which can be flaky. You can read more about this tactic at this blog post [Why Your Development Team Should Use data-testid Attributes](https://medium.com/@automationTest/why-your-development-team-should-use-data-testid-attributes-a83f1ca27ebb)

I will go ahead and merge this, but am open to any feedback you may want to give @roschaefer @soey

PR #1807, opened 2024-03-13
1806 docker setup on m1

soey:
Some suggestions that seems to fix the setup issues on my machine. As my knowledge in that area is fairly limited this has to be checked and tested by someone.


Gemfile.lock
```diff
@@ -429,6 +431,7 @@ GEM
     zeitwerk (2.6.6)
 
 PLATFORMS
+  arm64-darwin-23
```
roschaefer, line 434: Is this because you're using an apple laptop and ARM processor?

.env.template
```diff
@@ -1,6 +1,6 @@
 HUNDRED_EYES_PROJECT_NAME=100eyes
 POSTGRES_PASSWORD=
-POSTGRES_USER=
+POSTGRES_USER=postgres
```
roschaefer, line 3: I think this is not necessary, as it needs to be edited by the developer. But I'm super OK to merge in the other changes as long as they do not break our builds.

docker-compose.override.yml
```diff
@@ -35,6 +35,8 @@ services:
   assets:
     command: npm run dev
     <<: *x-dev-defaults
+    volumes:
```
roschaefer, line 38: This is redundant. I'm on `main` (28f5456ed44dd1337b00696981f8778c0a4cb70c) and when I do `docker compose config` I see this:
```
...
  assets:
    build:
      context: /home/robert/Development/tactile.news/100eyes-closed/100eyes
      dockerfile: Dockerfile
    command: npm run dev
    environment:
      CHROME_URL: http://chrome:3333
      HUNDRED_EYES_PROJECT_NAME: 100eyes
      MAILSERVER_HOST: mailserver
      MAILSERVER_PORT: 1025
      NODE_ENV: development
      POSTGRES_HOST: db
      RAILS_ENV: development
      WEBPACKER_DEV_SERVER_HOST: 0.0.0.0
    networks:
      default: null
    stdin_open: true
    tty: true
    volumes:
      - bind:
          create_host_path: true
        source: /home/robert/Development/tactile.news/100eyes-closed/100eyes
        target: /app
        type: bind
```

It already has the volume definition, it's on line 18 in this file.

roschaefer (changes requested):
long time no see

Hi @soey long time no see! Glad that you're taking the time to try out the setup and thank you so much for the PR. I will go ahead and merge the parts that may help developers on an apple machine.

By the way, I only dockerize the postgres service because the rest is installed natively on my system: `docker compose up db`. If you find it helpful we could add it to the README.

roschaefer (approved):

Gemfile.lock
soey: Yes - and i could either solve it by forcing nokogiri to use the ruby version `bundle config force_ruby_platform true` or adding my platform as recommended here: https://nokogiri.org/tutorials/installing_nokogiri.html#supported-platforms.

.env.template
soey: on my env it took "root" as default  and then exploded as there is no root user set on pgsql

docker-compose.override.yml
soey: I really don't understand this part. (I'm really guessing more than allowed here) Problem was that without it complained with:

```
assets-1      | > @ dev /app
assets-1      | > run-p -l 'build:** -- --watch'
assets-1      | 
assets-1      | sh: run-p: not found
assets-1      | npm ERR! code ELIFECYCLE
assets-1      | npm ERR! syscall spawn
assets-1      | npm ERR! file sh
assets-1      | npm ERR! errno ENOENT
assets-1      | npm ERR! @ dev: `run-p -l 'build:** -- --watch'`
assets-1      | npm ERR! spawn ENOENT
assets-1      | npm ERR! 
assets-1      | npm ERR! Failed at the @ dev script.
assets-1      | npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
assets-1      | npm WARN Local package.json exists, but node_modules missing, did you mean to install?
assets-1      | 
assets-1      | npm ERR! A complete log of this run can be found in:
assets-1      | npm ERR!     /root/.npm/_logs/2024-03-11T20_26_40_283Z-debug.log
assets-1      | 
assets-1 exited with code 1
```
this could either be resolved with an 
`docker compose run assets npm install`
what installed the node-modules in the app directory (should be noted in readme if we want this)
I think assume that by adding that volume declaration, I seem to override the volume linking from the defaults that might have created the issue? ?? <- just my guess

PR #1808, opened 2024-03-13
Avoid eager loading sender when it's not being used

### LINKED ISSUE #1799: Optimize db queries for contributors
Related to #1778 

### DevNotes

I think it's reasonable to optimize the queries on the contributors index and show pages.

mattwr18:
Closes #1799

### DIFF
--- a/app/models/contributor.rb
+++ b/app/models/contributor.rb
@@ -146,7 +146,7 @@ class Contributor < ApplicationRecord
   end
 
   def recent_replies
-    result = replies.includes(:request, :sender).reorder(created_at: :desc)
+    result = replies.includes(:request).reorder(created_at: :desc)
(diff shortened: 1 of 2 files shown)


app/models/contributor.rb
```diff
@@ -146,7 +146,7 @@ def tags?
   end
 
   def recent_replies
-    result = replies.includes(:request, :sender).reorder(created_at: :desc)
+    result = replies.includes(:request).reorder(created_at: :desc)
```
soey, line 149: isn't that needed for this line?
https://github.com/tactilenews/100eyes/blob/main/app/components/requests_feed/requests_feed.html.erb#L22

soey (approved):

PR #1818, opened 2024-03-28
Improve setting channels logic

### LINKED ISSUE #1817: [improvement] Channels in settings form should not be checked by default
```
As an admin in 100eyes
I want to see the channels that are configured in the list
So that we do not expose channels to onboard that are not configured
```

### Dev Notes

Only channels that are configured to be used should be shown in these settings and only configured channels should have a button to onboard new contributors.

Screenshot from 2024-03-28 14-08-31
Screenshot from 2024-03-28 14-12-32

mattwr18:
whether a channel is available for onboarding should be determined by whehter it is
  configured or not, to start. An admin should be able to make the
channel unavailable, but it should not be available if it is not
configured

Fixes #1817

### Dev Notes

It became clear while adding automated tests that the test env was actually loading values from a devs local .env environment. I believe this is less than ideal since we can have different .env values and some values I have configured are actually used to test that services work before deploying them.

The change was made to add a `.env.local.test` file and override the behavior in the test config. This change was necessary to add automated tests for the `spec/system/settings/onboarding_channels_spec.rb`


app/components/onboarding_channels_checkboxes/onboarding_channels_checkboxes.html.erb
```diff
@@ -7,10 +7,25 @@
...
+          <%= c 'checkbox', id: "setting[channels][#{key}][allow_onboarding]", checked: value[:allow_onboarding] %>
+          <label for=<%= "setting[channels][#{key}][allow_onboarding]" %>><%= key.to_s.camelize %>
+        </li>
+      <% else %>
+        <input
+          id='<%= "setting[channels][#{key}][allow_onboarding]" %>'
+          name='<%= "setting[channels][#{key}][allow_onboarding]" %>'
+          value="<%= value[:allow_onboarding] %>"
+          type="hidden"
+        >
```
mattwr18, line 27: I don't know if this is the best approach because hidden inputs are still available in the dom tree and could be modified by a malicious user. In this case, this is not a public website and this section is only visible to admin, so I think it's ok.

The reason this conditional exists is because I want to have the entire hash saved in Setting.channels sent from the form to the controller because when the update occurs it simply replaces the entire has with a potentially modified one.

config/routes.rb
```diff
@@ -20,21 +20,25 @@
...
-      get '/telegram/', to: 'telegram#show'
-      get '/telegram/link/:telegram_onboarding_token', to: 'telegram#link', as: 'telegram_link'
-      get '/telegram/fallback/:telegram_onboarding_token', to: 'telegram#fallback', as: 'telegram_fallback'
-      post '/telegram/', to: 'telegram#create'
+      constraints(-> { Setting.telegram_configured? }) do
+        get '/telegram/', to: 'telegram#show'
+        get '/telegram/link/:telegram_onboarding_token', to: 'telegram#link', as: 'telegram_link'
+        get '/telegram/fallback/:telegram_onboarding_token', to: 'telegram#fallback', as: 'telegram_fallback'
+        post '/telegram/', to: 'telegram#create'
+      end
```
mattwr18, line 39: This is partially clean up. Now that we have these class methods, we want to use them to not create routes for channels that are not configured.

config/environments/test.rb
```diff
@@ -7,6 +7,7 @@
 
 Telegram.reset_bots
 Telegram::Bot::ClientStub.stub_all!
+Dotenv.overload('.env.local.test')
```
soey, line 10: In general I like the idea to have a well defined set of env variables for the tests. Perfectly valid. The thing that worries me is that this gem is 4 years not maintained and not widely used. Im wondering why it has been in the gemfile if not used before this PR..

config/routes.rb
soey: Are those methods for the onboarding only? If so shouldn't they check for allow_onboarding?

spec/requests/onboarding/whatsapp_spec.rb
```diff
@@ -7,12 +7,11 @@
     subject { { get: '/onboarding/whats-app' } }
 
     describe 'when no Whatsapp number was configured' do
-      before { allow(Setting).to receive(:whats_app_server_phone_number).and_return('') }
+      before { allow(Setting).to receive(:whats_app_configured?).and_return(false) }
       it { should_not be_routable }
     end
 
     describe 'but when a Whatsapp number was configured' do
-      before { allow(Setting).to receive(:whats_app_server_phone_number).and_return('+49123456789') }
```
soey, line 15: The tests were more understandable with that mocked values. Now you have to understand that the Setting values fallback to the .env value. Maybe add a comment on that for each of the test just to be nice to everybody?

soey (changes requested):
lgtm besides possibly one small detail and eventually some comments

config/routes.rb
mattwr18: Yes, they are for onboarding. We filter out the available channels here https://github.com/tactilenews/100eyes/pull/1818/files#diff-c013d1db563a45b6fcd49779b9cade0a0748018246f3f8f919a7c87ac4f39c34R16, but if the route is available for a configured channel, like email(which needs to be configured for user managment), and the client wants to disallow onboarding, someone who knows what they are doing would still be able to onboard. So, good call, I'll updated these routes.

spec/requests/onboarding/whatsapp_spec.rb
mattwr18: >  Now you have to understand that the Setting values fallback to the .env value. 

I don't know if I understand this sentence correctly, but this is not how the Setting works for `whats_app_server_phone_number`. There are no fallbacks, if there is nothing configured in the .env variables it will return nil. See [Setting model](https://github.com/tactilenews/100eyes/blob/main/app/models/setting.rb#L86):

```rb
field :whats_app_server_phone_number, readonly: true, default: ENV.fetch('WHATS_APP_SERVER_PHONE_NUMBER', nil)
```

With the description:
> but when a Whatsapp number was configured

you need to understand how we configure messengers and need to understand that we have a dedicated `.env` for automated tests.

I don't like adding too many comments in code, to be honest. Of course, we don't have to only follow my preferences. We could maybe add:

```
in .env variables
```

to the description, if you think that would help.

mattwr18:
@soey thanks for the review. This is blocked by https://github.com/tactilenews/100eyes/pull/1837 though. Unfortunately, I missed your latest comment...

config/routes.rb
mattwr18: Ok, I've updated the constraints with the additional condition that onboarding must be allowed. I've also noticed that I missed that there were no constraints on the email onboarding paths, which is the reason we added the feature to disallow onboarding in the first place. I've added the constraint now.

See 4e87091a

I've updated the routing spec we have for WhatsApp. We don't have any other routing specs. I have seen a lot of posts discouraging the use of them, not sure your opinion. This also brought to my attention that we do not have any WhatsApp request specs for onboarding besides these limited routing specs. I've added an issue for it https://github.com/tactilenews/100eyes/issues/1848

I've also added request specs to check that the constraints are properly set up. See c825f9fe

soey (approved):

config/routes.rb
soey: Well just my opinion if I understand it correctly: I think you are already answering yourself to some extent in that comment. I think it's absolutely fine to test for that 404s if that is part of the application logic. Regarding the positive cases it would be better to have request specs instead of routing specs that only test for a 200 - but the later might be better than nothing. So I'm fine with a follow up

soey (approved):

mattwr18:
I will add a simple playbook to run the temporary rake task on all production instances at once.

soey:
LGTM ready to merge

PR #1827, opened 2024-04-03
Bump puma to v6.3.1 to fix critical security alert

mattwr18:
See https://github.com/tactilenews/100eyes/security/dependabot/65

There was a breaking change in Puma v6 that required upgrading Capybara together with this.

See [Welcome to Puma 6: Sunflower - Upgrade](https://github.com/puma/puma/blob/master/6.0-Upgrade.md#upgrade)

> We've removed the following public methods on Puma::Server: Puma::Server#min_threads, Puma::Server#max_threads. Instead, you can pass in configuration as an option to Puma::Server#new. This might make certain gems break (capybara for example).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ gem 'rails', '~> 6.1.6.1'
 # Use postgresql as the database for Active Record
 gem 'pg', '>= 0.18', '< 2.0'
 # Use Puma as the app server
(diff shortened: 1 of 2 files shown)


mattwr18:
@soey I'm going to go ahead and get this out since it was a critical security issue and it's passing the build and I've manually tested it on staging.100ey.es.

PR #1831, opened 2024-04-05
Bump view_component to fix moderate security alert

mattwr18:
See https://github.com/tactilenews/100eyes/security/dependabot/73

### DIFF
--- a/app/components/request_metrics/request_metrics.rb
+++ b/app/components/request_metrics/request_metrics.rb
@@ -2,10 +2,10 @@
 
 module RequestMetrics
   class RequestMetrics < ApplicationComponent
-    def initialize(request:, **)
+    def initialize(request_for_info:, **)
       super
 
-      @request = request
+      @request_for_info = request_for_info
     end
 
     def call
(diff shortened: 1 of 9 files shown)


app/components/request_metrics/request_metrics.rb
```diff
@@ -2,10 +2,10 @@
 
 module RequestMetrics
   class RequestMetrics < ApplicationComponent
-    def initialize(request:, **)
+    def initialize(request_for_info:, **)
```
mattwr18, line 5: These changes to rename our params from `request` to `request_for_info` became necessary with this upgrade to view_component because to fix the security alert, they added a check to see if the request(http) was of format html.

See https://github.com/ViewComponent/view_component/blob/v2.83.0/lib/view_component/base.rb#L318
> return text if request && !request.format.html?

Because of our unfortunate naming of our Request model, we are susceptible to naming conflicts. Since `request` is a reserved word in Rails, we should consider renaming our model.

soey (approved):

PR #1833, opened 2024-04-11
Fix specs failing when run in docker

mattwr18:
We mock the value of Setting.signal_server_phone_number, but were filtering out the value of ENV.fetch('SIGNAL_SERVER_PHONE_NUMBER', nil). Filtering out the env variables themselves is flaky because it depends on local .env variables, which is different locally, either with or without docker, and on different machines, like ci, or other teammates.

This PR proposes to filter out the value of
Setting.signal_server_phone_number, which would be the same as the request that is being received by vcr.


spec/vcr_setup.rb
```diff
@@ -10,7 +10,7 @@
   c.filter_sensitive_data('*100EYES') { ENV.fetch('THREEMARB_API_IDENTITY', nil) }
   c.filter_sensitive_data('THREEMARB_API_SECRET') { ENV.fetch('THREEMARB_API_SECRET', nil) }
   c.filter_sensitive_data('THREEMARB_PRIVATE') { ENV.fetch('THREEMARB_PRIVATE', nil) }
-  c.filter_sensitive_data('SIGNAL_SERVER_PHONE_NUMBER') { ENV.fetch('SIGNAL_SERVER_PHONE_NUMBER', nil) }
+  c.filter_sensitive_data('SIGNAL_SERVER_PHONE_NUMBER') { Setting.signal_server_phone_number }
```
soey, line 13: should we apply the same change to the other filters?

mattwr18: I think it would be nice for it to be consistent and I guess it would make sense to apply the change in the same PR. Alternatively, I went with another approach in[ Improve setting channels logic #1818](https://github.com/tactilenews/100eyes/pull/1818) that would also fix this problem and I think I actually prefer it.

I think we do not want to be loading values from our local .env files. Having a dedicated `.env.local.test` would allow us to make sure we are not loading our local .env files if they exist and would then mean we would be testing the same between our machines and ci.

soey: If you have good experience with that gem and see no issues, I would also go with it but see comment https://github.com/tactilenews/100eyes/pull/1818#pullrequestreview-2005405247

soey: said that I'm then fine to have this PR merged as it is.

soey (approved):

mattwr18:
@soey so having a closer look at this after our call, what I remember is that we agreed that we would update the vcr_setup to filter out the methods being called on Setting model for consistency and because that is typically what we mock in our test setups. Unfortunately though, two of the six are not methods in the Setting model:

```
THREEMARB_API_SECRET
THREEMARB_PRIVATE
```

are only in the env variables. So, we could either accept this inconsistency or solve this issue by using the dedicated .env.local.test file and leaving the vcr_setup as is. I think I prefer the latter. You?

soey:
See my proposal. I would fix what is reasonable here and stay with this inconsistency as that is already there when it comes to the implementation of the Setting fields itself.

mattwr18:
@soey ok, I don't agree as I think we should focus on the much more important issue that we having leaking `.env` variables in our tests. That should be the first priority. With that being fixed, we would not have the issue with Signal failing tests and we would not have the inconsistency in the vcr setup.

However, for some reason, it's important to you to have these tests passing first, so I'll merge.

soey:
@mattwr18 I agree with you that specific test env variable would be a better fix. I tried to get that fix running but I couldn't - see https://github.com/tactilenews/100eyes/pull/1837#issuecomment-2067117912

PR #1837, opened 2024-04-17
Favor dedicated .env for test env

mattwr18:
While working on https://github.com/tactilenews/100eyes/pull/1818, it became clear that the test  was actually loading values from a devs local .env environment. I believe this is less than ideal since we can have different .env values and some values I have configured are actually used to test that services work before deploying them.

@soey it turns out that the `Dotenv.overload` was not necessary because I named the file incorrectly. It should be `.env.test.local` and I had named it `.env.local.test`.

See [Load Order](https://github.com/bkeepers/dotenv#load-order)


soey:
have you double checked that the .env.test.local is really used. On my local machine it still fails if .env has a bad value set. Is that expected?

mattwr18:
> have you double checked that the .env.test.local is really used. On my local machine it still fails if .env has a bad value set. Is that expected?

I did check for one value that I knew to be set in my `.env` and unset in the `.env.test.local`, but there seems to be something off because there are many failing tests. I'll have to see what's happening.

mattwr18:
@soey the tests are passing now. What I did was merge in the latest changes to master. I think they were failing because I have signal configured in the `.env.test.local` file and it was set to the value that was previously in `.env.template`. With your changes to remove the guard clause comparing the server_phone_number to the Signal.server_phone_number and the change I added to mock the value of the signal_cli_rest_api_endpoint, they are all green now.

I also removed some superfluous mocks in [`06355b6`](https://github.com/tactilenews/100eyes/pull/1837/commits/06355b63f1c1ee937403567f49224370f9be6e50)

It might be better to go through all mocks in our tests and see if they are actually needed? Or when updating a test file in the future check if there are any superfluous mocks?

soey:
I would really love to see this working but it doesn't seem to work as expected. Tests are only working because of the other changes in main.

 I have set a .env with SIGNAL_SERVER_PHONE_NUMBER='+49123456789' and a env.test.local with SIGNAL_SERVER_PHONE_NUMBER='+4912345678'. 
 
 When I do `$ docker compose run -it app bash` and `$ RAILS_ENV=test rails c` then `ENV.fetch('SIGNAL_SERVER_PHONE_NUMBER', nil)` I get `"+4912345789"`. But the expected would be `+4912345678`

soey (approved):

PR #1847, opened 2024-05-02
refa: change the seed filling to be more realistic

soey:
-  to have a valid order of timestamps on requests and replies
-  to have correct associations

### DIFF
--- a/db/seeds/fill.rb
+++ b/db/seeds/fill.rb
@@ -6,10 +6,12 @@ require 'faker'
 contributors_count = 30
 request_count = 30
 replies_count = 30
+file_replies_count = 3
+photo_replies_count = 3
 
 users = User.all
 
-# images = 10.times.map { URI(Faker::Avatar.image(size: '50x50', format: 'png', set: 'set5')) }
+images = 10.times.map { URI(Faker::Avatar.image(size: '50x50', format: 'png', set: 'set5')) }
 
 FactoryBot.modify do
   factory :contributor do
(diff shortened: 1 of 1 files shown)


db/seeds/fill.rb
```diff
@@ -37,31 +39,21 @@
...
-  request.save!
-end
-
-Rails.logger.debug 'Seeding contributors..'
-contributors = FactoryBot.create_list(:contributor, contributors_count)
-
 FactoryBot.modify do
   factory :message do
-    created_at { message_time }
-    updated_at { message_time }
```
mattwr18, line 52: I think I understand what you mean by:

> to have a valid order of timestamps on requests and replies

but I don't think I agree that this achieves that. My understanding would be that this would simply change the `created_at` and `updated_at` attrs from 14.days.ago to Time.current. How does that achieve a "valid order"? What am I missing?

I think what might be better is to use something like:

```rb
Faker::Time.between_dates(from:  Date.today - 14, to: Date.today, period: :all)
```

As I'm writing this, I realize that maybe you are trying to address the issue that the request itself was not created 14 days ago, so how could the messages come in/or go out before the request?

If that's the case, I still think it would be an improvement to set the `created_at` of the request to some time in the past and then set the replies to come in sometime between that date and today.

db/seeds/fill.rb
```diff
@@ -37,31 +39,21 @@
...
-Rails.logger.debug 'Seeding messages..'
-FactoryBot.build_list(:message, message_count) do |message|
-  message.class.skip_callback(:create, :after, :send_if_outbound, raise: false)
-  message.save!
+Rails.logger.debug 'Seeding requests..'
+FactoryBot.build_list(:request, request_count) do |request|
+  Message.skip_callback(:commit, :after, :send_if_outbound, raise: false)
+  request.save!
+  Rails.logger.debug 'Seeding requests replies...'
+  FactoryBot.create_list(:message, replies_count, request: request)
```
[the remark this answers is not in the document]
mattwr18: I can of course also test this locally, I've held off on approving this since this would make a big impact for me as I have actually production values in my `.env` setup that could cost us money if the seeds were run and the callbacks were not properly skipped.

db/seeds/fill.rb
soey: The previous code just created a bunch of requests. After all requests have been created the messages are created on a batch. So if you look at conversations view you will find all requests at the top only then a bunch of contributor messages. To fix this you have to create a requests, send out the messages and make the contributor respond to them before sending the next request. Only then the order of message outbound/inbound is correct.

mattwr18 (approved):
Moves in the right direction! We can continue to improve it as we work.

PR #1852, opened 2024-05-06
Fix broken scroll to message after adding turbo-frame

### LINKED ISSUE #1851: [BUG] Scroll message into view broken
### Steps to reproduce

* Click on a link Zur Antwort/Zur Nachricht in the ActivityNotifications on `/dashboard`
* Request show page is lazy loaded with the skeleton component first and does not scroll to the message as it should with an anchor in the url

## Dev Notes

This was broken when we added turbo frames to improve performance on the slow loading request show page.
You can also go to the `/community` page, choose a contributor who has been active and click on one of the links in `Letzte Antworten`

mattwr18:
Fixes #1851

### DIFF
--- a/app/components/message_groups/message_groups.js
+++ b/app/components/message_groups/message_groups.js
@@ -1,0 +1,20 @@
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
+  connect() {
+    document.addEventListener('turbo:frame-render', event => {
+      const anchor = new URL(document.location).hash;
+      if (!anchor) return;
+
+      const turboFrame = event.target;
+      const elementToScrollTo = turboFrame.querySelector(anchor);
+      if (!elementToScrollTo) return;
+
+      elementToScrollTo.scrollIntoView({ behavior: 'smooth', block: 'center' });
+      Object.assign(elementToScrollTo.style, {
+        borderColor: '#007eff',
+        boxShadow: '0 0 0 3px #7ac8ff',
+      });
+    });
+  }
+}
--- a/app/components/message_groups/message_groups.js
+++ b/app/components/message_groups/message_groups.js
@@ -11,6 +11,9 @@ export default class extends Controller {
       if (!elementToScrollTo) return;
 
       elementToScrollTo.scrollIntoView({ behavior: 'smooth', block: 'center' });
+      // This was added to mimic the styles in ChatMessage:target
+      // See https://developer.mozilla.org/en-US/docs/Web/CSS/:target
+      // and https://github.com/tactilenews/100eyes/pull/1852#discussion_r1590751639
       Object.assign(elementToScrollTo.style, {
         borderColor: '#007eff',
         boxShadow: '0 0 0 3px #7ac8ff',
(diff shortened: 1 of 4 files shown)


app/components/message_groups/message_groups.js
```diff
@@ -0,0 +1,20 @@
...
+
+      const turboFrame = event.target;
+      const elementToScrollTo = turboFrame.querySelector(anchor);
+      if (!elementToScrollTo) return;
+
+      elementToScrollTo.scrollIntoView({ behavior: 'smooth', block: 'center' });
+      Object.assign(elementToScrollTo.style, {
+        borderColor: '#007eff',
+        boxShadow: '0 0 0 3px #7ac8ff',
+      });
```
mattwr18, line 17: This was added to mimic the styles in `ChatMessage:target`. When relying the native browser functionality, the `:target` pseudo class applies the styles to the element being targeting in the url's anchor. When using js, this does not work, so we need to apply the styles with js.

See https://developer.mozilla.org/en-US/docs/Web/CSS/:target

Maybe we should remove the CSS since it would never used(?) Maybe, for development, by toggling the `:target` pseudo class in the browser tools.

app/components/message_groups/message_groups.js
```diff
@@ -0,0 +1,20 @@
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
+  connect() {
+    document.addEventListener('turbo:frame-render', event => {
```
mattwr18, line 5: See https://turbo.hotwired.dev/reference/events

app/components/requests_feed/requests_feed.html.erb
```diff
@@ -19,7 +19,7 @@
         <li>
           <a
             class="Link"
-            href="<%= request_path(reply.request, anchor: "contributor-#{contributor.id}") %>"
+            href="<%= request_path(reply.request, anchor: "message-#{reply.id}") %>"
```
mattwr18, line 22: This was updated for consistency and because this PR would change the way it operated before this bug was introduced in [Optimize requests show page #1791](https://github.com/tactilenews/100eyes/pull/1791) .

There are currently no styles to highlight the ChatMessageGroup with like we do for MessageGroup and the RequestRow.

With these changes, there would be a blue border around the entire ChatMessageGroup instead of just the latest message in that group.

app/components/message_groups/message_groups.js
soey: In this case we really should add an appropriate comment into the code to explain exactly that. 

Side note: Maybe it's good practice to put inline code whenever an explanatory comments seems needed in the PR. Same as if one feels the need to put an inline comment it's better to refactor the code ..

soey (changes requested):
I would be happy about a comment. Please cross reference to the JS section where you put the styles from the CSS. Omitting the CSS would be fine if it is not a component to be used somewhere else (outside Turbo) - as it is a component that should be reusable, I would keep the CSS. Besides that good fix - thnx. Still feels **ouch** that we are in the need for that ..

mattwr18:
> I would be happy about a comment. Please cross reference to the JS section where you put the styles from the CSS. Omitting the CSS would be fine if it is not a component to be used somewhere else (outside Turbo) - as it is a component that should be reusable, I would keep the CSS. Besides that good fix - thnx. Still feels **ouch** that we are in the need for that ..

I added the comment in 0908732f

PR #1862, opened 2024-05-14
Add basic stats with breakdown of channels counts

### LINKED ISSUE #1861: [Feature] Add basic stats page for admin
```
As an admin of 100eyes
I would like to have more information on which channels are being used
So that I can focus on improving UX for those channels
```

mattwr18:

Added a tab in the admin page for stats. This page currently simply has the count of contributors by channel.
It does not filter out for active contributors or anything else currently.

### Whats not included in this PR and why

Currently, it seems pretty limited what we can easily achieve with administrate gem. Our css styles do not get loaded or applied and we cannot connect to our stimulus controllers without overriding administrate's js files, it seems. Have not tried. It might be nice to have a pie chart that displays this info.

Maybe, we could display info on our dashboard page? Is it only interesting for admin? If we wanted to scope it to admin only, we could easily add a flag to display it only to admin.

Closes #1861

### DIFF
--- a/app/components/admin_stats/admin_stats.html.erb
+++ b/app/components/admin_stats/admin_stats.html.erb
@@ -1,0 +1,8 @@
+<section style="padding: 20px">
+  <h1 style="padding-bottom: 20px">Stats</h1>
+  <p><strong>Email contributors:</strong> <%= email_contributors_count %></p>
+  <p><strong>Threema contributors:</strong> <%= threema_contributors_count %></p>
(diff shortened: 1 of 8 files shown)


soey:
lgtm - but I don't get what app/dashboards/stat_dashboard.rb is for

soey:
lgtm - but I don't get what app/dashboards/stat_dashboard.rb is for

roschaefer (approved):
count

Yes, this looks like a mergeable PR!

PR #1877, opened 2024-05-22
[refactor] specs to use skip validations trait

mattwr18:

We recently added a new trait in the contributors factory to skip validations in https://github.com/tactilenews/100eyes/pull/1862. This is useful as we do this to bypass our ThreemaValidator ensuring that we do not create/update threema_id to invalid values. This PR refactors to make use of this trait to DRY out the code.

### DIFF
--- a/spec/adapters/threema_adapter/inbound_spec.rb
+++ b/spec/adapters/threema_adapter/inbound_spec.rb
@@ -4,7 +4,7 @@ require 'rails_helper'
 
 RSpec.describe ThreemaAdapter::Inbound do
   let(:threema_id) { 'V5EA564T' }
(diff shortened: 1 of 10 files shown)


roschaefer (approved):
factory

I will give this my approval although I have mixed feelings about it. My biggest concern is that it is so easy to overlook `:skip_validations`.

But I can't find of a better solution. Somehow, I believe our issues with the factories here show us that at least the HTTP call in `ThreemaValidator` needs to be disabled for testing. But this would bring `production` and `test` environment further away from each other.

I think the PR as of now is OK to get merged, I will try to debug our CI server and then click the green button.

roschaefer:
Hm, looks as if we have flaky tests.

PR #1881, opened 2024-05-27
refactor: avoid null-byte characters

roschaefer:
Motivation
----------
This is a better way of avoiding null-byte characters when creating fake Threema IDs for testing.
See this commit: https://github.com/tactilenews/100eyes/pull/1878/commits/f8fb2d7d66e500cff7583e3a3fc8cacc60794793

How to test
-----------
1. `bin/rspec`
2. Tests pass (hopefully more reliably)


mattwr18 (approved):
Even better!!

PR #1887, opened 2024-05-29
Lets see how it works

mattwr18:


mattwr18:
@roschaefer as expected, we got the error in the cd while it was on a lower version, I verified that setting the version allows it to pass cd... we should have a follow up PR that removes the version at some point.

PR #1889, opened 2024-06-01
refactor: ansible backup for managed cluster

roschaefer:
Motivation
----------
Almost all our deployments run on a managed cluster, as a matter of fact, we want to use managed clusters only in the future.

The previous ansible configuration used the obsolete `db` docker container only to connect to the managed postgres cluster, which is very complicated.

This will remove the unnececcessary `docker exec ...` and also DRYs some of the following configuration.

How to test
-----------
1. Follow the README on any production instance.
2. You should have a local backup in `ansible/backups/`

### DIFF
--- a/ansible/roles/backup/tasks/main.yml
+++ b/ansible/roles/backup/tasks/main.yml
@@ -3,42 +3,22 @@
   debug:
     msg: "Backing up to folder: {{ backup_directory }}"
 
-- name: Query id of database container
-  command:
-    cmd: docker compose -f docker-compose.yml -f docker-compose.prod.yml ps -q db
-    chdir: /home/ansible
-  register: get_container_id
-  tags:
-    - work-in-progress
+- name: Install pg_dump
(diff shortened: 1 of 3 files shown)


mattwr18 (approved):
Great! Tested this yesterday for all production instances.
We ran into an issue with running these on instances running on Ubuntu 20.04 as `postgresql-client-14` was not available to be installed, but I was able to get around this by adding the repo to the instance with:

```
sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
```
Then, I imported the repository signing key, with:

```
wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
```

I updated the package list:

```
sudo apt-get update
```

then, I was able to successfully run this playbook and take a backup.

PR #1890, opened 2024-06-01
build: configure staging to use managed db cluster

roschaefer:
Motivation
----------
To have more consistency between our deployments, we use managed db clusters for all instances. This will update our staging deployment.

How to test
-----------
1. Merge it in
2. CD pipeline still works


roschaefer:
We want to manually remove the dangling `db` container after this.

PR #1891, opened 2024-06-01
Remove superfluous db service from production

mattwr18:

We are using managed dbs and do not need a db service in production. It has been moved to the override to use for local development.


roschaefer (approved):
volume

Let's avoid failed deployments because of dangling volumes and postgres containers

PR #1895, opened 2024-06-04
Draft solution for multi-tenancy

mattwr18:


app/models/setting.rb
```diff
@@ -40,46 +40,21 @@ def self.whats_app_configured?
     twilio_configured? || three_sixty_dialog_configured?
   end
 
-  field :project_name, default: ENV['HUNDRED_EYES_PROJECT_NAME'] || '100eyes'
   field :application_host, readonly: true, default: ENV['APPLICATION_HOSTNAME'] || 'localhost:3000'
 
   field :git_commit_sha, readonly: true, default: ENV.fetch('GIT_COMMIT_SHA', nil)
   field :git_commit_date, readonly: true, default: ENV.fetch('GIT_COMMIT_DATE', nil)
 
-  field :onboarding_logo_blob_id, type: :integer
```
soey, line 49: To be discussed: Should we keep global fallbacks?

mattwr18:
This is outdated, was a spike, and has been implemented in a different way so it is outdated.

PR #1911, opened 2024-06-24
Fix signal onboarding

### LINKED ISSUE #1908: [BUG] Signal onboarding broken
### Steps to reproduce

* Onboard with Signal to an instance you have not onboarded with yet. Or you can simply delete your contributor locally and run `signal-cli --config /home/.local/share/signal-cli removeContact <SIGNAL_PHONE_NUMBER> --forget` and onboard again
* As part of the process, you will be requested to send a hello to the signal_server_phone_number registered to the instance you are attempting to onboard with.
* This results in an error:

```
SignalAdapter::UnknownContributorError

Received a message on signal from an unknown sender: <UUID>
```

this is due to Signal not including the phone number in the first incoming message any longer. See https://github.com/AsamK/signal-cli/discussions/1414

### Dev Notes

Looking into this, we seem to get the phone number if we send the welcome message first, but we also seem to get hit by rate limit challenges that need to be manually solved when sending a message to someone who has not sent to us first.  These errors happen on the first message even for contributors who have gone through the normal onboarding flow and have sent us a message first. It's unclear if they would happen more often if we start sending out messages before receiving them, or not.

We also can send out messages without the phone number, using the contributor's username and I did not experience any errors for rate limiting doing this. Usernames are not visible on your profile and Signal advises to only share your username with people you trust, so it's possible that Signal does not see messages using someone's username as in need of as much spam protection.

When we send a message out to someone using the username instead of phone number, we automatically have a contact created with their uuid and username, we also have their identity added, which can be fetched with:

```
signal-cli --config /home/.local/share/signal-cli listIdentities
```

There is no need to run the `SignalAdapter::CreateContactJob`.

However, we still only receive their uuid with an incoming message, we do not receive their username, so we still need to find a way to associate the uuid with the contributor.

There is currently no endpoint for `listContacts` in **signal-cli-rest-api**.  There is an issue https://github.com/bbernhard/signal-cli-rest-api/issues/238 that the maintainer mentions that it could be added, but it is not possible currently.

We could manually complete the onboarding by running the command in **signa-cli** directly; however, this is obviously less than ideal.

We could also send them an onboarding token, similar to how we do with telegram when we can't automatically complete their onboarding. Then, we could look up the contributor by their onboarding token and save their uuid for future messages.

It's not clear how many Signal users already have a username, or how easy it would be to instruct them on how to create one.
See https://signal.org/blog/phone-number-privacy-usernames/

mattwr18:
Fixes #1908

We decided to update our Signal onboarding to create a signal_onboarding_token similar to what we do with Telegram. We create the contributor in our database and instruct them to send us a message containing only the token, by which we can identify them. Once we identify them by their token, we can save their uuid and complete their onboarding.

We then send them messages with their uuid and identify them with incoming message by it as well.


app/adapters/signal_adapter/inbound.rb
```diff
@@ -54,25 +54,37 @@ def trigger(event, *args)
...
 
       unless sender
-        trigger(UNKNOWN_CONTRIBUTOR, signal_phone_number)
+        trigger(UNKNOWN_CONTRIBUTOR, signal_message.dig(:envelope, :source))
         return nil
       end
 
       if sender.signal_onboarding_completed_at.blank?
-        trigger(CONNECT, sender)
+        trigger(CONNECT, sender, signal_uuid)
```
mattwr18, line 77: Maybe we should refactor this into two methods, like:

```rb
def initialize_contributing_sender; end
def initialize_new_sender; end
```

docker-compose.yml
```diff
@@ -23,7 +23,7 @@ services:
     depends_on: [ signal ]
 
   signal:
-    image: bbernhard/signal-cli-rest-api:0.81
+    image: bbernhard/signal-cli-rest-api:0.162-dev
```
mattwr18, line 26: There is an issue with sending messages by uuid, which has been fixed, but not yet released. See https://github.com/bbernhard/signal-cli-rest-api/issues/550

mattwr18:
@roschaefer this is failing the build, I need to revert some changes to tests that I thought needed to be changed and I also need to write tests for the new onboarding/sending/receiving flow.
You could test it out though as I think it works as I imagine it should.

config/locales/de.yml
```diff
@@ -84,16 +84,14 @@ de:
...
       heading: Fast geschafft.
-      text: Damit wir Ihnen zukünftig auf Signal schreiben können, müssen Sie uns zuerst kontaktieren.
-      action: Senden Sie uns hierzu eine Nachricht auf Signal. Ein kurzes „Hallo” reicht aus! Unsere Nummer ist <strong>%{number}</strong>.
-      link: Zur Signal-App
-      copy: Handynummer kopieren
-      help:  Über den Button gelangen Sie direkt zu unserem Kontakt in der Signal-App. Alternativ können Sie unsere Nummer kopieren.
+      text: Verbinden Sie Ihr Signal-Konto in der Signal-App, um die Anmeldung abzuschließen.
+      steps:
+        - "Finden Sie uns auf Signal. Suchen Sie dazu nach <strong>%{signal_server_phone_number}</strong>."
+        - "Wenn Sie unser Signal-Konto gefunden haben, geben Sie diesen Code ein, sobald sie danach gefragt werden: <strong>%{token}</strong>"
```
roschaefer, line 90:
```suggestion
        - "Wenn Sie unser Signal-Konto gefunden haben, geben Sie diesen Code ein, sobald Sie danach gefragt werden: <strong>%{token}</strong>"
```

Very formal German..

roschaefer (approved):
onboarding

Looks good to me and I can't find anything problematic code-wise.

app/jobs/signal_adapter/attach_contributors_avatar_job.rb
```diff
@@ -4,8 +4,11 @@ module SignalAdapter
   class AttachContributorsAvatarJob < ApplicationJob
     queue_as :attach_signal_avatar
 
-    def perform(contributor)
-      avatar = "/app/signal-cli-config/avatars/profile-#{contributor.signal_phone_number}"
+    def perform(contributor_id:)
+      contributor = Contributor.find_by(id: contributor_id)
+      return unless contributor
+
+      avatar = "/app/signal-cli-config/avatars/profile-#{contributor.signal_uuid}"
```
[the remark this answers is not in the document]
mattwr18: I think so because we only ever call this job when we connect someone and now we only have the option of using a person's uuid. After we sunset the current instances, we can completely rip out the `signal_phone_number` as it won't be used any longer.

app/models/contributor.rb
```diff
@@ -54,7 +54,7 @@ class Contributor < ApplicationRecord
   scope :with_email, -> { where.not(email: nil) }
   scope :with_threema, -> { where.not(threema_id: nil) }
   scope :with_telegram, -> { where.not(telegram_id: nil) }
-  scope :with_signal, -> { where.not(signal_phone_number: nil, signal_onboarding_completed_at: nil) }
```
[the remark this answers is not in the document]
mattwr18: I think this change was unintentional... I must go through more carefully and add back the `signal_onboarding_completed_at` that I removed when I thought we would go a different route.

db/migrate/20240618054327_update_signal_attrs_on_contributors.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class UpdateSignalAttrsOnContributors < ActiveRecord::Migration[6.1]
+  def change
+    change_table :contributors, bulk: true do |t|
+      t.column :signal_uuid, :string, default: nil
+      t.column :signal_onboarding_token, :string, unique: true
+    end
+  end
```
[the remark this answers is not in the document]
mattwr18: Yes, I think, cause we need them to send us a message first containing only the `signal_onboarding_token` to be able to connect them, ie add their `signal_uuid`, which is necessary to identify them in future inbound messages.

config/locales/de.yml
mattwr18: Yes, I remember that we wanted informal. I copy/pasted this from the `onboarding_telegram_fallback` component's steps... Could you suggest the informal version?

app/components/contributor_signal_settings/contributor_signal_settings.rb
```diff
@@ -11,5 +11,25 @@ def initialize(contributor:, **)
...
+      if contributor.signal_phone_number.present?
+        t('.complete.text.phone_number',
+          name: contributor.name,
+          phone_number: contributor.signal_phone_number&.phony_formatted)
+      else
+
+        t('.complete.text.uuid',
+          name: contributor.name,
+          uuid: contributor.signal_uuid)
+      end
```
mattwr18, line 25: This is maybe weird @roschaefer ... I'm not sure that displaying the UUID for the clients is interesting for them

PR #1919, opened 2024-06-29
refactor: local restore

roschaefer:
Motivation
----------
I'm trying to write a regression test for #1914. For that I need to setup signal locally. I do it with restoring a backup from one of our demo instances.

It turns out that `signal-cli-config` was not restored. Second, I removed obsolete `docker-compose.backup.yml` and unnecessary `tmp/db` folders by just using `docker compose cp`, plus I DRYed the ansible playbook.

How to test
-----------
1. `ansible-playbook ansible/restore_locally.yml`
2. Runs faster, no temporary folders created.

### DIFF
--- a/ansible/roles/local_restore/tasks/restore_backup.yml
+++ b/ansible/roles/local_restore/tasks/restore_backup.yml
@@ -1,23 +1,19 @@
 ---
-- name: Create tmp/db/ folder for sharing files between host and container
+- name: Wipe out target locations of backup folders
   file:
-    dest: "{{ playbook_dir }}/../tmp/db"
-    state: directory
-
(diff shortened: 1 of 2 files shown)


mattwr18 (commented):
I'm not sure I actually like these changes @roschaefer ... I suggested making similar changes when we introduced the backup of the signal-cli-config and Till pointed out the potential danger in doing so. Have a look.

https://github.com/tactilenews/100eyes/pull/1380

I have a signal-cli-config for development. Maybe a better solution is to save a zip of this to 1password that you could use whenever you want to test out signal stuff? I also like the idea of not having actual contributors from a production instance in your db while you are testing out Signal since you could easily send out a message to everyone without intending on it.

PR #1920, opened 2024-06-29
test: regression test for #1914

roschaefer:
Motivation
----------
PR #1914 did not regression test the bug, so we're doing it here.

Note that we ran into non-reproducible errors in the test setup with `vcr`. @mattwr18 and I could only receive the signal messages only once during our pair-programming and then nothing anymore, no matter what I tried on my Signal app on my phone or desktop.

We captured the three received messages from the shell history and put it manually into the `vcr` cassette.

How to test
-----------
1. `bin/rspec spec/jobs/signal_adapter/receive_polling_job_spec.rb:159`
2. Test passes
3. `git revert b254b9a62fe980742a2db72762414180469dac17`
4. `bin/rspec spec/jobs/signal_adapter/receive_polling_job_spec.rb:159`
5. Test fails

### DIFF
--- a/spec/jobs/signal_adapter/receive_polling_job_spec.rb
+++ b/spec/jobs/signal_adapter/receive_polling_job_spec.rb
@@ -155,6 +155,17 @@ RSpec.describe SignalAdapter::ReceivePollingJob, type: :job do
           end
         end
       end
+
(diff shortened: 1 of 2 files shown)


mattwr18 (approved):
Great @roschaefer thanks for adding this test. It should have been added when I added the delivery receipts, but better late than never :)

PR #1926, opened 2024-07-03
Remove unused Chat component

### LINKED ISSUE #1897: [chore] Remove superfluous Chat component
After merging in the conversations feature, we do not expose a direct way to reach the old Chat component. However, we do redirect the user there after creating a manual message. We need to clean up, remove all links/redirects to the old UI and point to the new UI and remove the component.

mattwr18:
Closes #1897

clean up from PR #1849


roschaefer (approved):
clean

PR #1928, opened 2024-07-15
Favor telegram webhook method in development

mattwr18:

Telegram runs in bot poller mode in development by default. This can be convenient, but has some limitations. It requires having either the bot key configured in env variables or have a default bot set. Moving to introduce multi-tenancy means that we will not longer have a default bot configured. We could set `BOT=1` in our .env variables, but this has some downsides as well, most importantly it will not work for testing multiple bots. One can manually run the `Telegram::Bot::UpdatesPoller .fetch_updates`, but this requires a manual step and more importantly is not how Telegram works in production.

Documentation has been added for setting the webhook, which will work for multiple bots as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,17 @@ We're using [our own fork](https://github.com/tactilenews/threema) of the `three
 bundle config local.threema .
 ```
 
+#### Telegram
+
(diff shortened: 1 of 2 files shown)


roschaefer (approved):
remove

Yes! I was about to do this as well.

Issue #1896, opened 2024-06-06
[improvement] Improve conversations create message flow


mattwr18:
```
As a user of the conversations flow
I would like to scroll have a better UX when creating a new message in the conversation
So that I don't need a full refresh and scroll all the way to the bottom to create a new message
```

soey:
As the user requested we will flip the order of messages to have the latest on top and add the input field on top as well.
With that the user don't have to scroll down anymore to get to the latest conversation and can directly respond from the top of the screen.

PR #1930, opened 2024-07-17
Improvement #1896: reverse conversations order and put form on top

soey:
Reverses the order of the conversations on the new introduced view and puts the message box on the top to improve usability as being requested by the users


spec/models/contributor_spec.rb
```diff
@@ -359,11 +359,11 @@
       expect(contributor.conversations).to include(sent_message)
     end
 
-    it 'sorts the messages so that the oldest first' do
+    it 'sorts the messages so that the newest first' do
```
roschaefer, line 362:
```suggestion
    it 'sorts the messages so that the newest comes first' do
```

roschaefer (approved):
sure

Feel free to merge!

PR #1939, opened 2024-07-22
translations(de): "leeren" sounds really odd

roschaefer:
image


soey:
this is perfectly fine when the specs are green

PR #1943, opened 2024-07-23
Report ActiveRecord::RecordNotFound errors to Sentry

### LINKED ISSUE #1942: [chore] Remove ActiveRecord::RecordNotFoun error from sentry excluded exceptions
I just realized that we do not report `ActiveRecord::RecordNotFound` errors because they are by default excluded from being reported in sentry-ruby.

```
User excluded error: #<ActiveRecord::RecordNotFound: Couldn't find Contributor with 'id'=111 [WHERE "contributors"."organization_id" = $1]>
```

We don't want to swallow these errors because it can mean we silently are not delivering messages to contributors. We want to report these issues so we can look into the cause of them.

mattwr18:
these errors are excluded by default with sentry-ruby

Fixes #1942

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -4,5 +4,6 @@ Sentry.init do |config|
   config.breadcrumbs_logger = [:active_support_logger]
   config.capture_exception_frame_locals = true
   config.server_name = ENV['APPLICATION_HOSTNAME'] if ENV['APPLICATION_HOSTNAME'].present?
+  config.excluded_exceptions -= ['ActiveRecord::RecordNotFound']
   config.excluded_exceptions += ['ErrorNotifier::IgnoredError']
 end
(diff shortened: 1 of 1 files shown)


config/initializers/sentry.rb
```diff
@@ -4,5 +4,6 @@
   config.breadcrumbs_logger = [:active_support_logger]
   config.capture_exception_frame_locals = true
   config.server_name = ENV['APPLICATION_HOSTNAME'] if ENV['APPLICATION_HOSTNAME'].present?
+  config.excluded_exceptions -= ['ActiveRecord::RecordNotFound']
   config.excluded_exceptions += ['ErrorNotifier::IgnoredError']
```
roschaefer, line 8: I find it super hard to see that we *remove* something here.

```suggestion
  unignored_exceptions = ['ActiveRecord::RecordNotFound']
  ignored_exceptions = ['ErrorNotifier::IgnoredError']

  config.excluded_exceptions -= unignored_exceptions
  config.excluded_exceptions += ignored_exceptions
```

roschaefer (changes requested):
exception

PR #1944, opened 2024-07-23
Avoid infinite loop in postmark

### LINKED ISSUE #1941: [BUG] If Postmark::InactiveRecipientError is called with an admin's email it gets stuck in a loop
### Steps to reproduce

* Probably best not to, but if you try to send a message to an email that has unsubscribed, it will try to send out the contributor marked inactive email, which will trigger the error and get caught in an infinite loop.

We should check to see that the email we are sending a notification about is not the email that unsubscribed.

mattwr18:
if we try to send a message to an admin notifying them that a contributor is inactive, but that admin has the same email as the contributor who has unsubscribed, it gets stuck in an infinite loop

Fixes #1941

### DIFF
--- a/spec/adapters/postmark_adapter/outbound_spec.rb
+++ b/spec/adapters/postmark_adapter/outbound_spec.rb
@@ -310,7 +310,7 @@ RSpec.describe PostmarkAdapter::Outbound, type: :mailer do
         end
       end
 
-      context 'user, not admin' do
+      context 'user without an admin role' do
         let(:admin) { build(:user) }
         let(:contributor) { create(:contributor) }
 
@@ -319,18 +319,18 @@ RSpec.describe PostmarkAdapter::Outbound, type: :mailer do
         end
       end
 
-      context 'admin email equals contributo marked inactive email' do
-        let(:admin) { build(:user, id: 231_123, admin: true, email: 'my-email@example.org') }
-        let(:contributor) { build(:contributor, id: 231_123, email: 'my-email@example.org') }
+      context 'admin email equals contributor email' do
+        let(:admin) { create(:user, admin: true, email: 'my-email@example.org') }
+        let(:contributor) { create(:contributor, email: 'my-email@example.org') }
 
         it 'does not enqueue a Mailer' do
           expect { subject }.not_to have_enqueued_job
         end
       end
 
-      context 'with an admin and contributor' do
-        let(:admin) { create(:user, admin: true) }
-        let(:contributor) { create(:contributor) }
+      context 'with an admin and contributor without the same email address' do
+        let(:admin) { create(:user, admin: true, email: 'admin@example.org') }
+        let(:contributor) { create(:contributor, email: 'contributor@example.org') }
 
         it 'enqueues a Mailer' do
           expect { subject }.to have_enqueued_job.on_queue('default').with(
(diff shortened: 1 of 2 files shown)


spec/adapters/postmark_adapter/outbound_spec.rb
```diff
@@ -288,5 +288,62 @@
...
+      context 'user, not admin' do
+        let(:admin) { build(:user) }
+        let(:contributor) { create(:contributor) }
+
+        it 'does not enqueue a Mailer' do
+          expect { subject }.not_to have_enqueued_job
+        end
+      end
+
+      context 'admin email equals contributo marked inactive email' do
```
roschaefer, line 322:
```suggestion
      context 'admin email equals contributor email' do
```

spec/adapters/postmark_adapter/outbound_spec.rb
```diff
@@ -288,5 +288,62 @@
...
+      context 'no contributor' do
+        let(:admin) { build(:user, admin: true) }
+        let(:contributor) { nil }
+
+        it 'does not enqueue a Mailer' do
+          expect { subject }.not_to have_enqueued_job
+        end
+      end
+
+      context 'user, not admin' do
```
roschaefer, line 313:
```suggestion
      context 'user has no admin role' do
```

spec/adapters/postmark_adapter/outbound_spec.rb
```diff
@@ -288,5 +288,62 @@
...
+        let(:contributor) { create(:contributor) }
+
+        it 'does not enqueue a Mailer' do
+          expect { subject }.not_to have_enqueued_job
+        end
+      end
+
+      context 'admin email equals contributo marked inactive email' do
+        let(:admin) { build(:user, id: 231_123, admin: true, email: 'my-email@example.org') }
+        let(:contributor) { build(:contributor, id: 231_123, email: 'my-email@example.org') }
```
roschaefer, line 324:
```suggestion
        let(:admin) { build(:user, admin: true, email: 'my-email@example.org') }
        let(:contributor) { build(:contributor, email: 'my-email@example.org') }
```

Why the hard-coded id? I would expect it's not relevant? Users and contributors are two different models, or did I miss something?

spec/adapters/postmark_adapter/outbound_spec.rb
```diff
@@ -288,5 +288,62 @@
...
+
+        it 'does not enqueue a Mailer' do
+          expect { subject }.not_to have_enqueued_job
+        end
+      end
+
+      context 'with an admin and contributor' do
```
roschaefer, line 331:
```suggestion
      context 'with an admin and contributor who do not have the same email address' do
```

roschaefer (changes requested):
I love the bug that we're fixing here. It's so silly.

full circle

Nevertheless I requested some changes. Everything is just a wording issue except the primary keys on line 323 and 324. Without this, I would have hit approve.

spec/adapters/postmark_adapter/outbound_spec.rb
mattwr18: It was just to avoid creating the records. If we just build them, no ids are issued and that results in a serialization error.

roschaefer: Ok, what about `create` instead of `build`?

I was just confused, I thought it has something to do with the feature itself.

PR #1945, opened 2024-07-26
Scope additional resources at the organization level

mattwr18:


app/activity_notifications/chat_message_sent.rb
```diff
@@ -3,15 +3,16 @@
 class ChatMessageSent < Noticed::Base
   deliver_by :database, format: :to_database, association: :notifications_as_recipient
 
-  param :contributor_id, :request_id, :user_id, :message_id
+  param :contributor_id, :request_id, :user_id, :message_id, :organization_id
```
soey, line 6: From a db normalization perspective we are denormalizing  the organization_id, if a request always belongs to one organization. This might already be the same for user_id and contributor_id as they might be deducted from the message itself.

app/activity_notifications/message_received.rb
```diff
@@ -3,14 +3,15 @@
 class MessageReceived < Noticed::Base
   deliver_by :database, format: :to_database, association: :notifications_as_recipient
 
-  param :contributor_id, :request_id, :message_id
+  param :contributor_id, :request_id, :message_id, :organization_id
```
soey, line 6: same here.

app/activity_notifications/request_scheduled.rb
```diff
@@ -3,12 +3,13 @@
 class RequestScheduled < Noticed::Base
   deliver_by :database, format: :to_database, association: :notifications_as_recipient
 
-  param :request_id
+  param :request_id, :organization_id
```
soey, line 6: same here.

app/components/request_form/request_form.rb
```diff
@@ -2,20 +2,17 @@
 
 module RequestForm
   class RequestForm < ApplicationComponent
-    def initialize(organization:, request:)
+    def initialize(organization:, request:, available_tags:)
```
soey, line 5: Well, it has already been here. Why aren't we deducting the organization from the request itself?

app/components/request_form/request_form.rb
```diff
@@ -2,20 +2,17 @@
...
     end
 
     private
 
-    attr_reader :organization, :request
-
-    def available_tags
-      Contributor.all_tags_with_count.to_json
-    end
+    attr_reader :organization, :request, :available_tags
```
soey, line 15: organization could be just delegated to request

app/models/message.rb
```diff
@@ -5,7 +5,8 @@ class Message < ApplicationRecord
 
   default_scope { order(created_at: :desc) }
 
-  multisearchable against: :text, if: :reply?
+  multisearchable against: :text, if: :reply?,
+                  additional_attributes: ->(message) { { organization_id: message.request.organization_id } }
```
soey, line 10: should we delegate "organization" from "message" to "request" as a message always belongs to one request?

app/models/request.rb
```diff
@@ -69,7 +71,7 @@ def messages_by_contributor
   def self.broadcast!(request)
     if request.planned?
       BroadcastRequestJob.delay(run_at: request.schedule_send_for).perform_later(request.id)
-      RequestScheduled.with(request_id: request.id).deliver_later(User.all)
+      RequestScheduled.with(request_id: request.id, organization_id: request.organization.id).deliver_later(User.all)
```
soey, line 74: Why have organization_id as its own parameter if we can deduct it from the request itself? Doesn't that add a possible source of errors?

app/views/requests/edit.html.erb
```diff
@@ -6,6 +6,6 @@
   <% end %>
 
   <%= c 'section' do %>
-    <%= c 'request_form', organization: @organization, request: @request %>
+    <%= c 'request_form', organization: @organization, request: @request, available_tags: @available_tags %>
```
soey, line 9: Same here - and it was already like this. Shouldn't we just use request.organization internally and get rid of the parameter organization?

app/views/requests/new.html.erb
```diff
@@ -7,6 +7,6 @@
   <% end %>
 
   <%= c 'section' do %>
-    <%= c 'request_form', organization: @organization, request: @request %>
+    <%= c 'request_form', organization: @organization, request: @request, available_tags: @available_tags %>
```
soey, line 10: Same here.

app/controllers/search_controller.rb
```diff
@@ -4,6 +4,8 @@ class SearchController < ApplicationController
   def index
     @results = []
     query = params[:q]
-    @results = PgSearch.multisearch(query).map(&:searchable) if query
+    return unless query
+
+    @results = PgSearch.multisearch(query).where(organization_id: @organization.id).map(&:searchable)
```
roschaefer, line 9: Good catch :sweat:

db/data/20240726061113_add_organization_id_to_activity_notifications.rb
```diff
@@ -0,0 +1,29 @@
...
+class AddOrganizationIdToActivityNotifications < ActiveRecord::Migration[6.1]
+  def up
+    ActiveRecord::Base.transaction do
+      organization = Organization.singleton
+      return unless organization
+
+      ActivityNotification.find_each do |notification|
+        notification.organization_id = organization.id
+        notification.save!
+        Rails.logger.debug '.'
```
roschaefer, line 12: What's up with the dot? 

In previous times, applications would print a dot to stdout to have some IO and thus get priority by the operating system scheduler. :thinking:

db/seeds/multi_tenancy.rb
```diff
@@ -0,0 +1,85 @@
...
+Rails.logger.debug 'Seeding messages..'
+message_count = request_count * (contributors.count * 0.5).to_i
+message_count.times do
+  request = requests.sample
+  created_at = Time.zone.at(rand(request.broadcasted_at..Time.current))
+  FactoryBot.build(:message, request: request, created_at: created_at, updated_at: created_at) do |message|
+    message.class.skip_callback(:create, :after, :send_if_outbound, raise: false)
+    message.save!
+  end
+end
```
roschaefer, line 85: Good idea to simulate multiple organizations with seeds :muscle:

spec/models/message_spec.rb
```diff
@@ -124,13 +124,34 @@
...
 
-      it_behaves_like 'an ActivityNotification', 'MessageReceived'
+      context 'But it is for inbound messages' do
+        subject { create(:message, :inbound, request: request) }
+
+        it 'creates a notification of type MessageReceived' do
+          expect { subject }.to change(ActivityNotification.where(type: 'MessageReceived'), :count).by(organization.users.count + User.admin.count)
+        end
+
+        it 'for each user and admin' do
```
roschaefer, line 146: The description could do a better job on describing what's asserted here.

roschaefer (approved):
I agree with @soey regarding too many foreign keys.

However, I am not so sure about how much potential there is regarding DB normalization and delegation within the application code.

From my side, I don't see any blockers code-wise.

norma

app/activity_notifications/chat_message_sent.rb
mattwr18: So, we discussed this by phone, but I'll just write here for our future selves. In the case of the notifications, we added the organization_id as a param and save it to a activity_notification record in the database because we have admin who do not belong to any organization and we have potential clients that belong to multiple organizations so we use the organization id to filter the notifications to the organization that is set by the route.

Since we don't actually load the request itself, but simply pass in the request_id, it would be more complicated to derive the contributor, message, user, and organization from the activity_notification record without passing in these ids. 

What we do do, is that when we create the activity_notification record, we do use the request, sender, organization, etc. from the message record, which I guess addresses your concern that they somehow don't match. See :point_down: 

```rb
MessageReceived.with(
  contributor_id: sender_id,
  request_id: request.id,
  message_id: id,
  organization_id: organization.id
).deliver_later(organization.users + User.admin.all)
```

app/activity_notifications/message_received.rb
mattwr18:
```rb
MessageReceived.with(
  contributor_id: sender_id,
  request_id: request.id,
  message_id: id,
  organization_id: organization.id
).deliver_later(organization.users + User.admin.all)
```

app/activity_notifications/request_scheduled.rb
mattwr18:
```rb
RequestScheduled.with(request_id: request.id, organization_id: request.organization.id).deliver_later(User.all)
```

app/components/request_form/request_form.rb
mattwr18: Good point

app/models/request.rb
mattwr18: See :point_up:

db/data/20240726061113_add_organization_id_to_activity_notifications.rb
mattwr18: I like to see output when I run rake tasks so that I know something is happening hehe. Don't know if I should try to break this habit...

spec/models/message_spec.rb
mattwr18: See 704bc515

PR #1946, opened 2024-07-26
DRY out broadcast! method

mattwr18:
### What changed and Why

This PR reduces complexity and duplication by handling the message creation in `BroadcastRequestJob` instead of only handling planned requests there.


app/components/request_metrics/request_metrics.rb
```diff
@@ -21,7 +21,7 @@ def metrics
       [
         {
           value: stats[:counts][:contributors],
-          total: stats[:counts][:recipients],
+          total: request_for_info.organization.contributors.active.with_tags(request_for_info.tag_list).count,
```
mattwr18, line 24: This change became necessary because relying on the count of created outbound messages is flaky, especially when scheduling a job that creates them. What this meant in practice is that the user was redirected to the request show page after the request was successfully created, but since the message records themselves had not been created it falsely reported that no one had received the message.

spec/models/request_spec.rb
```diff
@@ -285,63 +285,19 @@
...
-        create(:contributor, :inactive, id: 5, telegram_id: 24, organization: organization)
+    it 'schedules a job to broadcast the request' do
+      expect { subject.call }.to have_enqueued_job(BroadcastRequestJob).with do |params|
+        expect(params[:request_id]).to eq(request.id)
       end
-
-      it { should change { Message.count }.from(0).to(1) }
-      it { should change { Message.pluck(:recipient_id) }.from([]).to([4]) }
-      it { should change { Message.pluck(:sender_id) }.from([]).to([request.user.id]) }
-      it { should change { Message.pluck(:broadcasted) }.from([]).to([true]) }
```
mattwr18, line 344: These specs have been migrated to the `spec/jobs/broadcast_request_job_spec.rb` See b5db6640

soey:
There are a few things that are not that intuitive for me ...

- Proposal: Why is broadcast! a class method if you handover then an instance of this class itself? Can't this not just be an instance method .. maybe to be renamed to sth like 'trigger_broadcast', 'schedule_broadcast' or 'handle_broadcasting' as it will not directly broadcast but handover the responsibility of broadcasting to the job.
- Remark: Even with the comment it took me a while to understand why we need to check for the request.planned? in the job again. First i thought we should put this decision solely to the calling method. What is a little odd, is that we now have to check the same conditional in to subsequently called methods. Well but it seems that it is necessary if we don't accept to delay the  RequestScheduled notification. (Maybe extend the comment?)
- Optional: I think it would be nice to move out the inner loop in the job and move it in an instance method "send_request(request)" on the contributor itself maybe even calling an send_message(message) on the Contributor instance after the message is set up -that would allow to remove the send_if_outbound from the after_create_commit on the message itself and allow to have the message commit only taking place after the message has really been sent - or if needed setting only the broadcasted to true if the sending succeeded.

PR #1949, opened 2024-08-01
refactor: OrganizationComponent

soey:
remaining code from #1949

- OrganizationComponent class has been added that provides access to the organization helper method to prevent prop drilling

Co-authored-by: Samuel Oey <samuel.oey@gmail.com>

### DIFF
--- a/app/components/organization_component.rb
+++ b/app/components/organization_component.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class OrganizationComponent < ApplicationComponent
+  def initialize(organization:, **)
+    @organization = organization
+
+    super
+  end
+
+  attr_reader :organization
+end


app/controllers/onboarding_controller.rb
```diff
@@ -131,7 +133,7 @@ def jwt_param
     params.require(:jwt)
   end
 
-  def organization
+  def organization_from_token
```
mattwr18, line 136: Can't we remove this now? We should no longer encode the organization id in the jwt, but instead use the organization from the invites route. Or do we keep this for some time to not invalidate existing links?

app/components/organization_component.rb
```diff
@@ -0,0 +1,11 @@
...
+
+class OrganizationComponent < ApplicationComponent
+  def initialize(organization:, **)
+    @organization = organization
+
+    super
+  end
+
+  attr_reader :organization
+end
```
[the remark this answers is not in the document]
roschaefer: Ok, I didn't rebased this branch but created PRs #1960 and #1959 to let CI check these changes separately and I will just rebase this PR so that only `OrganizationComponent` remains in the Files changed.

app/controllers/onboarding_controller.rb
roschaefer: Ah! I didn't know that. Then I will stop #1960 from getting merged.

PR #1951, opened 2024-08-02
Edit: scope the requests routes below the organization

soey:
changes the requests routes to be scoped below the organization

### DIFF
--- a/app/components/hero/hero.rb
+++ b/app/components/hero/hero.rb
@@ -1,6 +1,16 @@
 # frozen_string_literal: true
 
 module Hero
+  # FIXME: this component seems to be unused
   class Hero < ApplicationComponent
+    def initialize(organization:)
+      @organization = organization
+
+      super
+    end
+
+    private
+
+    attr_reader :organization
   end
 end
--- a/app/controllers/onboarding_controller.rb
+++ b/app/controllers/onboarding_controller.rb
@@ -25,8 +25,10 @@ class OnboardingController < ApplicationController
   end
 
   def create
+    # FIXME: there is a inherent inconsistency possible here the overall @organization is taken from
+    # the onboarding paths (if scoped) whereas the context of the contributor is taken from the token itself.
     @contributor = Contributor.new(contributor_params.merge(json_web_token_attributes: { invalidated_jwt: jwt_param },
-                                                            organization: organization))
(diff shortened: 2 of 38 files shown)


app/controllers/onboarding_controller.rb
```diff
@@ -25,8 +25,10 @@ def show
   end
 
   def create
+    # FIXME: there is a inherent inconsistency possible here the overall @organization is taken from
+    # the onboarding paths (if scoped) whereas the context of the contributor is taken from the token itself.
```
mattwr18, line 29: I remember this being discussed, but I'm not sure if it was over the phone, or written.

Do we need to keep the `organization_from_token` at all? I would expect that we would no longer maintain this way of assigning the organization to the contributor.

Would the changes to the routing proposed in these several related PRs invalidate all existing onboarding links? They would not contain the org id, and we don't have a redirect so my assumption would be that it would invalidate all existing onboarding links.

If this is the case, and we don't want to look into how to solve this, I would remove the org id from the jwt creation and remove this.

If we somehow want to maintain this method to extract the organization from the jwt, I would expect that we would have a check that it is the same organization as the one derived from the route and if not to throw an error.

Two questions:
* Are these changes in the onboarding controller somehow related to the requests routes, or should they be moved?
* Is there a use case for sending one link to a contributor and them onboarding for multiple organizations? I guess for now, we are going to keep it simple and the contributor would need to onboard to the orgs separately, with separate links.

mattwr18 (approved):
I got a url generation error locally cause my seeded requests did not have an organization_id so there was a missing constraints organization_id. This is to be expected, of course, but I think it would be worth making a sanity check on production instances. We ran the **AddOrganizationIdToRequests** data migration so there should not be any, but it would be nice to double check.

See my comments below :point_down:

app/components/hero/hero.rb
```diff
@@ -1,6 +1,16 @@
 # frozen_string_literal: true
 
 module Hero
+  # FIXME: this component seems to be unused
```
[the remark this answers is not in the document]
soey: https://github.com/tactilenews/100eyes/pull/1976

app/controllers/onboarding_controller.rb
soey: obsolete fixed in another branch

PR #1952, opened 2024-08-04
feat: contributors path scoped by organization

soey:
scopes the contributors routes by the organization

### DIFF
--- a/spec/components/contributor_email_settings_spec.rb
+++ b/spec/components/contributor_email_settings_spec.rb
@@ -18,7 +18,7 @@ RSpec.describe ContributorEmailSettings::ContributorEmailSettings, type: :compon
   it { should have_css('h2', text: 'E-Mail') }
   it { should have_css('p', text: 'Max Mustermann hat sich via E-Mail angemeldet.') }
 
-  it { should have_css("form[action='/contributors/#{contributor.id}']") }
+  it { should have_css("form[action='/#{contributor.organization_id}/contributors/#{contributor.id}']") }
   it { should have_css('input[type="email"][value="muster@example.org"]') }
   it { should have_css('button', text: 'Speichern') }
 end
--- a/spec/system/dashboard/activity_notifications_spec.rb
+++ b/spec/system/dashboard/activity_notifications_spec.rb
@@ -43,7 +43,7 @@ RSpec.describe 'Activity Notifications' do
         "#{contributor.name} hat sich via #{contributor.channels.first.to_s.capitalize} angemeldet."
       )
       expect(page).to have_text('vor eine Minute')
-      expect(page).to have_link('Zum Profil', href: contributor_path(contributor))
+      expect(page).to have_link('Zum Profil', href: organization_contributor_path(contributor.organization, contributor))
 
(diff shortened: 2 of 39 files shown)


spec/components/contributor_email_settings_spec.rb
```diff
@@ -5,20 +5,22 @@
...
   end
 
-  let(:params) { { contributor: contributor } }
+  let(:params) { { contributor: contributor, organization: organization } }
 
   it { should have_css('h2', text: 'E-Mail') }
   it { should have_css('p', text: 'Max Mustermann hat sich via E-Mail angemeldet.') }
 
-  it { should have_css("form[action='/contributors/#{contributor.id}']") }
+  it { should have_css("form[action='/#{contributor.organization_id}/contributors/#{contributor.id}']") }
```
roschaefer, line 23: I prefer to set hard-coded ids and repeat them in the assertions.

spec/system/dashboard/activity_notifications_spec.rb
```diff
@@ -126,15 +127,16 @@
       reply_text = "I'm a reply from the same contributor: #{contributor_two.name}"
       expect(page).to have_text(reply_text)
       find('p', text: reply_text).hover
-      reply_path = conversations_contributor_path(id: contributor_two, reply_to: reply_by_same_contributor.id, anchor: 'chat-form')
+      reply_path = conversations_organization_contributor_path(organization, contributor_two, reply_to: reply_by_same_contributor.id,
```
roschaefer, line 130: Removed `id: ` named param is fine, I guess.

roschaefer (approved):
yes

Yes! This one looks good and mergeable.

roschaefer:
Merge conflicts after rebase, huh? I just figured that base branch here is not `origin/main`. Now I'm confused. I would like to merge this into `main`. Can we?

soey:
> Merge conflicts after rebase, huh? I just figured that base branch here is not `origin/main`. Now I'm confused. I would like to merge this into `main`. Can we?

We decided to merge all in the releases/route-scoping to do an accceptance testing once there.

spec/system/dashboard/activity_notifications_spec.rb
soey: the path helpers are so clever, yes it works..but tbh. i've mixed feelings on the cleverness. I am a bit  puzzled that id:contributor_two seem to have worked..

spec/components/contributor_email_settings_spec.rb
soey: you're right on this as it won't work when a n:n is introduced for contributors and organizations. changed with 9ef1a28cee914e2892effb458dfbd97ee0a48702

mattwr18: I think that is not what @roschaefer was saying. I know from working with Robert that he prefers to set the id to some hard-coded value `234_321` and then assert this hard-coded value over using something like `organization.id`, which can give false positives.

mattwr18 (approved):
Also looks good to me! :tada:

PR #1953, opened 2024-08-05
Feat: scope dashboard routes

soey:
Scopes the dashboard routes by the organization 

TODOS:

- [x] rebase on scope-routes-for-onboarding
- [x] keep the unauthorized messages configurable by organization
- [x] change the analytical queries to be scoped by the organization

### DIFF
--- a/app/controllers/otp_setup_controller.rb
+++ b/app/controllers/otp_setup_controller.rb
@@ -14,7 +14,7 @@ class OtpSetupController < ApplicationController
       current_user.save!
       session[:otp_verified_for_user] = current_user.id
 
-      redirect_back_or dashboard_path
+      redirect_back_or organizations_path
     else
       flash.now[:error] = I18n.t('sessions.errors.otp_incorrect')
       render :show, status: :unauthorized
@@ -24,7 +24,7 @@ class OtpSetupController < ApplicationController
   private
 
   def redirect_if_set_up
-    redirect_to dashboard_path if current_user.otp_enabled?
+    redirect_to organizations_path if current_user.otp_enabled?
   end
 
   def otp_params
--- a/app/views/organizations/index.html.erb
+++ b/app/views/organizations/index.html.erb
@@ -1,0 +1,16 @@
+<%= c 'main', style: :lightestGrayBackground do %>
+  <%= c 'page_header', styles: [:wide, :inheritBackgroundColor, :xlargePaddingTop, :spaceBetween] do %>
+    <%= c 'heading' do %>
+      <%= t ".heading", first_name: current_user.first_name %>
+    <% end %>
+  <% end %>
+
+
+  <%= c 'section', styles: [:wide, :largeBottomMargin], class: 'OrganizationsList' do %>
+    <ul>
+      <% @organizations.each do |organization| %>
+        <li><%= link_to organization.name, organization_dashboard_path(@organization) %><li>
+      <% end %>
+    <ul>
+  <% end %>
+<% end %>
--- a/app/views/organizations/index.html.erb
+++ b/app/views/organizations/index.html.erb
@@ -6,11 +6,5 @@
   <% end %>
 
 
-  <%= c 'section', styles: [:wide, :largeBottomMargin], class: 'OrganizationsList' do %>
-    <ul>
-      <% @organizations.each do |organization| %>
-        <li><%= link_to organization.name, organization_dashboard_path(@organization) %><li>
-      <% end %>
-    <ul>
-  <% end %>
+  <%= c 'set_organization_form', organizations: @organizations %>
 <% end %>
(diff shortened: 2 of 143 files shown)


app/controllers/otp_setup_controller.rb
```diff
@@ -24,7 +24,7 @@ def create
   private
 
   def redirect_if_set_up
-    redirect_to dashboard_path if current_user.otp_enabled?
+    redirect_to organizations_path if current_user.otp_enabled?
```
roschaefer, line 27: How would onboarding of users look like? Is it OK to validate that a user has at least one organization or is an admin? I think it would be odd to get redirected to the organizations listing page.

roschaefer (changes requested):
image


I'll press "Request changes" here. I would like to find a better solution than an organization listing.

not quite

@soey speaking of your comments, I'm super happy to move `onboarding_unauthorized_heading` and `onboarding_unauthorized_text` out of the organization.

app/controllers/otp_setup_controller.rb
soey: On the organizations page is a redirect to the organization_dashboard page if the user only has/belongs to one organization. So technically there will be 2 redirects following each other. So UX wise nothing changes to a user being onboarded to one organization (see system spec).

soey:
> soey speaking of your comments, I'm super happy to move `onboarding_unauthorized_heading` and `onboarding_unauthorized_text` out of the organization.

done with: https://github.com/tactilenews/100eyes/pull/1953/commits/62600a572ef5b8e6cbc278e591c73c834637b8e3

mattwr18:
> @mattwr18 @roschaefer - please have a short look here. It seems that I have to move the onboarding_unauthorized_heading and the onboarding_unauthorized_text out of the organization scope if we leave the routes where JWT validation is done in a global scope.

I don't really understand this comment. Why would the JWT validation be done with a global scope?

I think you provided some more context in your comment in [Slack](https://tactilenews.slack.com/archives/D06PL6BG2GY/p1723450676016959?thread_ts=1723446044.913649&cid=D06PL6BG2GY).

Yes, I would expect that we would a) remove the organization_id from the JWT as we no longer need this because b) We are now going to scope the routes themselves by the organization the contributor will onboard to.

I don't really mind if we want to have a generic unauthorized message, that is something you could talk to someone on the team that has more contact with the clients to see if they actually customize this ever, I guess they do since the default asks them to send a message to our email.

app/views/organizations/index.html.erb
```diff
@@ -0,0 +1,16 @@
...
+
+
+  <%= c 'section', styles: [:wide, :largeBottomMargin], class: 'OrganizationsList' do %>
+    <ul>
+      <% @organizations.each do |organization| %>
+        <li><%= link_to organization.name, organization_dashboard_path(@organization) %><li>
+      <% end %>
+    <ul>
+  <% end %>
+<% end %>
```
mattwr18, line 16: I'm also not satisfied with how this looks. I proposed some changes that I think are an improvement of this in https://github.com/tactilenews/100eyes/pull/1969

It would look like this.

Screenshot_20240812_112111
Screenshot_20240812_112138

It's an auto submit form that sets the organization for the admin/user with multiple organizations.

It does not include the navbar since we don't want to give a way to navigate around the site if we don't know which organization to set.

app/controllers/otp_setup_controller.rb
mattwr18: Although, the redirect in the organizations path does have the benefit of preventing a user who belongs to only 1 organization of visiting `/organizations` and setting an organization they do not belong to.

mattwr18:
@soey they biggest thing I am concerned about this PR, which maybe solved in another PR that I have not reviewed yet, is this allows a user to access any other organizations routes. There is no check to see if the organization in the path is one that the current user belongs to.

app/views/organizations/index.html.erb
soey: thnx, just integrated most of your proposal with https://github.com/tactilenews/100eyes/pull/1953/commits/703942b3d4e44e96193f625ce36845ad353b85b1

is there any reason that you didn't use plain old links so that we have the form and route overhead?

app/controllers/otp_setup_controller.rb
soey: I think there is no change needed, as currently user were always belonging only to one organization - and so there will always the redirect. As soon as we add the n:n relation here the code in the OrganizationsController has to be changed anyway.

app/views/organizations/index.html.erb
mattwr18: I'm not sure what you mean by:

> so we have the form and route overhead

Plain old links could have some benefits over a select. They are a little less flexible when you have many options, but in this use case, it's hard to say what the likelihood is that we will have 20 organization, much less that some user will be part of all 20.

When I looked at this bit of code, it wasn't obvious to me why the extra link was being created. I did not attempt to style the links to remove the native browsers bullet points, which probably would have hidden this extra empty link.

PR #1956, opened 2024-08-05
feat: scope invites

soey:


soey:
replaced by https://github.com/tactilenews/100eyes/pull/1965

PR #1960, opened 2024-08-06
refactor: better naming - organization_from_token

roschaefer:
taken from #1949

Can be tested on CI independently.


soey:
alternative solution see https://github.com/tactilenews/100eyes/pull/1967/commits/eab701b67449d287a997a0a286cfeda5a37172a3

PR #1962, opened 2024-08-06
Feat: Nav Bar component knows about the organization

soey:
This allows the upcoming route change PRs to be more independent of each other


roschaefer (approved):
There is no change in behaviour regarding this PR, right?

why not

PR #1963, opened 2024-08-06
feat: scope settings routes by organization

soey:


roschaefer (approved):
fine

mattwr18 (commented):
Sorry, I started to review this, but didn't hit submit, see my question below @soey

PR #1964, opened 2024-08-07
Feat: scope search cleaned

soey:


roschaefer (approved):
I tested it with #1968 and it looked fine.

The list of requests on `/requests` is not scoped at all at the moment.

scope

roschaefer:
Closed by accident

PR #1967, opened 2024-08-08
Feat: scope routes for onboarding

soey:


app/controllers/onboarding_controller.rb
```diff
@@ -131,12 +131,6 @@ def jwt_param
     params.require(:jwt)
   end
 
-  def organization
-    decoded_token = JsonWebToken.decode(jwt_param)
-    organization_id = decoded_token.first['data']['organization_id'].to_i
-    Organization.find(organization_id)
-  end
-
```
roschaefer, line 139: I guess that's the only tricky part.

roschaefer (changes requested):
On f761c38f7212c070329f41d73e2c1c92beca5295 after `bin/rails db:drop db:create db:migrate db:seed` and after I login in with the default user, when I do the following:

1. Create an invitation link
2. Paste the invitation link into a new tab
3. I see this:

image

app/controllers/onboarding_controller.rb
soey: We should probably add some channels where onboarding is allowed in the seeds. This is caused by: https://github.com/tactilenews/100eyes/blob/main/app/controllers/onboarding_controller.rb#L145

soey: Sidemark: This is not caused by this PR but also in 'main'. The current implementation checks if there is any channel available and set up and only renders the invitations page in this case. This need to be done on production instances anyway but to be nice to us devs I have a PR that saves us some work: https://github.com/tactilenews/100eyes/pull/1979

roschaefer (approved):

PR #1971, opened 2024-08-12
Add many-to-many relationship between users and organizations

mattwr18:

This introduces the many-to-many relationship that is a requirement for some of our clients to belong to more than one organization.

Instead of a simply HABTM, it uses `has_many, through:`. This gives us more flexibility to save information on the relationship itself. For example, how do we handle active/inactive state. One idea would be to add a `deactivated_at` in the `users_organizations` table.

### TODOs

- [ ] Add automated tests that test for a user that belongs to multiple organizations
- [x] Decide on the deactivated/reactivated logic
- [x] When an admin is creating a new user, try to find the user by their email first and do something - ie either simply add the user to the organization, or throw a validation error and direct the admin to edit the existing user. EDIT: The email is already validated for uniqueness by the clearance gem. An error is displayed if someone tries to create a user with the same email.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,8 +2,7 @@
 
 class ApplicationController < ActionController::Base
   include Clearance::Controller
-  before_action :require_login, :require_otp_setup
-  before_action :set_organization
+  before_action :require_login, :require_otp_setup, :set_organization, :user_permitted?
 
   def require_otp_setup
     redirect_to otp_setup_path if signed_in? && !current_user.otp_enabled?
@@ -14,7 +13,7 @@ class ApplicationController < ActionController::Base
 
     super(user) do |status|
       if status.success?
-        redirect_to organizations_path
+        redirect_to redirect_path
       else
         flash.now.alert = status.failure_message
         render template: 'sessions/new', status: :unauthorized
@@ -34,9 +33,21 @@ class ApplicationController < ActionController::Base
 
   private
 
+  def user_permitted?
+    raise ActionController::RoutingError, 'Not Found' unless current_user.admin? || @organization.in?(current_user.organizations)
+  end
+
   def set_organization
     @organization = Organization.find(params[:organization_id]) if params[:organization_id].present?
   rescue ActiveRecord::RecordNotFound
     raise ActionController::RoutingError, 'Not Found'
+  end
+
+  def redirect_path
+    if current_user.admin? || current_user.organizations.count > 1
+      organizations_path
+    else
+      organization_dashboard_path(current_user.organizations.first)
+    end
   end
 end
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -1,13 +1,15 @@
 # frozen_string_literal: true
 
 class OrganizationsController < ApplicationController
-  skip_before_action :set_organization
+  skip_before_action :user_permitted?, :set_organization
   layout 'minimal'
 
   def index
-    # TODO: change this when having a organization habtm user relation
-    redirect_to organization_dashboard_path(current_user.organization) if current_user.organization.present?
-
     @organizations = Organization.all
+  end
+
+  def set_organization
+    organization = Organization.find(params[:organization_id])
+    redirect_to organization_dashboard_path(organization)
   end
 end
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -10,7 +10,7 @@ class Organization < ApplicationRecord
   belongs_to :business_plan
   belongs_to :contact_person, class_name: 'User', optional: true
   has_many :users_organizations, dependent: :destroy
-  has_many :users, through: :users_organizations, dependent: :destroy
+  has_many :users, through: :users_organizations, dependent: :restrict_with_exception
   has_many :contributors, dependent: :destroy
   has_many :requests, dependent: :destroy
   has_many :notifications_as_mentioned, class_name: 'ActivityNotification', dependent: :destroy
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -10,7 +10,7 @@ class Organization < ApplicationRecord
   belongs_to :business_plan
   belongs_to :contact_person, class_name: 'User', optional: true
   has_many :users_organizations, dependent: :destroy
-  has_many :users, through: :users_organizations, dependent: :restrict_with_exception
+  has_many :users, through: :users_organizations
   has_many :contributors, dependent: :destroy
   has_many :requests, dependent: :destroy
   has_many :notifications_as_mentioned, class_name: 'ActivityNotification', dependent: :destroy
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -9,7 +9,8 @@ class Organization < ApplicationRecord
 
   belongs_to :business_plan
   belongs_to :contact_person, class_name: 'User', optional: true
-  has_many :users, class_name: 'User', dependent: :destroy
+  has_many :users_organizations, dependent: :destroy
+  has_many :users, through: :users_organizations, dependent: :destroy
   has_many :contributors, dependent: :destroy
   has_many :requests, dependent: :destroy
   has_many :notifications_as_mentioned, class_name: 'ActivityNotification', dependent: :destroy
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -9,7 +9,8 @@ class User < ApplicationRecord
            dependent: :destroy
   has_many :notifications_as_mentioned, class_name: 'ActivityNotification', dependent: :destroy
   has_many :messages, as: :sender, dependent: :destroy
-  belongs_to :organization, optional: true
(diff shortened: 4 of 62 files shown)


soey:
Personally I would start with the most simple version (habtm) and only extend it to a has_many_through when it is really needed. But I have no hard feeling either against it.

app/models/organization.rb
```diff
@@ -9,7 +9,8 @@ class Organization < ApplicationRecord
 
   belongs_to :business_plan
   belongs_to :contact_person, class_name: 'User', optional: true
-  has_many :users, class_name: 'User', dependent: :destroy
+  has_many :users_organizations, dependent: :destroy
+  has_many :users, through: :users_organizations, dependent: :destroy
```
soey, line 13: Would that destroy a user if the organization is destroyed? I think we should keep the user as it can also belong to other organizations. The standard behavior of this should be to do nothing on the user itself and just destroy the relation.

app/models/user.rb
```diff
@@ -9,7 +9,8 @@ class User < ApplicationRecord
            dependent: :destroy
   has_many :notifications_as_mentioned, class_name: 'ActivityNotification', dependent: :destroy
   has_many :messages, as: :sender, dependent: :destroy
-  belongs_to :organization, optional: true
+  has_many :users_organizations, dependent: :destroy
+  has_many :organizations, through: :users_organizations, dependent: :destroy
```
soey, line 13: same as above. destroying a user should never destroy the organization.

db/migrate/20240812102032_create_users_organizations.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateUsersOrganizations < ActiveRecord::Migration[6.1]
```
soey, line 3: we might need a follow up migration to remove the organization_id from the user

spec/factories/users_organizations.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
```
soey, line 1: that would probably be obsolete if we would stick with a simple habtm for now..

spec/models/users_organization_spec.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe UsersOrganization, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
soey, line 6: ;) habtm unless we have something here?

soey (changes requested):
Please see the remarks and please (as we talked about toray) add e.g. "user_permitted?" into the ApplicationController called by the setOrganization to check if a user is allowed to access the set organization.

spec/models/users_organization_spec.rb
roschaefer: @soey I thought you and I we both agreed that it would be the cleanest solution to have a join model between users and organization? I wouldn't be against starting with `has_many, through:`

roschaefer:
@mattwr18 I'm willing to continue on this PR and address the requested changes, unless you want to do it.

mattwr18:
@roschaefer I don't know which changes I'll actually implement. I don't think changing to a HABTM association for no real benefit at the expense of flexibility is something I would. There are also other changes that we discussed that @soey didn't mention. We talked about the need for a validation on user email, for example.

spec/models/users_organization_spec.rb
soey: It's ok for me.

soey: @roschaefer: I think you are referring to the relationship between contributor and organization .. where we probably have to implement the note and mount the taggings  to

db/migrate/20240812102032_create_users_organizations.rb
mattwr18: Yes, we do need one. We should create an issue for this before you merge/deploy this.

app/models/organization.rb
```diff
@@ -9,7 +9,8 @@ class Organization < ApplicationRecord
 
   belongs_to :business_plan
   belongs_to :contact_person, class_name: 'User', optional: true
-  has_many :users, class_name: 'User', dependent: :destroy
+  has_many :users_organizations, dependent: :destroy
+  has_many :users, through: :users_organizations, dependent: :restrict_with_exception
```
soey, line 13: can't we just leave the dependent param out. To my understanding the users is just kind of a shortcut going through the users_organization. If the users_organizations is destroyed (which is correct) then there is nothing to be done at the users at all.

mattwr18: We can. I thought that rubocop would complain and I think that raising an exception is a valid thing to do in this case, but the default is nil, which also makes sense. The thing is there is should be no UI to delete a user as this  leads to an exception if there are any requests by that user.

```
ActiveRecord::InvalidForeignKey: PG::ForeignKeyViolation: ERROR:  update or delete on table "users" violates foreign key constraint "fk_rails_8ead8b1e6b" on table "requests"
DETAIL:  Key (id)=(5) is still referenced from table "requests".
```

There is also no UI to delete an organization.

See 8e2b1fb0

db/migrate/20240812102032_create_users_organizations.rb
soey: #1977

app/controllers/application_controller.rb
```diff
@@ -35,10 +35,17 @@ def delete_otp_session_variables
   private
 
   def set_organization
-    @organization = if params[:organization_id].present?
-                      Organization.find(params[:organization_id])
-                    else
-                      Organization.singleton
-                    end
+    organization = if params[:organization_id].present?
```
soey, line 38: just as an idea for a different possible implementation:
```
def set_organization
   collection = current_user.admin? ? Organization : current_user.organizations
   collection.find(params[:organization_id]) if params[:organization_id].present?
   ...
```

the additional user_permitted? method would then be obsolete. What I like on your approach is that is quite explicit on checking but it has some overhead compared to the above.

db/data/20240812102718_migrate_users_to_users_organizations.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class MigrateUsersToUsersOrganizations < ActiveRecord::Migration[6.1]
+  def up
+    User.admin(false).find_each do |user|
```
soey, line 5: I was just wondering about L52 in User.rb. With that in place I assume that there are situations where an admin also has the organization_id set. We should decide if we either remove that admin scope there or include the admins in the migration as well.

spec/factories/users_organizations.rb
```diff
@@ -0,0 +1,8 @@
...
+
+FactoryBot.define do
+  factory :users_organization do
+    user { nil }
```
soey, line 5: just for my interest .. why is this { nil } here?

spec/requests/sessions_request_spec.rb
```diff
@@ -3,7 +3,8 @@
 require 'rails_helper'
 
 RSpec.describe 'Sessions', type: :request do
-  let!(:user) { create(:user, email: 'zora@example.org', password: '12345678', otp_enabled: otp_enabled) }
+  let(:organization) { create(:organization) }
+  let!(:user) { create(:user, email: 'zora@example.org', password: '12345678', otp_enabled: otp_enabled, organizations: [organization]) }
   let(:otp_enabled) { false }
 
   before { create(:organization) }
```
soey, line 10: why a let and a separate org creation here?

soey (changes requested):

app/controllers/application_controller.rb
```diff
@@ -2,8 +2,7 @@
 
 class ApplicationController < ActionController::Base
   include Clearance::Controller
-  before_action :require_login, :require_otp_setup
-  before_action :set_organization
+  before_action :require_login, :require_otp_setup, :user_permitted?, :set_organization
```
soey, line 5: I would just reorder to
`  before_action :require_login, :require_otp_setup, :set_organization, :user_permitted?`

soey (changes requested):
just some minor change proposals to the application controller- all other is good for now.

app/controllers/organizations_controller.rb
```diff
@@ -1,13 +1,15 @@
...
   layout 'minimal'
 
   def index
-    # TODO: change this when having a organization habtm user relation
-    redirect_to organization_dashboard_path(current_user.organization) if current_user.organization.present?
-
     @organizations = Organization.all
   end
+
+  def set_organization
```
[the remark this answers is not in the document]
mattwr18: because we haven't opened a PR that moves from setting the organization with a form to simply displaying links.
It should be removed in that PR.

spec/factories/users_organizations.rb
mattwr18: this was auto generated when I did

```
be rails g model UsersOrganizations user:references organization:references
```

we could maybe remove the factory as it is unused.

soey (approved):

PR #1972, opened 2024-08-14
release: integration branch for all missing route scoping

soey:
sets up a release branch for route-scoping

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,6 @@
 name: CI
 
-on:
-  pull_request:
-    branches: [ main, 'releases/**']
+on: pull_request
 
 jobs:
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   pull_request:
-    branches: [ main ]
+    branches: [ main, 'releases/**']
 
 jobs:
   test:
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -10,5 +10,4 @@ class OrganizationsController < ApplicationController
 
     @organizations = Organization.all
   end
-
 end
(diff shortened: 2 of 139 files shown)


.github/workflows/ci.yml
```diff
@@ -2,7 +2,7 @@ name: CI
 
 on:
   pull_request:
-    branches: [ main ]
+    branches: [ main, 'releases/**']
```
roschaefer, line 5: The meaning of this is to run CI on PRs targeting any `releases/**` branch. What about removing line 5 and run it on any PR? It's just CI not CD.

roschaefer (approved):
I would give this an "Approve" if the files changed would only contain `.github/workflows/ci.yml`. Maybe @soey can help me, why the other changes?

not sure

.github/workflows/ci.yml
mattwr18: I agree with @roschaefer ... that is what I'd like.

mattwr18:
> I would give this an "Approve" if the files changed would only contain `.github/workflows/ci.yml`. Maybe @soey can help me, why the other changes?
 
I think the purpose of this PR is to be able to manually test all changes together as they would be when deploying the feature, but still have the flexibility to have smaller PRs that can be merged into a base branch before that base branch is ultimately merged into main and deployed.

I find this PR helpful to be able to test, but I also would want to actually test the base branch after everything is merged into it because I think there will be merge conflicts that need to be worked through and sometimes we make mistakes, which can break things.

.github/workflows/ci.yml
soey: done with: eca70b8a0a4d87a978e408db2a3da0b3e645646a

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    # TODO: change this when having a organization habtm user relation
+    redirect_to organization_dashboard_path(current_user.organization) if current_user.organization.present?
+
+    @organizations = Organization.all
+  end
+
+  def set_organization
+    organization = Organization.find(params[:organization_id])
+    redirect_to organization_dashboard_path(organization)
+  end
```
mattwr18, line 17:
```suggestion
 ```

I think this was copied from the PR that I suggested styling the organizations index page. I was playing around with something and it didn't reallly work as I expected. This method is never called and should be removed.

PR #1973, opened 2024-08-14
improvement: test flakiness

soey:
Try to fix phony caused flakyness... let's see..

### List of recently failed specs in CI

- [x] rspec ./spec/system/admin/auth_spec.rb:9 # Auth as signed-out user user tries to access. Reason:

```
Ferrum::ProcessTimeoutError
```

- [x] rspec ./spec/mailboxes/replies_mailbox_spec.rb:34 # RepliesMailbox given a contributor is expected not to change `Message.count`. Reason:

```
 ActiveRecord::RecordInvalid:
       Gültigkeitsprüfung ist fehlgeschlagen: Telegram bot username ist bereits vergeben
```
- [x] rspec ./spec/models/request_spec.rb:234 # Request#stats given a number of requests, replies and photos [:counts][:replies] messages from us are excluded. Reason: Also telegram bot username.

- [x] rspec ./spec/system/admin/auth_spec.rb:9 # Auth as signed-out user user tries to access. Reason: Same timeout issue
- [x] rspec ./spec/models/contributor_spec.rb:677. Reason:  Telegram bot username uniqueness
- [x] rspec ./spec/system/admin/auth_spec.rb:9 # Auth as signed-out user user tries to access. Reason: Same timeout issue
- [x] rspec ./spec/adapters/whats_app_adapter/twilio_outbound/text_spec.rb:68. Reason:

```
ActiveRecord::RecordInvalid:
       Gültigkeitsprüfung ist fehlgeschlagen: Whats app phone number ist keine gültige Nummer
```

- [x] rspec './spec/jobs/resubscribe_contributor_job_spec.rb[1:1:1:2:1:2]: Invalid Signal number
- [x] rspec ./spec/system/admin/auth_spec.rb:9 Same timeout error
- [x] rspec ./spec/system/admin/auth_spec.rb:9 Same timeout error
- [x] rspec ./spec/system/admin/stats_spec.rb:19 Telegram bot uniqueness.
- [x] rspec ./spec/system/dashboard/activity_notifications_spec.rb

```
expect(page).to have_css('svg.Avatar-initials')
       expected to find css "svg.Avatar-initials" but there were no matches
```

### DIFF
--- a/spec/factories/contributors.rb
+++ b/spec/factories/contributors.rb
@@ -43,7 +43,7 @@ FactoryBot.define do
     trait :signal_contributor do
       after(:build) do |contributor|
         contributor.email = nil
-        contributor.signal_phone_number = Faker::PhoneNumber.cell_phone_in_e164
+        contributor.signal_phone_number = Faker::PhoneNumber.unique.cell_phone_in_e164
       end
     end
 
@@ -58,7 +58,7 @@ FactoryBot.define do
       after(:build) do |contributor|
         contributor.email = nil
         # FIXME: Faker seems to create invalid phone numbers sometimes
-        contributor.whats_app_phone_number = Faker::PhoneNumber.cell_phone_in_e164
+        contributor.whats_app_phone_number = Faker::PhoneNumber.unique.cell_phone_in_e164
       end
     end
 
--- a/spec/factories/contributors.rb
+++ b/spec/factories/contributors.rb
@@ -43,7 +43,7 @@ FactoryBot.define do
     trait :signal_contributor do
       after(:build) do |contributor|
         contributor.email = nil
-        contributor.signal_phone_number = Faker::PhoneNumber.cell_phone
+        contributor.signal_phone_number = Faker::PhoneNumber.cell_phone_in_e164
       end
     end
 
@@ -58,7 +58,7 @@ FactoryBot.define do
       after(:build) do |contributor|
         contributor.email = nil
         # FIXME: Faker seems to create invalid phone numbers sometimes
-        contributor.whats_app_phone_number = Faker::PhoneNumber.cell_phone
+        contributor.whats_app_phone_number = Faker::PhoneNumber.cell_phone_in_e164
       end
     end
 
--- a/spec/factories/organizations.rb
+++ b/spec/factories/organizations.rb
@@ -10,9 +10,9 @@ FactoryBot.define do
     onboarding_success_heading { File.read(File.join('config', 'locales', 'onboarding', 'success_heading.txt')) }
     onboarding_success_text { File.read(File.join('config', 'locales', 'onboarding', 'success_text.txt')) }
     onboarding_data_protection_link { 'https://tactile.news/100eyes-datenschutz/' }
-    signal_server_phone_number { Faker::PhoneNumber.cell_phone }
-    whats_app_server_phone_number { Faker::PhoneNumber.cell_phone }
-    telegram_bot_username { Faker::Internet.username }
+    signal_server_phone_number { Faker::PhoneNumber.unique.cell_phone_in_e164 }
+    whats_app_server_phone_number { Faker::PhoneNumber.unique.cell_phone_in_e164 }
+    telegram_bot_username { Faker::Internet.unique.username }
 
     transient do
       users_count { 0 }
--- a/spec/support/cuprite_setup.rb
+++ b/spec/support/cuprite_setup.rb
@@ -38,7 +38,8 @@ Capybara.register_driver(:better_cuprite) do |app|
       browser_options: remote_chrome ? { 'no-sandbox' => nil } : {},
       inspector: true,
       headless: !ENV['HEADLESS'].in?(%w[n 0 no false]),
-      process_timeout: 20
+      process_timeout: 20,
+      timeout: 20
     }.merge(remote_options)
   )
 end
--- a/spec/support/cuprite_setup.rb
+++ b/spec/support/cuprite_setup.rb
@@ -37,7 +37,8 @@ Capybara.register_driver(:better_cuprite) do |app|
       window_size: [1200, 800],
       browser_options: remote_chrome ? { 'no-sandbox' => nil } : {},
       inspector: true,
-      headless: !ENV['HEADLESS'].in?(%w[n 0 no false])
+      headless: !ENV['HEADLESS'].in?(%w[n 0 no false]),
+      process_timeout: 20
     }.merge(remote_options)
   )
 end
--- a/spec/system/admin/auth_spec.rb
+++ b/spec/system/admin/auth_spec.rb
@@ -5,38 +5,24 @@ require 'rails_helper'
 RSpec.describe 'Auth' do
   let(:organization) { create(:organization) }
 
-  describe 'as signed-out user' do
-    it 'user tries to access' do
-      visit admin_users_path
-      expect(page).to have_http_status(:not_found)
-    end
-  end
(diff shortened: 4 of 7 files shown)


mattwr18:
I looked a little more into this and found some information that might help us.
There is an issue on the Phony gem repo https://github.com/floere/phony/issues/440 that is a bit outdated, but has some nice info.

Currently, we use `Faker::PhoneNumber.cell_phone` to create signal/whats_app phone numbers. This produces numbers like _163-1160010_. If I generate 1000 of these numbers and check how many of them do not pass the `plausible?` check from phony. I get ~650. Whereas if I change this to ``Faker::PhoneNumber.cell_phone_with_country_code`, it produces numbers like _+49 167-2217487_ and I get ~2 of 1000 invalid. Changing to `Faker::PhoneNumber.cell_phone_in_e164` produces numbers like _+491672549559_ , with similar fails percentages.

All the invalid ones seems to have the format `1672`.

Although there would still be some flakiness, I think this flakiness might reflect some flakiness in the app itself and would be a huge step in the right direction. I personally would go with the `Faker::PhoneNumber.cell_phone_in_e164` because that seems most similar to how the numbers are normalized before being saved to the db.

soey:
> I looked a little more into this and found some information that might help us. There is an issue on the Phony gem repo [floere/phony#440](https://github.com/floere/phony/issues/440) that is a bit outdated, but has some nice info.
> 
> Currently, we use `Faker::PhoneNumber.cell_phone` to create signal/whats_app phone numbers. This produces numbers like _163-1160010_. If I generate 1000 of these numbers and check how many of them do not pass the `plausible?` check from phony. I get ~650. Whereas if I change this to ``Faker::PhoneNumber.cell_phone_with_country_code`, it produces numbers like _+49 167-2217487_ and I get ~2 of 1000 invalid. Changing to `Faker::PhoneNumber.cell_phone_in_e164` produces numbers like _+491672549559_ , with similar fails percentages.
> 
> All the invalid ones seems to have the format `1672`.
> 
> Although there would still be some flakiness, I think this flakiness might reflect some flakiness in the app itself and would be a huge step in the right direction. I personally would go with the `Faker::PhoneNumber.cell_phone_in_e164` because that seems most similar to how the numbers are normalized before being saved to the db.

Great. Just one thought ..I assume that we somehow normalize also the incoming phone numbers and save them in the db in a normalized format. I think it would be best to use this format (is this in_e164?) - or maybe even create a number that goes through that normalizer once on setup?

mattwr18:
> Great. Just one thought ..I assume that we somehow normalize also the incoming phone numbers and save them in the db in a normalized format. I think it would be best to use this format (is this in_e164?) - or maybe even create a number that goes through that normalizer once on setup?

According to [Phony's description](https://github.com/floere/phony?tab=readme-ov-file#description):

> This gem normalizes, formats and splits [E164 phone numbers](http://en.wikipedia.org/wiki/E.164). A valid E164 phone number must include a country code.

PhonyRails is just a wrapper for Phony gem.

spec/factories/contributors.rb
```diff
@@ -58,7 +58,7 @@
       after(:build) do |contributor|
         contributor.email = nil
         # FIXME: Faker seems to create invalid phone numbers sometimes
-        contributor.whats_app_phone_number = Faker::PhoneNumber.cell_phone
+        contributor.whats_app_phone_number = Faker::PhoneNumber.unique.cell_phone_in_e164
```
mattwr18, line 61: These changes are attempting to address the issue with so many invalid phone numbers being created. it does not completely solve it. See comments in PR conversation.

spec/factories/organizations.rb
```diff
@@ -10,9 +10,9 @@
     onboarding_success_heading { File.read(File.join('config', 'locales', 'onboarding', 'success_heading.txt')) }
     onboarding_success_text { File.read(File.join('config', 'locales', 'onboarding', 'success_text.txt')) }
     onboarding_data_protection_link { 'https://tactile.news/100eyes-datenschutz/' }
-    signal_server_phone_number { Faker::PhoneNumber.cell_phone }
-    whats_app_server_phone_number { Faker::PhoneNumber.cell_phone }
-    telegram_bot_username { Faker::Internet.username }
+    signal_server_phone_number { Faker::PhoneNumber.unique.cell_phone_in_e164 }
+    whats_app_server_phone_number { Faker::PhoneNumber.unique.cell_phone_in_e164 }
+    telegram_bot_username { Faker::Internet.unique.username }
```
mattwr18, line 15: The first two lines are for consistency, even though we do not normalize nor do we validate on the organizaitons mode, yet. Maybe with the addition of our team adding these to the orgs, we should do so.

The last line seems to have gotten rid of the unique telegram bot errors. At least on 6 or so runs, it has not been encountered again.

spec/support/cuprite_setup.rb
```diff
@@ -37,7 +37,9 @@
       window_size: [1200, 800],
       browser_options: remote_chrome ? { 'no-sandbox' => nil } : {},
       inspector: true,
-      headless: !ENV['HEADLESS'].in?(%w[n 0 no false])
+      headless: !ENV['HEADLESS'].in?(%w[n 0 no false]),
+      process_timeout: 20,
+      timeout: 20
```
mattwr18, line 42: This is an attempt to deal with the `Ferrum::ProcessTimeout` errors that always happen on the first system spec. It seems to help, after adding this, I have not seen the error again, but it would be nice to see if there is some reason the first test takes more than 10 seconds at times.

spec/system/admin/auth_spec.rb
```diff
@@ -5,38 +5,24 @@
...
-      click_link 'Admin'
+    visit organization_dashboard_path(organization, as: admin)
+    click_link 'Admin'
 
-      expect(page).to have_http_status(:ok)
+    expect(page).to have_http_status(:ok)
 
-      expect(page).to have_link('New user')
-    end
+    expect(page).to have_link('New user')
```
mattwr18, line 26: This change just helps speed up the test suite by not having separate setup/teardown for the same basic test.

spec/system/dashboard/activity_notifications_spec.rb
```diff
@@ -49,26 +49,26 @@
...
       visit organization_dashboard_path(organization, as: user)
 
-      expect(page).to have_css('svg.Avatar-initials')
       expect(page).to have_text(
         "#{contributor_without_avatar.name} hat auf die Frage „#{request.title}” geantwortet."
       )
       expect(page).to have_text('vor einem Tag')
       expect(page).to have_link('Zur Antwort',
                                 href: organization_request_path(request.organization_id, request, anchor: "message-#{reply.id}"))
+      expect(page).to have_css('svg.Avatar-initials')
```
mattwr18, line 71: I haven't seen this failure often, but the idea is to check for text first. It's not clear to me whether the svg takes longer to render, or if `have_css` has a built-in wait functionality, or not.

spec/system/profile/index_spec.rb
```diff
@@ -159,7 +160,8 @@
     # valid_until set to 1 year from now
     expect(page).to have_content("Mindeslaufzeit: bis #{I18n.l(1.year.from_now, format: '%m/%Y')}")
 
-    Timecop.travel(6.months.from_now + 1.minute)
+    Timecop.travel(6.months.from_now + 3.hours.from_now)
+
```
mattwr18, line 164: I put this is a separate PR, but after re-running the test 4 times, I decided to include it here because the flakiness has become untenable

PR #1980, opened 2024-08-22
Draft: Feat/remove organization singleton

soey:


mattwr18 (approved):
Looks good! That was simpler than I expected. :tada:

PR #1981, opened 2024-08-22
fix: align wording for signal

soey:


config/locales/de.yml
```diff
@@ -88,7 +88,7 @@ de:
         - "Wenn Sie unser Signal-Konto gefunden haben, geben Sie diesen Code ein: <strong>%{token}</strong>"
     onboarding_signal_form:
       heading: Mit Signal teilnehmen
-      text: Geben Sie bitte unten die Handynummer ein, mit der Sie bei Signal registriert sind.
+      text: Geben Sie uns zunächst Ihre Einwilligung. Anschließend erhalten Sie einen Code zur Registrierung.
```
mattwr18, line 91:
```suggestion
      text: Geben Sie uns zunächst Ihr Einverständnis. Wir werden dann einen Code generieren, den Sie an uns senden können, um Ihre Registrierung abzuschließen.
```

We do not send them a code, we create a code, which they need to send to us. Maybe something like :point_up:

soey: ok .. maybe

PR #1982, opened 2024-08-22
Feat: Scoping todos for discussion

soey:
Hey @mattwr18, I roughly scanned through the application to find relations that might need to be scoped by the correct organization. Would be glad if we can talk about this.


app/controllers/messages/highlights_controller.rb
```diff
@@ -12,6 +12,7 @@ def update
     private
 
     def set_message
+      # TODO: scope by organization
```
mattwr18, line 15: I'm not sure about this one, finding a message by its id should not have any way to leak from an organization to another. On the other hand, if we assign the message correctly to the organization, we shouldn't have an issue finding it within the organizations messges.

app/controllers/contributors_controller.rb
```diff
@@ -24,6 +24,7 @@ def index
     @state = state_params
     @tag_list = tag_list_params
 
+    #TODO: scope by organization?
```
mattwr18, line 27: Implemented in https://github.com/tactilenews/100eyes/pull/1898/files#diff-1bd0a0fa1f76adfb12b7cc47fc7ab252b034c736ef7006a51b7353c10c581931R26-R28, but it's probably not easy to cherry pick the changes.

Still we should go through that PR since I caught a lot of these there.

app/controllers/contributors_controller.rb
```diff
@@ -58,6 +59,7 @@ def destroy
   end
 
   def count
+    # TODO: scope by organization?
```
mattwr18, line 62: I moved this to the requests controller in the other PR since it is only used by the request form to display a message of how many contributors would receive the message with said tags.

```rb
def contributors_count
  render json: { count: @organization.contributors.with_tags(params[:tag_list]).count }
end
```

app/controllers/contributors_controller.rb
```diff
@@ -69,10 +71,12 @@ def conversations
   private
 
   def set_contributor
+    # TODO: scope by organization?
```
mattwr18, line 74: Same as :point_up: in that I'm not sure that we need to scope it, but maybe we prefer to not have these lookups of the entire db?

app/controllers/contributors_controller.rb
```diff
@@ -69,10 +71,12 @@ def conversations
...
     @contributor = Contributor.find(params[:id])
   end
 
   def contributors_sidebar
+    # TODO: scope by organization
```
mattwr18, line 79: https://github.com/tactilenews/100eyes/pull/1898/files#diff-1bd0a0fa1f76adfb12b7cc47fc7ab252b034c736ef7006a51b7353c10c581931R71-R74

app/controllers/contributors_controller.rb
```diff
@@ -117,6 +121,7 @@ def state_params
   end
 
   def filtered_contributors
+    # TODO: scope by organization
```
mattwr18, line 124: https://github.com/tactilenews/100eyes/pull/1898/files#diff-1bd0a0fa1f76adfb12b7cc47fc7ab252b034c736ef7006a51b7353c10c581931R110-R119

app/activity_notifications/request_scheduled.rb
```diff
@@ -28,6 +28,7 @@ def group_message(notifications:)
       request_title: request.title,
       date: I18n.l(request.schedule_send_for, format: '%A, %-d.%-m.%Y'),
       time: I18n.l(request.schedule_send_for, format: '%H:%M'),
+      # probably scope by request.organization?
```
mattwr18, line 31: I don't think so as the activity notifications are scoped by the organization so only requests from that org appear in the dashboard. This is only accessing the time the request is scheduled for and displaying it.

We need to be careful, we don't have activity notifications of type request scheduled that have requests from other organizations, but other than that, at this level there is no need for additional scoping.

We create this type of notification in an after create commit here https://github.com/tactilenews/100eyes/blob/main/app/models/request.rb#L74 and we add the the organization id:

```rb
RequestScheduled.with(request_id: request.id, organization_id: request.organization.id).deliver_later(User.all)
```

Then in the dashboard controller, we only look for the notifications that are for that current user scoped by the organization id we have set. See https://github.com/tactilenews/100eyes/blob/main/app/controllers/dashboard_controller.rb#L16

Note that the :point_up: request scheduled would cause leaky notifications. See https://github.com/tactilenews/100eyes/pull/1983

app/controllers/messages/highlights_controller.rb
soey: better to scope it as if the users messes with the urls/requests messages from other organizations could be highlighted

app/components/move_message_form/move_message_form.rb
```diff
@@ -17,6 +17,7 @@ def request
     end
 
     def recent_requests
+      # TODO: scope by organization
```
soey, line 20: Done with: https://github.com/tactilenews/100eyes/pull/1988/files

app/controllers/contributors_controller.rb
soey: thnx. done with: https://github.com/tactilenews/100eyes/pull/1987

app/controllers/contributors_controller.rb
soey: I don't understand this. It's just adjusted with this PR https://github.com/tactilenews/100eyes/pull/1987 we can move it in a follow up

app/controllers/contributors_controller.rb
soey: It can be abused if user manipulates url. Fixed with: https://github.com/tactilenews/100eyes/pull/1987

app/controllers/contributors_controller.rb
soey: thnx ..done also in https://github.com/tactilenews/100eyes/pull/1987

app/controllers/contributors_controller.rb
soey: thnx. done with: https://github.com/tactilenews/100eyes/pull/1987

app/controllers/contributors_controller.rb
mattwr18: Sure, no problem. It just doesn't make much sense to me that it ever lived in the contributors controller when it is only ever used by the requests controller.
Will you open a PR to move it? Or shall I? Do you think it should just stay where it is for some reason I haven't thought of?

app/controllers/contributors_controller.rb
mattwr18: True, good point. This is where authorization comes in handy.

app/controllers/contributors_controller.rb
soey: Feel free to move it wherever it makes sense. I've not yet looked into this.

app/controllers/contributors_controller.rb
soey: done with: https://github.com/tactilenews/100eyes/pull/1987

app/controllers/dashboard_controller.rb
```diff
@@ -3,8 +3,11 @@
 class DashboardController < ApplicationController
   def index
     @activity_notifications = activity_notifications
```
soey, line 5: all done with https://github.com/tactilenews/100eyes/pull/1991

app/controllers/messages/requests_controller.rb
```diff
@@ -25,6 +25,7 @@ def update
     private
 
     def set_message
```
soey, line 27: https://github.com/tactilenews/100eyes/pull/1996

app/controllers/messages/highlights_controller.rb
soey: done with: https://github.com/tactilenews/100eyes/pull/1996

app/controllers/threema/webhook_controller.rb
```diff
@@ -57,7 +57,7 @@ def handle_callbacks
 
   def handle_delivery_receipt(delivery_receipt)
     return if delivery_receipt.message_ids.blank?
-
+    # TODO: scope by organization?
     messages = Message.where(external_id: delivery_receipt.message_ids)
```
soey, line 61: done with: https://github.com/tactilenews/100eyes/pull/1997

app/controllers/whats_app/webhook_controller.rb
```diff
@@ -149,6 +150,7 @@ def fetch_message_from_twilio(twilio_message_sid, organization)
 
     def handle_freeform_message_not_allowed_error(contributor, twilio_message_sid)
       message_text = fetch_message_from_twilio(twilio_message_sid, @organization)
+      # TODO: scope by organization?
```
soey, line 153: done with: https://github.com/tactilenews/100eyes/pull/1997

app/controllers/whats_app/webhook_controller.rb
```diff
@@ -173,7 +175,7 @@ def handle_unsuccessful_delivery
 
     def handle_successful_delivery
       return unless @contributor
-
+      # TODO: scope by organization
```
soey, line 178: done with: https://github.com/tactilenews/100eyes/pull/1997

app/jobs/broadcast_request_job.rb
```diff
@@ -13,6 +13,7 @@ def perform(request_id)
       return
     end
 
+    # TODO scope by request.organization
```
soey, line 16: done with: https://github.com/tactilenews/100eyes/pull/1998

app/jobs/broadcast_request_job.rb
```diff
@@ -21,6 +22,7 @@ def perform(request_id)
         request: request,
         broadcasted: true
       )
+      # TODO: scope by organization?
```
soey, line 25: can stay as it is..

app/models/contributor.rb
```diff
@@ -88,6 +88,7 @@ def name
     "#{first_name} #{last_name}"
   end
 
+  # TODO: Where is that used might have to be scoped by organization
```
soey, line 91: can stay as long as we stay with contributors belongs to one organization

app/models/request.rb
```diff
@@ -70,9 +70,11 @@ def messages_by_contributor
 
   def self.broadcast!(request)
     if request.planned?
+      # TODO: might need the organization id as well to scope properly
       BroadcastRequestJob.delay(run_at: request.schedule_send_for).perform_later(request.id)
```
soey, line 74: the job infers the organization from the request - so this can stay

app/models/request.rb
```diff
@@ -70,9 +70,11 @@ def messages_by_contributor
...
       RequestScheduled.with(request_id: request.id, organization_id: request.organization.id).deliver_later(User.all)
     else
+      # TODO: scope by request.organization
```
soey, line 77: done with https://github.com/tactilenews/100eyes/pull/1998

PR #1983, opened 2024-08-23
Fix leaky activity notification/add spec

mattwr18:


spec/support/shared_examples/activity_notifications.rb
```diff
@@ -1,18 +1,18 @@
...
 
     it 'for each user' do
       run_action(subject)
       recipient_ids = ActivityNotification.where(type: event_type).pluck(:recipient_id).uniq.sort
-      user_ids = User.pluck(:id).sort
-      expect(recipient_ids).to eq(user_ids)
+      user_ids = organization.users.pluck(:id)
+      admin_ids = User.admin.pluck(:id)
+      all_org_user_plus_admin = (user_ids + admin_ids).sort
+      expect(recipient_ids).to eq(all_org_user_plus_admin)
```
mattwr18, line 15: The changes to this file will cause other tests to break, but I think we should pass in the count that we expect to improve the readability and not hide what is going on with these counts.

Will fix the specs later.

mattwr18: I think it should be because it is the same for every notification. We create a notification for all users that are part of the organization and all system admin.

soey (approved):

PR #1986, opened 2024-08-23
feat: adds capybara for request specs

soey:
As @mattwr18 proposed according to https://thoughtbot.com/blog/faster-tests-with-capybara-and-request-specs to allow for testing response content on request specs


mattwr18 (approved):
Looks good to me!

PR #1987, opened 2024-08-23
Feat/contributor controller scoping

soey:
TDD on contributors controller scoping


soey:
Hey @mattwr18 and @roschaefer, I would like to have your opinion on some changes that are introduced in the branch:

1. To test the changes in the controller logic, I added specific matchers to the request specs (after aligning with @mattwr18 that that we want the changes to be tested as e2e as it makes sense)
2. I introduced a PORO data object for the existing index logic that could be alternatively or additionally be used to test against
3. I moved the contributors controller into the organizations module namespace (as this controller will only live underneath the organizations context)

It would be great if we could have a sync on these ideas and what are worth to carry on with when implementing the other needed changes.

spec/requests/contributors_spec.rb
```diff
@@ -10,23 +10,55 @@
...
     end
+
+    it 'contains only the contributors from the organization' do
+      expect(page).to have_content contributor.first_name
+
+      expect(page).to have_content('Aktiv 1')
+      expect(page).to have_content('Inaktiv 0')
+      expect(page).to have_content('Abbestellt 0')
+
+      expect(page).not_to have_content other_organizations_contributor.first_name
```
mattwr18, line 30: Why did you add the first_name to the let, but then not use this hard-coded value here?
I know @roschaefer favors using the hard-coded values. For me, I would expect either to give a hard-coded value that I then use to assert against, or don't give the value and use the `contributor.first_name` approach.

spec/requests/contributors_spec.rb
```diff
@@ -10,23 +10,55 @@
...
+      let!(:other_organizations_contributor) { create(:contributor, first_name: 'WhatAName') }
+
+      it 'renders not found' do
+        get organization_contributor_url(organization, create(:contributor), as: user)
+        expect(response).to be_not_found
+      end
+
+      it 'doesn\'t have the other contributor in the sidebar' do
+        subject.call
+        expect(page).not_to have_content other_organizations_contributor.first_name
```
mattwr18, line 52: same as :point_up:

mattwr18 (approved):
Looks good to me. See my comments.
Unfortunately, manual testing is difficult because we have so many other controllers that are unscoped so I run into bugs that I looked into and I don't believe have anything to do with this PR, but we'll have to do a massive manually testing after we think we've gotten everything.

app/controllers/organizations/contributors_controller.rb
```diff
@@ -0,0 +1,115 @@
...
+                                .with_attached_avatar
+    end
+
+    def toggle_active_state
+      if ActiveModel::Type::Boolean.new.cast(toggle_active_state_params[:active])
+        @contributor.reactivate!
+      else
+        @contributor.deactivate!(user_id: current_user.id)
+      end
+    end
```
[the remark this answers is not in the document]
soey: if we ever go that way

spec/requests/contributors_spec.rb
soey: relying on faker for the names makes the test flaky (i've encountered an error once).. and for me it is most descriptive this way.. (it should not contain the first name of the other contributor) whereas having only the name hardcoded would need the reader to reverse lookup the value to understand what it belongs to.

app/controllers/organizations/contributors_controller.rb
```diff
@@ -0,0 +1,115 @@
...
+    before_action :contributors_params, only: :index
+
+    def message
+      request = if message_params[:reply_to_id].present?
+                  reply_to = contributor.replies.find(message_params[:reply_to_id])
+                  reply_to.request
+                else
+                  contributor.active_request
+                end
+      render(plain: 'No active request for this contributor', status: :bad_request) and return unless request
```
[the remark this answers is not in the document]
soey: for me there is to much of application logic here to render a 404 .. i would keep the 500 but possible improve the wording

PR #1991, opened 2024-08-27
Feat/scope dashboard controller

soey:


app/controllers/organizations/dashboard_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      active_contributors_count = @organization.contributors.active.count
+      return 0 unless active_contributors_count.positive?
+
+      activity = (current_weeks_activity / active_contributors_count.to_f * 100)
+
+      return 100 if activity > 100
+
+      activity
+    end
+  end
```
mattwr18, line 45: I know you mentioned it, but is there a reason you decided not to follow the pattern you introduced and extract these into a PORO?

spec/requests/dashboard_spec.rb
```diff
@@ -2,27 +2,54 @@
...
+    it 'counts the requests for the organization' do
+      create_list(:request, 3, organization: organization)
+      create(:request)
+      subject.call
+      expect(page).to have_content '3 Fragen gestellt'
+    end
+
+    it 'counts the replies for the organization' do
+      create_list(:message, 3, organization: organization)
+      create(:message) # other message
```
mattwr18, line 41:
```suggestion
      create(:message, :inbound) # other message
      create(:message, :outbound) # shouldn't be counted
```
This is inconsistent with lines 50-51... By default, the message, from FactoryBot, is created to be inbound/ a reply so it can be omitted here and it be counted by the engagement metric. On lines 50-51, you did not omit the inbound trait and favored being more explicit.

I would favor being explicit here. I am the one that added that the message is inbound by default and I still needed to look at the factories definition to make sure what was happening here.

mattwr18 (approved):
See my comments on how I think we can improve one of the tests, but let's keep it moving :fire:

app/controllers/organizations/dashboard_controller.rb
soey: Just to much other work to do. We can do this always later on if we want more tests for that thing.

spec/requests/dashboard_spec.rb
soey: yes thnx, being as self explanatory as possible is good.  Then we should also change line 40 to be the same. Just followed what has been there...

PR #1996, opened 2024-08-28
feat: organizations messages only

soey:
fix: adjust specs

feat: adjust specs and get message request scoped

fix: rebase artefacts


app/controllers/messages_controller.rb
```diff
@@ -34,17 +34,17 @@ def update
...
 
   def set_message
-    @message = Message.find(params[:id])
+    @message = @organization.messages.find(params[:id])
   end
 
   def set_request
-    request_id = params[:id] ? Message.find(params[:id]).request_id : params[:request_id]
-    @request = Request.find(request_id)
+    request_id = @message ? @organization.messages.find(params[:id]).request_id : params[:request_id]
```
mattwr18, line 41:
```suggestion
    request_id = @message ? @message.request_id : params[:request_id]
```

if you have already found the message record and saved it to an instance variable and you use that instance variable in the first part of the condition, why not use it to save the request_id, this would save a db call.

spec/requests/messages_request_spec.rb
```diff
@@ -17,6 +17,22 @@
...
+
+      it 'renders unauthorized if not logged in' do
+        patch(organization_message_highlight_url(message, organization_id: message.organization_id, format: :json), params: params)
+        expect(response).to be_unauthorized
+      end
+
+      it 'renders not found for other users' do
+        patch(organization_message_highlight_url(message, organization_id: message.organization_id, format: :json, as: create(:user)),
+              params: params)
+        puts response.code
```
mattwr18, line 31:
```suggestion
        
```

spec/requests/messages_request_spec.rb
```diff
@@ -17,6 +17,22 @@
...
+        expect(response).to be_not_found
+      end
+    end
+
```
mattwr18, line 35: you added a shared example in the dashboard controllers PR, then in subsequent PRs here and https://github.com/tactilenews/100eyes/pull/1987/files#diff-b1db8ca954f27e6686bdee8dd73430451cca3d09ff0e6abfc8ba3b3c67b3eafdR80-R84, you decided not to use them.
Why?

spec/requests/messages_request_spec.rb
```diff
@@ -46,33 +62,57 @@
...
+          let(:message) { create(:message, sender: contributor, recipient: nil, organization: organization) }
+
+          it 'renders successfully' do
+            expect(response).to be_successful
+            expect(response.body).to include('Zora Zimmermann')
+          end
+        end
+
+        context 'given an outbound message' do
+          let(:message) { create(:message, sender: nil, recipient: contributor, broadcasted: true, organization: organization) }
```
mattwr18, line 93:
```suggestion
          let(:message) { create(:message, sender: user, recipient: contributor, broadcasted: true, organization: organization) }
```
Let's not add new tests with sender nil. While this is valid, for now, this comes from a time where we did not add the user as the sender and after some time, we will remove this possibility. Since the feature was introduced, we always add the current_user as the message sender.

spec/requests/messages_request_spec.rb
```diff
@@ -46,33 +62,57 @@
...
+            expect(response.body).to include('Zora Zimmermann')
+          end
+        end
+      end
 
-      it 'renders successfully' do
-        expect(response).to be_successful
-        expect(response.body).to include('Zora Zimmermann')
+      context 'with a message of another organization' do
+        let(:message) { create(:message, sender: nil, recipient: contributor, broadcasted: true) }
```
mattwr18, line 103: same as :point_up:

spec/requests/messages_request_spec.rb
```diff
@@ -46,33 +62,57 @@
...
+        it 'renders not found' do
+          get(organization_message_request_url(organization.id, message, as: user))
+          expect(response).to be_not_found
+        end
```
mattwr18, line 107: same as :point_up: ... why not use the shared example?

spec/requests/messages_spec.rb
```diff
@@ -45,33 +45,64 @@
...
-    it 'shows success notification' do
-      subject.call
-      expect(flash[:success]).not_to be_empty
-    end
+      subject { -> { patch organization_message_url(organization, message, as: user), params: { message: new_attrs } } }
 
-    it 'redirects to the conversation link the message belongs to' do
-      subject.call
-      expect(response).to redirect_to(message.chat_message_link)
+      it 'renders a notfound' do
```
mattwr18, line 55:
```suggestion
      it 'renders not found' do
```

shared example?

spec/requests/messages_spec.rb
```diff
@@ -45,33 +45,64 @@
...
-    context 'not manually created message' do
-      let(:message) { create(:message, creator_id: nil, request: request) }
+    context 'with a message of the organization' do
+      let(:message) { create(:message, creator_id: user.id, text: previous_text, request: request, sender: contributor) }
 
-      it 'does not update the requested message' do
+      subject { -> { patch organization_message_url(message.organization, message, as: user), params: { message: new_attrs } } }
+
+      it {
+        puts response.inspect
```
mattwr18, line 67:
```suggestion

```

mattwr18 (commented):
You haven't asked for my review, so take the early feedback for what it's worth

spec/requests/messages_request_spec.rb
soey: good to know. thnx.

app/controllers/messages_controller.rb
soey: Ouch .. yes. I really scratched my head over this part and was on it to change it properly .. I think its way better with the last commit now

spec/requests/messages_spec.rb
soey: Because the shared spec is only for unpermitted users. Here the not found is caused by a message not belonging to the org

spec/requests/messages_request_spec.rb
soey: Because the shared spec is only for unpermitted users. Here the not found is caused by a message not belonging to the org

spec/requests/messages_request_spec.rb
soey: Because the shared spec is only for unpermitted users. Here the not found is caused by a message not belonging to the org

spec/requests/messages_request_spec.rb
soey: Well yes, good question. They are not very usable as on the first case it's a JSON response with other behavior and expectations. Other case is not about users at all. Maybe the shared specs are DRYed too early.. let's see if they are of use at other place.. if not I'm going to remove them.

spec/requests/messages_request_spec.rb
mattwr18: I think the shared example, if we keep it, should cover any record not belonging to the organization.

See https://github.com/tactilenews/100eyes/pull/1999/files#diff-71c5d728e921a45b367bd360e1bbeb7333547d2b23b5f1105ca6b72a941e3ec2R102-R108

soey: mhh... maybe .. but then there is not really worth the indirection of a shared example.. i would then inline everything

mattwr18: Ok, I will do the same then

app/controllers/messages_controller.rb
mattwr18: I agree. I think it's an improvement. Thanks :blush:

spec/requests/errors_spec.rb
```diff
@@ -25,7 +25,7 @@
   describe '500 internal server error' do
     subject { -> { patch organization_message_url(organization, message, as: user) } }
 
-    let(:message) { create(:message, creator_id: user.id) }
+    let(:message) { create(:message, creator_id: user.id, organization: organization) }
```
mattwr18, line 28: I didn't know that you could create a resource by passing in the association like this if you delegate to another resource.
I knew you could use the delegate to look it up, but didn't realize you could also create it like this. Cool :sunglasses:

spec/requests/messages_request_spec.rb
```diff
@@ -46,33 +62,57 @@
...
+    it_behaves_like 'unauthenticated' do
+      before { get(organization_message_request_url(message.organization, message)) }
+    end
 
-      it 'renders successfully' do
-        expect(response).to be_successful
-        expect(response.body).to include('Zora Zimmermann')
-      end
+    it_behaves_like 'protected' do
+      before { get(organization_message_request_url(message.organization, message, as: create(:user))) }
```
mattwr18, line 75: I thought we agreed to remove the shared examples...

spec/requests/messages_spec.rb
mattwr18: there is still a type here where there is no space between not and found

spec/requests/messages_spec.rb
```diff
@@ -41,37 +41,83 @@
...
+        let(:message) { create(:message, creator_id: nil, request: request) }
+
+        it 'does not update the requested message' do
+          subject.call
+          expect(response).not_to be_successful
+        end
+
+        it 'shows error notification' do
+          subject.call
+          expect(flash[:error]).not_to be_empty
```
mattwr18, line 119: I have been refactoring these in my other PR to make use of the new `CapybaraPage` helper you added. So instead of just checking that the flash is not empty, we can check that the page has content xyz
See https://github.com/tactilenews/100eyes/pull/1999/files#diff-71c5d728e921a45b367bd360e1bbeb7333547d2b23b5f1105ca6b72a941e3ec2R175-R177
Do you like that change? It's more explicit which I know you favor in other cases.

mattwr18 (approved):
Little suggestions and still have a typo, but I'll approve as there is nothing critical that should prevent this from getting merged.

spec/requests/messages_spec.rb
mattwr18: Ok, it doesn't quite work like that because the `response.body` is the information that you are being redirected. I still updated the test to check that the flash success message contained the exact string, which contains the number of contributors I expect to receive the message based on belonging to the organization.

PR #1997, opened 2024-08-28
feat: scope the webhook controllers finder

soey:


app/controllers/whats_app/webhook_controller.rb
```diff
@@ -149,7 +149,7 @@ def fetch_message_from_twilio(twilio_message_sid, organization)
 
     def handle_freeform_message_not_allowed_error(contributor, twilio_message_sid)
       message_text = fetch_message_from_twilio(twilio_message_sid, @organization)
-      message = Message.find_by(text: message_text)
+      message = @organization.messages.find_by(text: message_text)
```
mattwr18, line 152: Looking through the code in this webhook, I think we could refactor the use of guard clauses, but it has nothing to do with the scope of this PR

spec/requests/threema/webhook_spec.rb
```diff
@@ -99,15 +99,20 @@
...
-          it 'updates all messages' do
+          it 'updates all messages belonging to the organization' do
             expect { subject }.to change { messages.first.reload.read_at }.from(nil).to(kind_of(ActiveSupport::TimeWithZone)).and \
               change { messages.second.reload.read_at }.from(nil).to(kind_of(ActiveSupport::TimeWithZone)).and \
                 change { messages.third.reload.read_at }.from(nil).to(kind_of(ActiveSupport::TimeWithZone))
           end
+
+          it 'doesn\'t update the other message' do
+            expect { subject }.not_to(change { other_message.reload.read_at })
+          end
```
mattwr18, line 115: I don't know how I feel about this test. I like the fact that it is testing that the other_message does not get updated as it does not belong to the organization, but I don't know how we could receive a message_id in the list of message_ids that belong to the threemarb_api_identity if the message does not belong to the organization that has that identity. If that were to happen, somehow, I would consider that a bug and would want to have an error thrown so I could look into what has happened.

mattwr18 (approved):
Question :point_down:

spec/requests/threema/webhook_spec.rb
soey: I took it the other way: If we are sure there is no way of doing so we could omit this. I was not 100% sure  (maybe tampering the payload and sending the request manually) - so better have it in place - it doesn't do any harm.

mattwr18: The payload is encrypted with secrets that only we know. If they are exposed, we would have bigger problems. However, I don't think it hurts, it's just that I would like to be active about alerting us to danger instead of ignoring it. I'm sure we've had this discussion and you favor not silently ignoring the problem.

PR #1998, opened 2024-08-28
feat: scope finder for contributors in broadcast request jobs and request model

soey:


mattwr18 (approved):
Super straightforward, besides that weird situation with the tests earlier and the fact that we probably need to refactor the `broadcast!` method, which I've started in https://github.com/tactilenews/100eyes/pull/1946

PR #1999, opened 2024-08-28
[fix] scope requests controller to organization

mattwr18:


app/controllers/requests_controller.rb
```diff
@@ -36,24 +36,19 @@ def create
...
-        }
-      else
-        redirect_to organization_request_path(@request.organization_id, @request),
-                    flash: { success: I18n.t('request.success', count: @request.stats[:counts][:recipients]) }
-      end
+      redirect_to organization_requests_path(@request.organization_id, filter: :planned), flash: {
+        success: I18n.t('request.schedule_request_success',
+                        count: @request.organization.contributors.active.with_tags(@request.tag_list).count,
+                        scheduled_datetime: I18n.l(@request.schedule_send_for, format: :long))
+      }
```
mattwr18, line 51: I removed the conditional here because the first thing we do before hitting the update method is ensure that the request is planned for the future. We disallow editing for requests that have already been broadcast.

spec/requests/requests_spec.rb
```diff
@@ -160,25 +420,188 @@
...
+            request.reload
+          end
+
+          it 'displays the stats' do
+            subject.call
+            expect(page).to have_content('6/8')
+            expect(page).to have_content('18')
+            expect(page).to have_content('10')
+          end
+        end
```
mattwr18, line 604: I'm not sure about this test. This is a lot of overhead to check for these numbers. What do you think @soey 
The only reason this is a separate action in the controller is because we lazy load these stats to improve the time it takes to render the requests index page. These stats calculations are quite expensive so we moved them to be calculated on demand.

app/controllers/requests_controller.rb
soey: Is it possible to update a former planned request to be sent now? In that case that might still have to stay...

spec/requests/requests_spec.rb
```diff
@@ -3,85 +3,243 @@
...
 
-    context 'with image file(s)' do
-      let(:params) do
-        { request: { title: 'Message with files', text: 'Did you get this image?', files: [fixture_file_upload('profile_picture.jpg')] } }
+    context 'authenticated and authorized' do
+      let(:user) { create(:user, organizations: [organization]) }
+
+      before do
+        create(:request, title: 'I belong to the organization', organization: organization)
+        create(:request, title: 'I do too!', organization: organization)
```
soey, line 37: I would propose to hold the specs as concise and short as possible. In this regard: Can we remove any of these stubs without loosing sth. logic wise?

spec/requests/requests_spec.rb
```diff
@@ -3,85 +3,243 @@
...
-          request = Request.last
-          expect(response).to redirect_to organization_request_path(organization, request)
+      it 'should be successful' do
+        subject.call
+        expect(response).to be_successful
+      end
+
+      context 'no params' do
+        it 'displays only sent messages' do
+          expect(page).to have_content('I belong to the organization')
```
soey, line 52: as this is about sent messages, it would be nice to directly see that this is a sent message (which is currently implicitly known) - and I think it's enough to have 2 expectations (negative/positive) in the block.

spec/requests/requests_spec.rb
```diff
@@ -3,85 +3,243 @@
...
+          expect(page).to have_content('I do too!')
+          expect(page).not_to have_content('Planned for the future')
         end
 
-        it 'shows success notification' do
-          subject.call
-          expect(flash[:success]).not_to be_empty
+        it 'display only requests from the organization' do
+          expect(page).to have_content('Gestellt 2')
```
soey, line 58: Can this be reduced to 2 expectations... or is there sth. else tested here?

spec/requests/requests_spec.rb
```diff
@@ -3,85 +3,243 @@
...
       end
 
-      describe 'multiple image files' do
-        let(:params) do
-          { request: { title: 'Message with files', text: 'Did you get this image?',
-                       files: [fixture_file_upload('profile_picture.jpg'), fixture_file_upload('example-image.png')] } }
+      context 'sent filter param' do
+        let(:params) { { filter: :sent } }
+
+        it 'displays only sent messages' do
```
soey, line 69: does it behave the same as without parameters? Maybe a shared?

spec/requests/requests_spec.rb
```diff
@@ -3,85 +3,243 @@
...
+          expect(page).to have_content('I do too!')
+          expect(page).not_to have_content('Planned for the future')
         end
+      end
 
-        it 'redirects to requests#show' do
-          response = subject.call
-          request = Request.last
-          expect(response).to redirect_to organization_request_path(organization, request)
+      context 'planned filter param' do
```
soey, line 76: wording

spec/requests/requests_spec.rb
```diff
@@ -3,85 +3,243 @@
...
+        let(:params) { { filter: :planned } }
+
+        it 'displays only sent messages' do
```
soey, line 79: wording of

spec/requests/requests_spec.rb
```diff
@@ -3,85 +3,243 @@
...
+          expect(page).to have_content('Gestellt 2')
+          expect(page).to have_content('Geplant 1')
+          expect(page).to have_content('Planned for the future')
+          expect(page).not_to have_content('Also planned, but from another organization')
         end
       end
+    end
+  end
+
+  describe 'GET /{organization_id}/requests' do
```
soey, line 95:
```suggestion
  describe 'GET /:organization_id/request/:id' do
```

spec/requests/requests_spec.rb
```diff
@@ -3,85 +3,243 @@
...
-          expect(flash[:success]).not_to be_empty
+        end
+
+        context 'with image file(s)' do
+          let(:params) do
+            { request: { title: 'Message with files', text: 'Did you get this image?',
+                         files: [fixture_file_upload('profile_picture.jpg')] } }
+          end
+
+          describe 'an image file' do
```
soey, line 203:
```suggestion
          describe 'one image file' do
```
Is there a significant difference btw. one and many image files that justifies its own block & specs

spec/requests/requests_spec.rb
```diff
@@ -3,85 +3,243 @@
...
-RSpec.describe 'Requests', telegram_bot: :rails do
+RSpec.describe 'Requests', type: :request do
   let(:organization) { create(:organization) }
 
-  describe 'POST /{organization_id}/requests' do
-    before(:each) { allow(Request).to receive(:broadcast!).and_call_original } # is stubbed for every other test
-    subject { -> { post organization_requests_path(organization, as: user), params: params } }
-    let(:params) { { request: { title: 'Example Question', text: 'How do you do?', hints: ['confidential'] } } }
-    let(:user) { create(:user, organizations: [organization]) }
+  describe 'GET /{organization_id}/requests' do
```
soey, line 9:
```suggestion
  describe 'GET /:organization_id/requests' do
```

spec/requests/requests_spec.rb
```diff
@@ -3,85 +3,243 @@
...
+      context 'request is part of organization' do
+        it 'should be successful' do
+          subject.call
+          expect(response).to be_successful
         end
+      end
+    end
+  end
+
+  describe 'POST /{organization_id}/requests' do
```
soey, line 139:
```suggestion
  describe 'POST /:organization_id/requests' do
```

spec/requests/requests_spec.rb
```diff
@@ -105,53 +263,155 @@
...
+              "Ihre Frage wurde erfolgreich geplant, um am #{I18n.l(request.schedule_send_for,
+                                                                    format: :long)} an 2 Community-Mitglieder gesendet zu werden."
+            )
+          end
+        end
       end
     end
+  end
+
+  describe 'DELETE /{organization_id}/requests/:id' do
```
soey, line 340:
```suggestion
  describe 'DELETE /:organization_id/requests/:id' do
```

spec/requests/requests_spec.rb
soey: I would say: If it's not tested it's probably broken. So I think testing is justified. My problem with that is that it's hard to understand how the numbers refer to the test setup. I would probably add more context to it. If you feel there is too much logic testing on the request then it could be justified to extract a data object into a PORO and test it encapsulated in a model spec.

soey (commented):
Quite interesting in how far the request specs escalate with the authentication and authorization - whereas this seems to be justified & needed from the functionality perspective it bothers me somehow that we need to repeat it all over again. Overall the functionality side is all good. I would propose to revise the specs and make them as concise and understandable as possible.

spec/requests/requests_spec.rb
mattwr18: No.

spec/requests/requests_spec.rb
mattwr18: It's tested what it says in the description. That it only displays requests from the organization.
It tests that the counts are scoped to the organization by setting up 2 planned requests, one of which is from the organization, and 3 sent requests, 2 of which are from the organization.

It asserts that the counts are correct and that only the requests which belong to the organization are displayed in the list.

spec/requests/requests_spec.rb
mattwr18: I don't think adding a shared for one duplication would be the way to go.

spec/requests/requests_spec.rb
mattwr18: don't know what this means. I added the wording that makes sense to me.

spec/requests/requests_spec.rb
mattwr18: same

spec/requests/requests_spec.rb
mattwr18: I added these specs 2 years ago. Looking the logic, it doesn't appear to test anything different, so I would say it's not needed. Since we are doing such a big refactor of these specs anywhere, I'll remove on of them vs doing it in a follow-up PR.

mattwr18: As far as your suggestion to improve the English. This is not an improvement as _an_ can only mean _one_

spec/requests/requests_spec.rb
mattwr18: You added this code in https://github.com/tactilenews/100eyes/pull/1972/commits/9783770b417694d6fcd0f1f9f37c0f606edea729#diff-71c5d728e921a45b367bd360e1bbeb7333547d2b23b5f1105ca6b72a941e3ec2R9

I favor using the rails annotation for these, which is why I've done that for the description I added. I don't think it's productive to try to force my styling on others. If you have changed your mind and prefer the annotation I added over the one you added, please feel free to update it going forward, or retroactively.

spec/requests/requests_spec.rb
mattwr18: same as :point_up:

spec/requests/requests_spec.rb
mattwr18: same as :point_up:

spec/requests/requests_spec.rb
mattwr18: Well, the request stats are tested in https://github.com/tactilenews/100eyes/blob/main/spec/models/request_spec.rb#L204-L275

The setup is copy/pasted from there. I agree it's not super intuitive. This is also only testing it at a unit level and is slightly different because here we test that the inline metrics display the "contributors" who replied of those that received as a division, ie `6/8` and is end-to-end.

We could simplify it to send the message to two recipients, one of which responded several times, one with a photo.

spec/requests/requests_spec.rb
soey: I just wanted to have it consistently and are willing to go either way. Just found to formats in the file and proposing the style that i found simpler to type at that moment

app/controllers/requests_controller.rb
mattwr18: Good catch! The only thing that is different is the success_message and the filter, but it is not good UX to be redirected to the incorrect path so I've added back complexity to improve the UX.

I've also added a test to not forget the purpose of the complexity in the future. See 7131b583

soey (approved):

PR #2000, opened 2024-08-29
[fix] Account for daylight savings when travelling to the future

mattwr18:


mattwr18:
Implemented in https://github.com/tactilenews/100eyes/pull/1973 out of frustration for not being able to get the test suite to pass for the flakiness of the tests.

PR #2002, opened 2024-08-29
Install passlib in runner

mattwr18:


mattwr18:
@roschaefer I finally got this working thanks to https://stackoverflow.com/questions/78525098/ansible-unable-to-encrypt-nor-hash-passlib-must-be-installed and https://github.com/dawidd6/action-ansible-playbook/issues/98

@soey there is nothing to review except for the actions history, where I invoked the deploy job by changing that it be run on every PR, got it to deploy, then reverted the change so it will only run after push to main

roschaefer:
Nice! Why so many of these?

image

PR #2004, opened 2024-09-01
feat: scope contributors channel uniqueness

soey:


app/models/contributor.rb
```diff
@@ -39,11 +39,12 @@ class Contributor < ApplicationRecord
...
-  validates :threema_id, uniqueness: { case_sensitive: false }, allow_blank: true
-  validates :telegram_id, uniqueness: true, allow_nil: true
-  validates :signal_phone_number, uniqueness: true, allow_nil: true
-  validates :whats_app_phone_number, uniqueness: true, allow_nil: true
+  validates :email, uniqueness: { case_sensitive: false, scope: :organization }, allow_nil: true, 'valid_email_2/email': true
+  validates :threema_id, uniqueness: { case_sensitive: false, scope: :organization }, allow_blank: true
+  validates :telegram_chat_id, uniqueness: { scope: :organization }, allow_nil: true
+  validates :telegram_id, uniqueness: { scope: :organization }, allow_nil: true
+  validates :signal_phone_number, uniqueness: { scope: :organization }, allow_nil: true
+  validates :whats_app_phone_number, uniqueness: { scope: :organization }, allow_nil: true
```
mattwr18, line 47:
```suggestion
  validates :whats_app_phone_number, uniqueness: { scope: :organization_id }, allow_nil: true
```

It seems to work with organization, but I'd consider using the `organization_id` since all the docs I can find on this scoping validation use the `_id`

spec/models/contributor_spec.rb
```diff
@@ -3,6 +3,22 @@
...
 
 RSpec.describe Contributor, type: :model do
+  shared_examples 'unique within an organization' do
+    before { create(:contributor, organization: organization, **attrs) }
+
+    it 'raises errors when not unique' do
+      expect { create(:contributor, organization: organization, **attrs) }.to raise_error(ActiveRecord::RecordInvalid)
+      expect do
+        build(:contributor, organization: organization, **attrs).save!(validate: false)
+      end.to raise_error(ActiveRecord::RecordNotUnique)
```
mattwr18, line 13: Why do you feel the need to check the `.save(validate: false)` here?
I assume what is happening here is that by adding validate false, you are bypassing the model unique validations and you are testing that the database index unique validations are preventing a duplicate record from being created even if we were to attempt to do that from the rails console, is that right?

mattwr18 (changes requested):
I would expect to have request specs that show that signing up with the same `attr_name => attr_value` to a different organization creates a new contributor.

spec/models/contributor_spec.rb
soey: Well basically I would have never thought of it, if not finding it on the existing code. What it does though is basically checking if there is a unique index properly set up - as you concluded.

app/models/contributor.rb
soey: I don't think that it improves something .. but let's get it merged: https://github.com/tactilenews/100eyes/pull/2004/commits/14510e3904cf0f78de52661b8277a7e09b0f4252

soey:
> I would expect to have request specs that show that signing up with the same `attr_name => attr_value` to a different organization creates a new contributor.

I'm not 100% sure what you mean by that. If you are referring to the uniqueness scope, this is already tested in the model spec L16-19 and we should not double this in the request spec:

`    it 'accepts duplicates for other organizations' do
      expect { create(:contributor, **attrs) }.not_to raise_error
      expect { build(:contributor, **attrs).save!(validate: false) }.not_to raise_error
    end`

mattwr18:
I mean that a model spec only tests it at the unit level, it does not test the implementation in the controller, which means it's untested. You recently commented that untested code is broken, so I would expect the code to be tested.

soey:
> I mean that a model spec only tests it at the unit level, it does not test the implementation in the controller, which means it's untested. You recently commented that untested code is broken, so I would expect the code to be tested.

If there is relevant responsibility & code in the controller then it should be tested in the request spec. This is e.g. relevant if you change scopes e.g. from `Contributor.all` to `@organization.contributor.all` in the controller. Then a possible error would be in the controller and the request spec should cover the correct scoping.

In this case it's only about the correct validation of the model. The responsibility lies solely in the model and though the testing should also be kept in the model as it is.

If you work TDD - implementation changes in the controller would need upfront spec expectations in the request spec. If you do changes in the model you would start with changing model specs.

mattwr18:
You didn't just change the model, you also changed the `contributor_exists?` in the controller. This is untested.

mattwr18:
I've added a spec for WhatsApp onboarding here https://github.com/tactilenews/100eyes/pull/2010
I don't know whether it's worth it to have a shared example for all messengers create actions, but probably.
Also, while adding this test, it became clear to me that the change to the `contributor_exists?` is not necessary. The test passes with or without it as there is no uniqueness thrown for a attr_name => attr_value pair unless it is the same organization. Then, it fails the model validations and the error bubbles up to this method.

mattwr18:
> I've added a spec for WhatsApp onboarding here #2010 I don't know whether it's worth it to have a shared example for all messengers create actions, but probably. Also, while adding this test, it became clear to me that the change to the `contributor_exists?` is not necessary. The test passes with or without it as there is no uniqueness thrown for a attr_name => attr_value pair unless it is the same organization. Then, it fails the model validations and the error bubbles up to this method.

Ok, I can see that removing the organization from the `contributor_exists?` serves the purpose of revealing the error that would be thrown if one tried to save the record and we can then redirect to success instead of allowing the error to be thrown.

soey:
replaced with: https://github.com/tactilenews/100eyes/pull/2010

PR #2006, opened 2024-09-02
Fix migrations in ansible installation playbook

mattwr18:
For some reason, running the schema migrations and data migrations together causing them to fail and requires an addtional step to ssh into every instance and run the data migrations after the schema migrations have run successfully. The data migrations many times rely on the schema being migrated, but it was my understanding that the db:migrate:with_data accounted for this. In practice, however, it reliably fails every time.


mattwr18:
@soey also nothing to review. This blows up when I deploy so one needs to add a data migration and then deploy to test.

PR #2008, opened 2024-09-04
Favor links on organizations page

mattwr18:


app/components/organizations_index/organizations_index.erb
```diff
@@ -0,0 +1,22 @@
...
+      <% organizations.each do |organization| %>
+        <li>
+          <%= c 'button', link: organization_dashboard_path(organization), styles: [:plain, :underline] do %>
+            <%= organization.name %>
+          <% end %>
+        </li>
+      <% end %>
+    </ul>
+  <% end %>
+<% end %>
```
mattwr18, line 22: @soey As we've discussed, I'm not really a fan of having a `_index` component simply for the fact of adding a css to the component level structure, as we do here.
I could go through my old PRs where I introduced adding css to views and we could change this here, but honestly, I think it's not related to this PR.
We could discuss this pattern and change it for the contributors_index as well in another PR. I'm not sure if there are other places that follow this pattern.

app/controllers/organizations_controller.rb
```diff
@@ -5,11 +5,6 @@ class OrganizationsController < ApplicationController
   layout 'minimal'
 
   def index
-    @organizations = Organization.all
-  end
-
-  def set_organization
-    organization = Organization.find(params[:organization_id])
-    redirect_to organization_dashboard_path(organization)
+    @organizations = current_user.admin? ? Organization.all : current_user.organizations
```
mattwr18, line 8: this fixes a bug that we do not want to introduce.

spec/requests/organizations_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+        it 'displays all organizations' do
+          Organization.find_each do |organization|
+            expect(page).to have_css("a[href='#{organization_dashboard_path(organization)}']", text: organization.name)
+          end
+        end
+      end
+    end
+  end
+end
```
mattwr18, line 50: adding request specs for this route

app/components/organizations_index/organizations_index.erb
soey: I very much like the idea to KISS

soey (approved):

PR #2010, opened 2024-09-04
Tdd contributor uniqueness

mattwr18:


soey:
@mattwr18 this lgtm wdyt?

mattwr18:
@soey I was trying to remove the unrelated changes removing the telegram_chat_id from the changes here and did an interactive rebase to squash some commits. Unfortunately, the changes remained and I was confused, until I realized that even though the changes had been merged they were not in main. I went back and saw that you used this PR as the base of the changes. I don't know if this was purposeful, but I think we should avoid this. Of course, if the changes are related and we want to merge small PRs into a release branch as we've done before, I think it's ok, but the changes were completely unrelated and cost time and will be associated with this PR even though they are not related.

I also added a test to the telegram webhook spec to show that the contributor who is completing their onboarding with the same telegram_id as another that is in another relationship works as expected. I wanted to add this line of code to the receive_polling_job_spec, but it's more complicated because it involves re-recording a VCR for the receive messages from the other organizations signal server phone number. Since, it is the same as Telegram, I decided not to spend more time now.

We can always extend it, if we decide.

PR #2011, opened 2024-09-04
Test drive successfully updating user's organizations

mattwr18:


spec/requests/admin/user_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+        expect { subject.call }.to (change { user.reload.organizations }).from([organization, another_organization]).to([organization])
+      end
+    end
+
+    context 'Deactivating user' do
+      let(:params) { { user: { active: false } } }
+
+      it "updates the user's active status" do
+        expect { subject.call }.to (change { user.reload.active? }).from(true).to(false)
+      end
```
mattwr18, line 37: I added this test here because of the simplification to the controller and to ensure that it did not break this implementation.
I considered removing the system spec all together in favor of request specs, but wanted to keep this as small as possible.

soey (approved):

PR #2023, opened 2024-09-11
fix: seeds

roschaefer:
Motivation
----------
Fix seeds which apparently broke in #1971.

How to test
-----------
1. `docker compose up db`
2. `bin/rails db:drop db:create db:migrate db:seed`
3. Does not fail anymore


soey (approved):

mattwr18:
Thanks @roschaefer ... I knew there was probably something broken, but didn't take the time to look into it.

PR #2024, opened 2024-09-11
refactor: remove obsolete `rails-settings-cached`

roschaefer:
Motivation
----------
With multi-tenancy, `rails-settings-cached` became obsolete. Let's remove it.

How to test
-----------
1. Test the app as usual, there should be no changes in behaviour


mattwr18 (approved):
I added a commit removing the now non-exsistent path from the `rubocop_todo.yml`, but other than that, I think it looks good! Big thanks! Really happy this is taken care of :smiley_cat: 

While I was looking for any settings references with `git grep`, I realized that we are still using the term for updating the organization, I will create an issue for this.

PR #2037, opened 2024-09-19
Make data migrations idempotent

mattwr18:


db/data/20240722092527_add_organization_id_to_requests.rb
```diff
@@ -3,7 +3,7 @@
 class AddOrganizationIdToRequests < ActiveRecord::Migration[6.1]
   def up
     ActiveRecord::Base.transaction do
-      organization = Organization.singleton
+      organization = Organization.first
```
soey, line 6: what happens later on instances when there are multiple organizations? I would even add sth like return if Organization.count != 1 to ensure that it is only run with exactly one organization

db/data/20240722092527_add_organization_id_to_requests.rb
```diff
@@ -16,7 +16,7 @@ def up
 
   def down
     ActiveRecord::Base.transaction do
-      organization = Organization.singleton
+      organization = Organization.first
       return unless organization
```
soey, line 20: see above

db/data/20240726061113_add_organization_id_to_activity_notifications.rb
```diff
@@ -3,7 +3,7 @@
 class AddOrganizationIdToActivityNotifications < ActiveRecord::Migration[6.1]
   def up
     ActiveRecord::Base.transaction do
-      organization = Organization.singleton
+      organization = Organization.first
       return unless organization
```
soey, line 7: see above

db/data/20240726061113_add_organization_id_to_activity_notifications.rb
```diff
@@ -16,7 +16,7 @@ def up
 
   def down
     ActiveRecord::Base.transaction do
-      organization = Organization.singleton
+      organization = Organization.first
       return unless organization
```
soey, line 20: see above

db/data/20240726063319_add_organization_id_to_taggings.rb
```diff
@@ -3,7 +3,7 @@
 class AddOrganizationIdToTaggings < ActiveRecord::Migration[6.1]
   def up
     ActiveRecord::Base.transaction do
-      organization = Organization.singleton
+      organization = Organization.first
       return unless organization
```
soey, line 7: see above

db/data/20240726063319_add_organization_id_to_taggings.rb
```diff
@@ -16,7 +16,7 @@ def up
 
   def down
     ActiveRecord::Base.transaction do
-      organization = Organization.singleton
+      organization = Organization.first
       return unless organization
```
soey, line 20: see above

soey (commented):
Ouch ..yes .. data migrations and code .. they are still painful. I would be even more pessimistic and add safeguards also for the cases if we have multiple organizations

mattwr18:
> Ouch ..yes .. data migrations and code .. they are still painful. I would be even more pessimistic and add safeguards also for the cases if we have multiple organizations

this is a good point and makes me want to remove this data migrate gem and go back to the temp rake tasks/ansible plays, which can be run once on whichever instances they need to be run and aren't run automatically on deploy.

mattwr18:
Superseded by https://github.com/tactilenews/100eyes/pull/2042

PR #2047, opened 2024-09-30
Fix image uploading/sending for 360dialog

### LINKED ISSUE #2046: [BUG] 360dialog file sending broken
The upload file job seems to succeed, but no job to send out the file is scheduled.

mattwr18:
Fixes #2046


app/jobs/whats_app_adapter/three_sixty_dialog/upload_file.rb
```diff
@@ -5,43 +5,41 @@
...
   module ThreeSixtyDialog
     class UploadFile < ApplicationJob
-      # rubocop:disable Metrics/AbcSize
       def perform(message_id:)
         @message_id = message_id
         message = Message.find_by(id: message_id)
-        return unless message
 
         request = message.request
         organization = Organization.find_by(id: request.organization.id)
```
roschaefer, line 13: So we let `NilPointerError` happen and let the job crash?

app/jobs/whats_app_adapter/three_sixty_dialog/upload_file.rb
```diff
@@ -5,43 +5,41 @@
...
-            'Content-Type': file.blob.content_type
+            'D360-API-KEY' => organization.three_sixty_dialog_client_api_key
           }
-          request = Net::HTTP::Post.new(url.to_s, headers)
-          request.body = File.read(ActiveStorage::Blob.service.path_for(file.blob.key))
+          params = {
+            'messaging_product' => 'whatsapp',
+            'file' => UploadIO.new(ActiveStorage::Blob.service.path_for(file.blob.key), file.blob.content_type)
+          }
+          request = Net::HTTP::Post::Multipart.new(url, params, headers)
```
roschaefer, line 25: It looks very much different from what was there before. What is the bug in #2046 and how does the PR actually fix it? It never succeeded? Errors being swallowed? Wrong status code? What happened?

roschaefer (approved):
yes but I still don't understand it

OK, I'll hit approve here. I don't understand how the fix works though.

How about VCR and recording a file upload during a test? :thinking:

app/jobs/whats_app_adapter/three_sixty_dialog/upload_file.rb
mattwr18: Yes, I added these guard clauses to turn the jobs into no-ops in the case that the message doesn't exist, for example, it was deleted before the job actually ran. However, after discussing this with @soey we agreed that for our use case, it would be more helpful if it did fail loudly vs quietly so that we know there is a potential issue.

We could change line 11 to a simple find, which will through an error if there is no message with the id message_id, but I'm not sure it's _that_ important the specific error.

app/jobs/whats_app_adapter/three_sixty_dialog/upload_file.rb
mattwr18: Sorry, that is a good point, I should have written more in the description.
The reason it failed silently was that the guard clause that looked to see if there was an organization and that that organization had a 360dialog client api key was incorrect. It said to return unless the attr was blank instead of return unless it's present.

The next issue was that when I removed the Setting in favor of looking up the env variables, I copy pasted the incorrect base uri for many requests. So, I needed to update to the correct base_uri.

Then, the api changed when we migrated to the Cloud API. Now, all requests need to have the `messaging_product` param set to `whatsapp` and the upload media post request needs to be posted with form data. See https://docs.360dialog.com/partner/messaging/media-messages/upload-retrieve-delete-media#upload-media

Finally, the return structure is different so that would have been another point of failure.

PR #2049, opened 2024-09-30
Fix welcome message template creation/usage

### LINKED ISSUE #2033: [improvement] 360dialog Add CreateTemplateJob
At the moment, we create the default welcome message and all the request templates programattically in the `WhatsAppAdapter::CreateTemplates` job after the api key is created for the client on onboarding. However, the welcome message is intended to be customized, which means that we have the need to create single templates on demand.

mattwr18:

This adds a job to create a new welcome message template, which is the only use case at the moment. Adding new request templates is not covered because there is no way to change them from the client side.

This also takes into account that multiple organizations can use an instance and they need to have the ability to create their own welcome messages with unique template names/text.

Fixes #2033


app/jobs/whats_app_adapter/three_sixty_dialog/create_welcome_message_template_job.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+require 'net/http'
+
+module WhatsAppAdapter
+  module ThreeSixtyDialog
+    class CreateWelcomeMessageTemplateJob < ApplicationJob
```
roschaefer, line 7: How do we trigger this and where does `template_text` come from?

Do we trigger everything manually from `bin/rails c`?

spec/support/shared_examples/contributor_resubscribes.rb
```diff
@@ -15,7 +15,7 @@
       expect(params[:organization_id]).to eq(organization.id)
       if adapter.eql?(WhatsAppAdapter::ThreeSixtyDialogOutbound::Text)
         expect(params[:payload][:to]).to eq(contributor.whats_app_phone_number.split('+').last)
-        expect(params[:payload][:template][:name]).to eq('welcome_message')
+        expect(params[:payload][:template][:name]).to eq("welcome_message_#{organization.project_name.parameterize.underscore}")
```
roschaefer, line 18:
```suggestion
        expect(params[:payload][:template][:name]).to eq("welcome_message_test_organization")
```

This assertion value should be hard-coded.

roschaefer (changes requested):
welcome

I have some mixed feelings about code that needs to get manually triggered. How about a UI for this?

I did the following on branch `main`:
```
git grep CreateWelcomeMessageTemplateJob
```
no results

mattwr18:
Yeah, we want to create a welcome message for the client after they update the default welcome message with their customized one. So, in theory we could trigger the job from the after_update_commit on the organization, which checks if either of the values that the welcome message is composed of have changed.

The issue with this is that the names must be unique. So, the first time they update it, it should work, but if they update it again, we need to delete the outdated one, wait some time, and then request a new one with the same name.

We could send a request to get all template names for the client, see if they already have a dynamic welcome message and delete it, in the case they do. Then, we could schedule a job to create a new one some time later.

app/jobs/whats_app_adapter/three_sixty_dialog/create_welcome_message_template_job.rb
mattwr18: Yes, it just saves me time from not needing to copy paste the net http commands. I was thinking of the template name and how it makes sense to just copy paste the way we  capture the text in the welcome message in the send_welcome_message!

roschaefer:
> Yeah, we want to create a welcome message for the client after they update the default welcome message with their customized one. So, in theory we could trigger the job from the after_update_commit on the organization, which checks if either of the values that the welcome message is composed of have changed.
> 
> The issue with this is that the names must be unique. So, the first time they update it, it should work, but if they update it again, we need to delete the outdated one, wait some time, and then request a new one with the same name.
> 
> We could send a request to get all template names for the client, see if they already have a dynamic welcome message and delete it, in the case they do. Then, we could schedule a job to create a new one some time later.

Urgh. That's quite an error-prone procedure and it doesn't have great good UX either. Shall we inform the user when they update the welcome message that we need to be informed and do things manually at least?

mattwr18:
>Urgh. That's quite an error-prone procedure and it doesn't have great good UX either. Shall we inform the user when they update the welcome message that we need to be informed and do things manually at least?

We already send out an email to all admin of the instance when someone updates either attribute that is used to make up the welcome message. See [notify_admin_of_welcome_message_change](https://github.com/tactilenews/100eyes/blob/main/app/models/organization.rb#L123-L130). This is my call to action to do something. We could do as I described :point_up: , but I after looking at the documentation for [templates](https://docs.360dialog.com/partner/messaging/template-messages), I noticed that we can edit templates in the 360dialog hub. I've just tested this out and it worked, so there is no need to delete and re-create the template.

spec/support/shared_examples/contributor_resubscribes.rb
mattwr18: I'm not sure about that @roschaefer ... Consider a test where the person setting it up wants to change the project name of the organization. Then it would not be `test_organization`, but something different.

roschaefer: Well, then we obviously set the name of the organization per test-case, no?

mattwr18: I think this is a stylistic preference that some devs on the team prefer and some don't, which indicates to me that we should allow space for differing styles, no?

app/jobs/whats_app_adapter/three_sixty_dialog/create_welcome_message_template_job.rb
```diff
@@ -5,14 +5,41 @@
 module WhatsAppAdapter
   module ThreeSixtyDialog
     class CreateWelcomeMessageTemplateJob < ApplicationJob
-      def perform(organization_id:, template_text:, token:)
+      def perform(organization_id:)
         @organization = Organization.find_by(id: organization_id)
-        @token = token
-        @template_text = template_text
+        @token = WhatsAppAdapter::ThreeSixtyDialog::TokenFetcherService.new.call
```
roschaefer, line 10: This looks like an obvious class method, doesn't it?

app/services/whats_app_adapter/three_sixty_dialog/token_fetcher_service.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module WhatsAppAdapter
+  module ThreeSixtyDialog
+    class TokenFetcherService
+      def initialize; end
+
+      def call
```
roschaefer, line 8:
```rb
      def self.call
```

For example. Feel free to use a name other than `call.`

app/jobs/whats_app_adapter/three_sixty_dialog/create_welcome_message_template_job.rb
```diff
@@ -5,14 +5,41 @@
...
+        @waba_account_id = organization.three_sixty_dialog_client_waba_account_id
+
+        if "welcome_message_#{organization.project_name.parameterize.underscore}".in?(existing_templates)
+          User.admin.find_each do |admin|
+            PostmarkAdapter::Outbound.welcome_message_updated!(admin, organization)
+          end
+        else
+          create_welcome_message_template
+        end
+      end
```
roschaefer, line 26: There is a lot going on here. In a perfect world I would write sth. like this:
```ruby
def update_welcome_message_templates(organization)
  User.admin.find_each { |admin| PostmarkAdapter::Outbound.welcome_message_updated!(admin, organization) }
end

# ...

        template_name = "welcome_message_#{organization.project_name.parameterize.underscore}"
        if template_name.in?(existing_templates)
          update_welcome_message_templates(organization)
        else
          create_welcome_message_template
        end
      end
```

spec/models/organization_spec.rb
```diff
@@ -3,23 +3,31 @@
...
 
+    let!(:admin) { create_list(:user, 3, admin: true) }
+    let!(:users) { create_list(:user, 2) }
+
     context 'does not schedule a PostmarkAdpater::Outbound job' do
       let(:params) { { name: 'SomethingElse', upgrade_discount: 100 } }
 
       it 'for other updates' do
-        expect { subject }.not_to have_enqueued_job
+        expect { subject }.not_to have_enqueued_job.on_queue('default').with(
```
roschaefer, line 18: This looks dangerous. Can we please stay specific for what we expected *to be* there and unspecific for things that we expect **not** to be there?

spec/vcr_setup.rb
```diff
@@ -11,6 +11,19 @@
...
+    ENV['THREE_SIXTY_DIALOG_PARTNER_PASSWORD']
+  end
+  c.filter_sensitive_data('THREE_SIXTY_DIALOG_PARTNER_USERNAME') do
+    ENV['THREE_SIXTY_DIALOG_PARTNER_USERNAME']
+  end
+  c.filter_sensitive_data('THREE_SIXTY_DIALOG_PARTNER_ID') do
+    ENV['THREE_SIXTY_DIALOG_PARTNER_ID']
+  end
+  c.filter_sensitive_data('Bearer <TOKEN>') do |interaction|
+    interaction.request.headers['Authorization'].try(:first)
```
roschaefer, line 24: Nice. This works?

spec/models/organization_spec.rb
```diff
@@ -54,6 +62,44 @@
...
+      allow(ENV).to receive(:fetch).with(
+        'THREE_SIXTY_DIALOG_PARTNER_REST_API_ENDPOINT', 'https://stoplight.io/mocks/360dialog/360dialog-partner-api/24588693'
+      ).and_return('https://stoplight.io/mocks/360dialog/360dialog-partner-api/24588693')
+    end
+
+    context 'does not schedule a PostmarkAdpater::Outbound job' do
+      let(:params) { { name: 'SomethingElse' } }
+
+      it 'for other updates' do
+        expect { subject }.not_to have_enqueued_job(WhatsAppAdapter::ThreeSixtyDialog::CreateWelcomeMessageTemplateJob)
```
roschaefer, line 78: Won't enqueue *any* job, no?

spec/support/shared_examples/contributor_resubscribes.rb
roschaefer: :laughing: You do you, I don't have a strong opinion.

I just had some bad experiences with assertion values that would change in an unexpected way. A classic is `expect(page).to have_text(expected_text)` and `expected_text` was `nil` which would always make the assertion true (-> false negative test).

In this case the risk is low. It's just a good routine.

It also reads a little better: `welcome_message_super_organization`

But do as you please, I won't argue over this :stuck_out_tongue:

roschaefer (approved):
worried

In general it's quite hard for me to review, but I'll give this an "Approve" because I cannot see any serious issue. Also the VCR cassette makes me a little less nervous.

app/jobs/whats_app_adapter/three_sixty_dialog/create_welcome_message_template_job.rb
mattwr18: Yes, I guess the reason I didn't do it that way is to allow for passing in some params, but I'm not sure we ever will and if we do, we can always update the code.

app/services/whats_app_adapter/three_sixty_dialog/token_fetcher_service.rb
mattwr18: The reason I had explained that this is a common pattern to name it call is because the service object should only handle one responsibility and the class should describe what it does, not the method.

app/jobs/whats_app_adapter/three_sixty_dialog/create_welcome_message_template_job.rb
mattwr18: sure, I like that the method explains what is happening.

spec/models/organization_spec.rb
mattwr18: I agree

mattwr18:
> In general it's quite hard for me to review, but I'll give this an "Approve" because I cannot see any serious issue. Also the VCR cassette makes me a little less nervous.

That's why I added it.
To be honest, there are still some jobs that don't have tests, but they weren't added in this PR so I've left it as is for now.

spec/support/shared_examples/contributor_resubscribes.rb
mattwr18: See 0caffe28
I generally agree with your reasons and I had a closer look, there is only one organization created that uses this shared example, so it was simple to make the change.

spec/vcr_setup.rb
mattwr18: Yes, I found this here https://stackoverflow.com/questions/71163339/replace-sensitive-data-in-vcr-gem-cassettes-request-body and other places. If you look at the cassette `vcr_cassettes/three_sixty_dialog_welcome_message_created.yml`, you can see that it filters out the token from the subsequent requests to fetch existing templates and create the new one.

PR #2052, opened 2024-10-02
Adapt Twilio template message sending to content api

### LINKED ISSUE #2051: Adapt Twilio WhatsApp to Content API

mattwr18:
Closes #2051

### DIFF
--- a/app/adapters/whats_app_adapter/twilio_outbound.rb
+++ b/app/adapters/whats_app_adapter/twilio_outbound.rb
@@ -55,8 +55,7 @@ module WhatsAppAdapter
       def send_message_template!(recipient, message)
         recipient.update(whats_app_message_template_sent_at: Time.current)
         content_sid = message.organization.twilio_content_sids["new_request_#{time_of_day}#{rand(1..3)}"]
-        WhatsAppAdapter::TwilioOutbound::Template.perform_later(organization_id: message.organization.id, contributor_id: recipient.id,
-                                                                content_sid: content_sid, message: message)
+        WhatsAppAdapter::TwilioOutbound::Template.perform_later(content_sid: content_sid, message_id: message.id)
       end
 
       private
--- a/app/adapters/whats_app_adapter/twilio_outbound.rb
+++ b/app/adapters/whats_app_adapter/twilio_outbound.rb
@@ -54,10 +54,9 @@ module WhatsAppAdapter
 
       def send_message_template!(recipient, message)
         recipient.update(whats_app_message_template_sent_at: Time.current)
-        text = I18n.t("adapter.whats_app.request_template.new_request_#{time_of_day}_#{rand(1..3)}", first_name: recipient.first_name,
-                                                                                                     request_title: message.request.title)
-        WhatsAppAdapter::TwilioOutbound::Text.perform_later(organization_id: message.organization.id, contributor_id: recipient.id,
-                                                            text: text)
+        content_sid = message.organization.twilio_content_sids["new_request_#{time_of_day}#{rand(1..3)}"]
+        WhatsAppAdapter::TwilioOutbound::Template.perform_later(organization_id: message.organization.id, contributor_id: recipient.id,
+                                                                content_sid: content_sid, message: message)
       end
 
       private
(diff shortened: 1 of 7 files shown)


app/adapters/whats_app_adapter/twilio_outbound/template.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module WhatsAppAdapter
+  class TwilioOutbound
+    class Template < ApplicationJob
+      queue_as :default
+
+      def perform(organization_id:, contributor_id:, content_sid:, message:)
```
roschaefer, line 8: @mattwr18 You sure that serializing objects to Redis won't cause us any problems here? I refer to `message:`

spec/adapters/whats_app_adapter/twilio_outbound_spec.rb
```diff
@@ -10,14 +10,14 @@
     create(:organization, onboarding_success_heading: onboarding_success_heading, onboarding_success_text: onboarding_success_text)
   end
   let(:contributor) { create(:contributor, email: nil) }
+  let(:onboarding_success_heading) { 'Thanks for onboarding' }
+  let(:onboarding_success_text) { 'We will start sending messages soon.' }
 
   describe '::send_welcome_message!' do
     let(:expected_job_args) do
       { organization_id: organization.id, contributor_id: contributor.id,
         text: ["*#{onboarding_success_heading}*", onboarding_success_text].join("\n\n") }
```
roschaefer, line 19:
```suggestion
        text: "*Thanks for onboarding*\n\nWe will start sending messages soon." }
```

:grimacing:

app/adapters/whats_app_adapter/twilio_outbound.rb
```diff
@@ -54,10 +54,9 @@ def send_resubscribe_error_message!(contributor, organization)
 
       def send_message_template!(recipient, message)
         recipient.update(whats_app_message_template_sent_at: Time.current)
-        text = I18n.t("adapter.whats_app.request_template.new_request_#{time_of_day}_#{rand(1..3)}", first_name: recipient.first_name,
-                                                                                                     request_title: message.request.title)
-        WhatsAppAdapter::TwilioOutbound::Text.perform_later(organization_id: message.organization.id, contributor_id: recipient.id,
```
roschaefer, line 59: Is it possible that we forgot to delete the file defininig `WhatsAppAdapter::TwilioOutbound::Text`? :thinking:

roschaefer (approved):
Same here, I don't see anything dangerous. So I won't block this from being merged :+1: 

ok

app/adapters/whats_app_adapter/twilio_outbound.rb
mattwr18: We didn't delete this class because it is still used to send freeform messages to contributors. If they respond to a template for example then it opens a 24-hour window where we can send them whatever message we want. This is how they receive the actual personalized message that is created for them when a request is created by a user of the app.

One thing is that the welcome message is a template and we do not account for this in this PR because the idea is to prevent loss of ongoing service, but not to support new contributors onboarding.

spec/adapters/whats_app_adapter/twilio_outbound_spec.rb
mattwr18: See c0c1f0f3

app/adapters/whats_app_adapter/twilio_outbound/template.rb
mattwr18: Yes, I am sure because the pattern in the project was to pass records into all jobs before I started to refactor away from that pattern. The reason I understand it is a good practice to pass in the id and look up the record is there is a chance that the record changed or was deleted after the job was scheduled, but before the job was executed. 

I think the reason it works is documented here https://guides.rubyonrails.org/v6.1/active_job_basics.html#globalid

That being said, I refactored this job and just pass in the  content_sid and the message_id, look up the message and derive the organization and contributor from it. I think this is an improvement and also addresses a concern that @soey has brought up with paramater drilling.

See 537f3cc4

PR #2053, opened 2024-10-05
[improvement] 360dialog extend request to receive message feature

### LINKED ISSUE #2031: [improvement] 360dialog extend request to receive message feature
This involves:

- [x] Add message_id in response to message record's external_id
- [x] Looking at incoming message for:

```
{ "context": {"message_id":"MESSAGE_ID"}}
```

and finding the message with this external id to deliver to the contributor.

mattwr18:

We now save the external id of outgoing 360dialog WhatsApp messages to the message record in the database. This allows us to:

a) Send out a previous personalized message that the contributor might have forgotten to ask for, but is still interested in responding to, and
b) Support a reply_to feature where the contributor might quote reply an older message and we can properly assign the incoming message to the correct request instead of trying to infer it solely based on the latest message they received and relying on the user to move the message to the correct request if it has been incorrectly assigned.

Closes #2031


roschaefer (approved):
Same here, I don't see anything serious, but I took me a while to get my head around this.

hm ok

mattwr18:
> Same here, I don't see anything serious, but I took me a while to get my head around this.

I've updated the description. I hope that helps.

PR #2061, opened 2024-10-10
Fix create templates job failing on first run

### LINKED ISSUE #2044: [BUG] WhatsAppAdapter::CreateTemplates fails when scheduled by CreateApiKey job
It fails with this message

```
Error occurred for WhatsApp with error code: 404 with message: {"meta": {"success": false, "http_code": 404, "developer_message": "Can't get WABA account. WABA account with id true was not found. Please, try another id.", "360dialog_trace_id": "<trace_id>"}}
```

It seems that the `organization.three_sixty_dialog_client_waba_account_id` is nil... re-running the job from the rails console works as this value is reloaded.

mattwr18:
Fixes #2044

The first time the job is run the
three_sixty_dialog_client_waba_account_id is nil so it calls the fetch_client_info method. This method successfully updates the organization, but as it's the last line of the method it saves the return of the update `true` to the local variable that is used in the url to create the templates. This results in an error. Subsequent times it's ran it does not error because it uses the value of the organization's attribute, which is the correct value.


app/jobs/whats_app_adapter/create_templates.rb
```diff
@@ -79,6 +79,7 @@ def fetch_client_info
       waba_account = client_hash['waba_account']
       organization.update!(three_sixty_dialog_whats_app_template_namespace: waba_account['namespace'],
                            three_sixty_dialog_client_waba_account_id: waba_account['id'])
+      waba_account
```
roschaefer, line 82: I've put a `binding.pry` here and it seems this code is not covered by any spec.

I would :heart: if we could test this method.

spec/services/whats_app_adapter/three_sixty_dialog/file_fetcher_service_spec.rb
```diff
@@ -7,7 +7,7 @@
   describe '#call' do
     subject { -> { described_class.new(organization_id: organization.id, file_id: file_id).call } }
 
-    let(:organization) { create(:organization) }
+    let!(:organization) { create(:organization) }
```
roschaefer, line 10: I've run:
```
git checkout origin/main -- spec/services/whats_app_adapter/three_sixty_dialog/file_fetcher_service_spec.rb
```

But this spec doesn't fail.

Instead, here and there, intermittent failures of other, unrelated specs.

@mattwr18 can you shine some light on this?

roschaefer (approved):
At last, a PR that contains a super easy to understand bugfix. :sweat_smile: 

:100: % approve.

absolutely

Only confusion is the changed line in the spec and I'm a little sad we don't have test coverage for the changed productive code.

app/jobs/whats_app_adapter/three_sixty_dialog/create_templates_job.rb
```diff
@@ -0,0 +1,75 @@
...
+module WhatsAppAdapter
+  module ThreeSixtyDialog
+    class CreateTemplatesJob < ApplicationJob
+      def perform(organization_id:)
+        @organization = Organization.find(organization_id)
+
+        @base_uri = ENV.fetch('THREE_SIXTY_DIALOG_WHATS_APP_REST_API_ENDPOINT', 'https://stoplight.io/mocks/360dialog/360dialog-partner-api/24588693')
+        existing_templates = WhatsAppAdapter::ThreeSixtyDialog::TemplateFetcherService.new(organization_id: organization.id).call
+
+        templates_to_create_array = whats_app_templates.keys.difference(existing_templates)
```
roschaefer, line 14: nice :+1:

app/jobs/whats_app_adapter/three_sixty_dialog/create_templates_job.rb
```diff
@@ -0,0 +1,75 @@
...
+        templates_to_create = whats_app_templates.select { |key, _value| key.in?(templates_to_create_array) }
+        templates_to_create.each do |key, value|
+          @template_name = key
+          @template_text = value
```
roschaefer, line 18: This looks weird, if we're in a loop, these two instance variables get overwritten on each iteration. Only the last iteration will matter.

app/jobs/whats_app_adapter/three_sixty_dialog/create_templates_job.rb
```diff
@@ -0,0 +1,75 @@
...
+          category: 'MARKETING',
+          components: [
+            {
+              type: 'BODY',
+              text: template_text,
+              example: {
+                body_text: [
+                  [
+                    'Jakob',
+                    'Familie und Freizeit'
```
roschaefer, line 49: This is production code? :thinking:

app/jobs/whats_app_adapter/three_sixty_dialog/create_welcome_message_template_job.rb
```diff
@@ -7,80 +7,40 @@ module ThreeSixtyDialog
...
 
-        @base_uri = ENV.fetch('THREE_SIXTY_DIALOG_PARTNER_REST_API_ENDPOINT', 'https://stoplight.io/mocks/360dialog/360dialog-partner-api/24588693')
-        @partner_id = ENV.fetch('THREE_SIXTY_DIALOG_PARTNER_ID', nil)
-        @waba_account_id = organization.three_sixty_dialog_client_waba_account_id
-        existing_templates = WhatsAppAdapter::ThreeSixtyDialog::TemplateFetcherService.new(
-          waba_account_id: waba_account_id,
-          token: token
-        ).call
+        @base_uri = ENV.fetch('THREE_SIXTY_DIALOG_WHATS_APP_REST_API_ENDPOINT', 'https://stoplight.io/mocks/360dialog/360dialog-partner-api/24588693')
+        existing_templates = WhatsAppAdapter::ThreeSixtyDialog::TemplateFetcherService.new(organization_id: organization.id).call
```
roschaefer, line 12: `SomeClass.new.call` could be a class method

app/services/whats_app_adapter/three_sixty_dialog/template_creator_service.rb
```diff
@@ -0,0 +1,41 @@
...
+        @payload = payload
+      end
+
+      attr_reader :base_uri, :organization, :payload
+
+      def call
+        url = URI.parse("#{base_uri}/v1/configs/templates")
+        headers = { 'D360-API-KEY' => organization.three_sixty_dialog_client_api_key, 'Content-Type' => 'application/json' }
+
+        request = Net::HTTP::Post.new(url.to_s, headers)
```
roschaefer, line 18: Can't we omit the `.to_s`?

spec/jobs/whats_app_adapter/three_sixty_dialog/create_templates_job_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      allow(ENV).to receive(:fetch).with(
+        'THREE_SIXTY_DIALOG_WHATS_APP_REST_API_ENDPOINT', 'https://stoplight.io/mocks/360dialog/360dialog-partner-api/24588693'
+      ).and_return('https://waba-v2.360dialog.io')
+      allow_any_instance_of(WhatsAppAdapter::ThreeSixtyDialog::CreateTemplatesJob).to receive(:whats_app_templates).and_return(
+        { 'new_request_morning2' => 'some text' }
+      )
+    end
+
+    describe 'ActivityNotifications', vcr: { cassette_name: :three_sixty_dialog_create_template } do
+      it 'creates a notification for all admin and users of an organizaiton' do
```
roschaefer, line 25:
```suggestion
      it 'creates a notification for all admins and users of an organizaiton' do
```

spec/jobs/whats_app_adapter/three_sixty_dialog/create_templates_job_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        subject.call
+        whats_app_template_created_notifications = ActivityNotification.where(type: WhatsAppTemplateCreated.name)
+        expect(whats_app_template_created_notifications.count).to eq(5)
+
+        recipient_ids = whats_app_template_created_notifications.pluck(:recipient_id).uniq.sort
+        user_ids = users_of_an_organization.pluck(:id).uniq
```
roschaefer, line 31:
```suggestion
        user_ids = users_of_an_organization.pluck(:id)
```
Why `uniq`?

spec/services/whats_app_adapter/three_sixty_dialog/file_fetcher_service_spec.rb
roschaefer: So it didn't fail on my machine because I had a `ATTR_ENCRYPTED_KEY` set?

app/services/whats_app_adapter/three_sixty_dialog/template_creator_service.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module WhatsAppAdapter
+  module ThreeSixtyDialog
+    class TemplateCreatorService
```
roschaefer, line 5: I see you calling this always with `TemplateCreatorService.new(/*...*/).call` you can make this a simple (class) method then.

roschaefer (approved):
I don't see anything serious here, so I'll approve. Here are some improvements though.

improve

app/jobs/whats_app_adapter/three_sixty_dialog/create_templates_job.rb
mattwr18: Yes, we just need placeholder values when we create the template so that WhatsApp knows the types of values we will be sending dynamically.

spec/jobs/whats_app_adapter/three_sixty_dialog/create_templates_job_spec.rb
mattwr18: I think all admin is actually better English than admins.

mattwr18:
Thanks for 2 reviews @roschaefer !!

PR #2062, opened 2024-10-10
Improve template creation

mattwr18:


mattwr18:
Changes incorporated in https://github.com/tactilenews/100eyes/pull/2061

PR #2070, opened 2024-11-07
[fix] Scope tag counts to active contributors

### LINKED ISSUE #2069: Tag count incorrect for @available_tags
### Steps to reproduce

* Tag two contributors with the same tag
* Mark one of them inactive, or unsubscribe from one of them
* Go to `/1/requests/new`
* Click on the TagsInput, which will open a dropdown with the available tags and the count, which incorrectly counts non-active contributors.

This can lead to confusion as once the person creates the request, they will see a different count of how many contributors the request was sent out to.

mattwr18:
Fixes #2069

### DIFF
--- a/app/controllers/organizations/contributors_controller.rb
+++ b/app/controllers/organizations/contributors_controller.rb
@@ -49,7 +49,7 @@ module Organizations
     end
 
     def count
-      render json: { count: @organization.contributors.with_tags(params[:tag_list]).count }
+      render json: { count: @organization.contributors.active.with_tags(params[:tag_list]).count }
     end
 
     def conversations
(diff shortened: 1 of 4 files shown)


roschaefer (approved):
Perfect! :+1: 

perfect

Maybe you could check if any extra DB calls are fired during evaluation, but I think it's fine for now.

mattwr18:
> Maybe you could check if any extra DB calls are fired during evaluation, but I think it's fine for now.

I found this fix in the issue list over at acts-as-taggable-on and it says there that it only generates one query. See https://github.com/mbleigh/acts-as-taggable-on/issues/719#issuecomment-172490087; however, ours is generated 5

I don't know if there is a way to optimize the query.

PR #2072, opened 2024-11-07
[improvement] WhatsApp file uploading/sending and error handling

mattwr18:
TODO:

- [ ] Add specs for statuses/errors in spec/requests/whats_app/three_sixty_dialog_webhook_spec.rb
- [x] refactor app/adapters/whats_app_adapter/three_sixty_dialog_outbound/file.rb to remove disabling rubocop

Specs for statuses/errors will be added in the PR that actually handles statuses.


app/adapters/whats_app_adapter/three_sixty_dialog_outbound.rb
```diff
@@ -114,17 +114,15 @@ def send_message_template(recipient, message)
...
-        end
+        WhatsAppAdapter::ThreeSixtyDialog::UploadFileJob.perform_later(message_id: message.id) if files.present?
+
+        return if message.text.blank?
+
+        WhatsAppAdapter::ThreeSixtyDialogOutbound::Text.perform_later(organization_id: message.organization.id,
+                                                                      payload: text_payload(
+                                                                        recipient, message.text
+                                                                      ),
+                                                                      message_id: message.id)
```
mattwr18, line 125: I think it might be a better UX to upload the file, if one is attached, then send out the file and after it was sent successfully. Alternatively, I've since clarified that we can send the message text as a caption with a limit of 1024 characters. So, we could potentially do that. For telegram, we check if the message text is greater than the limit and if it is we send it separately.

app/adapters/whats_app_adapter/three_sixty_dialog_outbound.rb
```diff
@@ -114,17 +114,15 @@ def send_message_template(recipient, message)
...
         files = message.files
 
-        if files.blank?
-          WhatsAppAdapter::ThreeSixtyDialogOutbound::Text.perform_later(organization_id: message.organization.id,
-                                                                        payload: text_payload(
-                                                                          recipient, message.text
-                                                                        ),
-                                                                        message_id: message.id)
-        else
-          files.each do |_file|
```
roschaefer, line 124: We used to enqueue `UploadFileJob`s for every `_file`. Does it matter that we enqueue only one job now?

roschaefer (approved):
This PR is looking good! :+1: 

We can do the caption idea that you mention, but we can also keep things simple stupid and just send files immediately followed by a text message, no? I do this often myself.

app/adapters/whats_app_adapter/three_sixty_dialog_outbound.rb
mattwr18: This actually was an oversight. We iterated through every file attached to a message to schedule a job to upload the file, but then iterated through a request's files to actually upload them so we were uploading many more files than were necessary.

PR #2089, opened 2025-01-14
[fix] quick reply button not responding with the request

### LINKED ISSUE #2081: [BUG] Quick reply button only works once
### Steps to reproduce

* Send your WhatsApp contributor a message template
* Send another one
* Click on the quick reply button to ask for the request
*  Click on the other template's quick reply button to receive the request
* Observe that you do not receive the other request.

### What is expected

Anytime someone clicks on the quick reply button to receive the request, they should receive it.

### LINKED ISSUE #2093: [BUG] Sending more info quick reply can result in duplicate messages
### Steps to reproduce

* Send out a request to your WhatsApp contributor
* Click on the More infos button
* Send out a new request
* You should receive the freeform message
* Respond to this message with whatever
* You receive the request text a second time

### What is happening

The way that it works now is that we mark the contributor record any time we send out a whats app message template. Then any message they send us, unless it's an unsubscribe, re-subscribe, or request for more info keyword, we respond with the latest request. The idea behind this was not to limit them to only sending us the Antworten keyword and or clicking on the Antworten button.

However, when they click on the Mehr Infos button, we mark that we have received a message from them, since we, in fact, have. This means that if we send out another request within the 24-hour window, we do not need to send a message template, we can send a "freeform" message. We don't reset the timestamp of when we sent the template because if they still want the actual message, we want to give it to them.

If someone asks for more info and then a new request were sent out which was answered, we see that we had very recently sent a message template and wrongly assume they are asking for the latest message, which in this edge case is the same message they are answering.

The easiest fix would be to only respond with the request if they either quote reply the message and/or answer with exactly the configured answer request keyword (default Antworten).

### LINKED ISSUE #2094: [BUG] 360 dialog request to receive message does not work properly
###  Steps to reproduce

* Send yourself two template messages
* Click on answer for the first of such messages
* See that you receive the latest message

### What's happening

When we send a template message we don't pass in a message id so we don't save the external_id that comes back, which is later received in the webhook's payload. Therefore, we can never find the previous message. The automated tests just checked that if the message did have the id, it would be found and sent, but didn't test that it was ever saved.

mattwr18:
Fixes #2081
Fixes #2093 
Fixes #2094

## What changed and Why

We now always trigger the request to receive request callback when we receive the keyword. This fixes a bug where we would only send the first request the contributor asked for. This happened cause we mark a timestamp when we send out a template message, when a contributor responds, we set it back to nil and mark a timestamp that they responded.

This also involves several big refactorings. These include:

### Refactoring `WhatsAppAdapter::ThreeSixtyDialogInbound`
* It now handles its own "callbacks" as it seems more the responsibility of the adapter to handle these errors/jobs to send out messages based on some trigger than the ProcessWebhookJob whose job it should be to simply accept the webhook and pass it on.
* `initialize_message` method to remove the multiple actions it was triggering. This had several downsides: there were several checks to see if the inbound message was ephemeral data and what type before continuing. Now, we check the text/payload to see if the inbound message is ephemeral, ie should not be persisted to the db before even initializing the message instance.
* We now assign the request to the inbound message in the adapter itself, which also handles persisting it.
* With this, we added the feature to support quote replying a previous message. If we receive a context object with a message's external id, which corresponds to a message we have saved, we know which request they are responding to and don't need to assume it's the latest message they received.

### Fix for the request to receive previous message not working

* We now have a new model `Message::WhatsAppTemplate`, which belongs to a message record, whose job it is to save the external id of the template message that was sent out. This allows us to find the message by its whats_app_template that a contributor is requesting. 

Before this PR, we never saved this external id when a message template was sent out, so the feature never actually worked.

Originally, when I first started working on this fix, I just saved the external_id to the message record itself, but this didn't sit right since it was only a temporary external_id, which would get updated when/if the contributor actually asked for the message. It has the downside of not being able to properly keep track of delivery receipts. Consider an example where we mark the message record as sent, received, and read, but the contributor never actually asked for the message, we just incorrectly marked it because we had the incorrect external id assigned to it.

### Refactor of the `WhatsAppAdapter::ThreeSixtyDialogOutbound::Text` job

* We move all the payloads to the place they are used. We also pass in a type so that we can create the template message record after we know that it was successfully sent out.

### Specs
* Missing specs were also added.


app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
```diff
@@ -22,19 +22,24 @@ def on(callback, &block)
...
-      @sender = initialize_sender(whats_app_message)
+      @sender = initialize_sender
       return unless @sender
 
-      @message = initialize_message(whats_app_message)
-      return unless @message
+      @text = initialize_text
 
-      @unsupported_content = initialize_unsupported_content(whats_app_message)
+      if ephemeral_data?
```
roschaefer, line 32: In general I prefer the more specific wording. Ie. instead of "ephemeral" you can just call it "template message"

app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
```diff
@@ -22,19 +22,24 @@ def on(callback, &block)
...
+        trigger_appropriate_callback
```
roschaefer, line 33: Same here, what's "appropriate"?

app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
```diff
@@ -47,7 +52,35 @@ def trigger(event, *args)
...
+          UNSUBSCRIBE_CONTRIBUTOR
+        elsif resubscribe_text?
+          RESUBSCRIBE_CONTRIBUTOR
+        elsif request_to_receive_message?
+          REQUEST_TO_RECEIVE_MESSAGE
+        end
+      params_array = [sender]
+      params_array.append(whats_app_message[:messages].first) if request_to_receive_message?
+
+      trigger(callback, *params_array)
```
roschaefer, line 80:
```suggestion
      if request_for_more_info?
        trigger(REQUEST_FOR_MORE_INFO, sender)
      elsif unsubscribe_text?
        trigger(UNSUBSCRIBE_CONTRIBUTOR, sender)
      elsif resubscribe_text?
        trigger(RESUBSCRIBE_CONTRIBUTOR, sender)
      elsif request_to_receive_message?
        trigger(REQUEST_TO_RECEIVE_MESSAGE, sender, whats_app_message[:message])
      end
```
I'm leaning towards "request changes" on this refactoring. The original implementation should work as well, though, just more code.

spec/adapters/whats_app_adapter/three_sixty_dialog_inbound_spec.rb
```diff
@@ -530,6 +530,17 @@
           expect(subject).to have_received(:call)
         end
       end
+
+      describe 'sending answer request keyword' do
+        before do
+          answer_request_keyword = organization.whats_app_quick_reply_button_text['answer_request']
+          whats_app_message[:messages].first[:text][:body] = answer_request_keyword
+        end
```
roschaefer, line 538: We have different test styles :slightly_smiling_face: I would do it as such:

```suggestion
      describe 'if the contributor responds with the answer request keyword' do
        before do
          organization.whats_app_quick_reply_button_text['answer_request']= 'Antworten'
          organization.save

          whats_app_message[:messages].first[:text][:body] = 'Antworten'
        end
```

I prefer to hard-code these values because they give a better idea `answer_request` could be.

This is obviously nothing for which I would press "request changes".

app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
```diff
@@ -47,7 +52,35 @@ def trigger(event, *args)
...
+      message[:text]&.dig(:body) || message[:button]&.dig(:text) || supported_file(message)&.dig(:caption)
+    end
+
+    def ephemeral_data?
+      return if text.blank?
+
+      request_for_more_info? || unsubscribe_text? || resubscribe_text? || request_to_receive_message?
+    end
+
+    def trigger_appropriate_callback
```
roschaefer, line 66: I think `ephemeral_data?` and `trigger_appropriate_callback` can be merged. They would always be called toghether and you do the same checks again, right?

roschaefer (changes requested):
I'm quite happy we had a phone call over this PR. I don't see until now where the external id gets overridden in case the responded message was a template message :shrug: 

So this PR review is more for the code itself.

As said in the phone call, I think the potential for bugs doesn't justify a larger refactoring. (Potential for bugs because the external id of will change from template message :arrow_right: text message and we loose the connection to the original template message)

callback

app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
mattwr18: It's not just a template message though. They can use the quick reply buttons to send two responses that come as part of the template, but they can simply send the text. Also, the other two words for unsubscribing and resubscribing have nothing to do with the templates.

app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
mattwr18: I wrote it that way originally, but it involves more code duplication. All branches trigger the callback and all pass in the sender as the 2nd argument. The only difference is that the request to receive message passes in the first message as a third parameter.
I considered refactoring so they all have the same API and let the method handle extracting the data it's interested in.

spec/adapters/whats_app_adapter/three_sixty_dialog_inbound_spec.rb
mattwr18: I had that as well. There's no need to update the organization as its default is `Antworten`, but that does rely on the reader of the code having this knowledge.

app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
mattwr18: I did that at first, but refactored it because I'd like the methods to have a single responsibility.

app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
roschaefer: OK, so you prefer it as it is? I would suggest to put line 78 as close as possible to the call of `REQUEST_TO_RECEIVE_MESSAGE`. Otherwise it's a potential for bugs in my opinion.

I can hit "Approve" if you want.

mattwr18:
> I'm quite happy we had a phone call over this PR. I don't see until now where the external id gets overridden in case the responded message was a template message

Assuming we have not received a message from a contributor within the last 24 hours, we schedule the job to send out the message template here https://github.com/tactilenews/100eyes/blob/main/app/adapters/whats_app_adapter/three_sixty_dialog_outbound.rb#L105-L110 ... This did not pass in the message_id, so there was nothing to update, but it meant that we could not find the message record in the db based on the webhook we received with the contributor asking for a request that was not the latest.

In this PR, I added the message_id to the job here https://github.com/tactilenews/100eyes/pull/2089/files#diff-9c246b02543babe1e600a56966fa2326510711ea1dac16af061e9a71baa19557R111, then in the happy path of the job's request, we update the message record with the external id. See https://github.com/tactilenews/100eyes/blob/main/app/adapters/whats_app_adapter/three_sixty_dialog_outbound/text.rb#L40. However, this is the external id that corresponds to the template message as up to this point, we have not actually sent them the freeform message that the message record represents.

If they respond to this template in any way, we trigger a request to receive message action, which then triggers the same job, this time the payload contains the message record's text. If that is successfully sent, we get back a different external id and then re-update the same message record's external id.

roschaefer (changes requested):
I don't see anything serious besides what we discussed in our pairing :slightly_smiling_face: 

fingers crossed

app/jobs/whats_app_adapter/three_sixty_dialog/handle_ephemeral_data_job.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module WhatsAppAdapter
+  module ThreeSixtyDialog
+    class HandleEphemeralDataJob < ApplicationJob
```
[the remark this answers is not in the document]
mattwr18: Good call :+1:

app/models/contributor.rb
```diff
@@ -69,9 +69,11 @@ def self.with_lowercased_email(email)
   end
 
   def reply(message_decorator)
-    request = active_request or return nil
+    message = message_decorator.message
```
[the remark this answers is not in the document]
mattwr18: yes, I don't actually think it should be the contributor model that handles assigning a request to the reply and saving it to the db.

I've removed this from the WhatsAppAdapter::ThreeSixtyDialogInbound class and would like to refactor all adapters similarly

spec/jobs/whats_app_adapter/three_sixty_dialog/handle_ephemeral_data_job_spec.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe WhatsAppAdapter::ThreeSixtyDialog::HandleEphemeralDataJob do
```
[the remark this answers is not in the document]
mattwr18: yes, I ended up moving the code back into the adapter and handling it there since whether its a job or a service, if its job is to handle many types of ephemeral data, there would likely to be some conditionals in both the adapter and the service object.

app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
```diff
@@ -19,20 +19,18 @@ def consume(organization, whats_app_payload)
       @quote_reply_message_id = message_payload.dig(:context, :id)
       @text = initialize_text
 
-      if ephemeral_data?
-        handle_ephemeral_data
-        return
-      end
+      return if ephemeral_data_handled?
```
roschaefer, line 22: That's better than duplicating the long `if-else` statement, I'd say. :shrug:

app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
```diff
@@ -19,20 +19,18 @@ def consume(organization, whats_app_payload)
...
       @message = initialize_message
-      @message.request = initialize_request
 
       @unsupported_content = initialize_unsupported_content
 
       files = initialize_file
       @message.files = files
+      @message.request = sender.active_request
 
-      yield(@message) if block_given?
```
roschaefer, line 39: Not sure if the block was used somewhere..

app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
```diff
@@ -19,20 +19,18 @@ def consume(organization, whats_app_payload)
...
+      return if ephemeral_data_handled?
 
       @message = initialize_message
-      @message.request = initialize_request
 
       @unsupported_content = initialize_unsupported_content
 
       files = initialize_file
       @message.files = files
+      @message.request = sender.active_request
```
roschaefer, line 30: The `reply-to` functionality was removed? I would expect it here, otherwise.

app/adapters/whats_app_adapter/three_sixty_dialog_outbound/text.rb
```diff
@@ -61,18 +61,12 @@ def welcome_message_payload
...
-        case current_time
-        when morning..day
+        case Time.current.hour
+        when 6..11
           'morning'
-        when day..evening
+        when 11..17
           'day'
-        when evening..night
+        when 17..23
```
roschaefer, line 69: Is this tested? :grimacing:

app/jobs/whats_app_adapter/three_sixty_dialog/process_webhook_job.rb
```diff
@@ -11,7 +11,7 @@ def perform(organization_id:, components:)
         @whats_app_message = components
         adapter = WhatsAppAdapter::ThreeSixtyDialogInbound.new
 
-        adapter.consume(organization, components) { |message| message.contributor.reply(adapter) }
```
roschaefer, line 36: Does this gets replaced with `@message.request = sender.active_request` from above?

app/models/contributor.rb
```diff
@@ -69,11 +69,9 @@ def self.with_lowercased_email(email)
   end
 
   def reply(message_decorator)
-    message = message_decorator.message
-    request = message.request || active_request
-    return unless request
-
+    request = active_request or return nil
```
roschaefer, line 72:
```suggestion
    request = active_request 
    return nil unless request
```
Looks a bit more idiomatic.

`and` and `or` in ruby have low precedence. It won't matter here but with the suggested change it's more obvious that the assignment gets executed first and we return if `request` is nil, not if the return value of the assignment is nil.
image

app/models/contributor.rb
```diff
@@ -69,11 +69,9 @@ def self.with_lowercased_email(email)
   end
 
   def reply(message_decorator)
-    message = message_decorator.message
```
roschaefer: Won't this change have a big impact on other messengers besides Whatsapp? I think the method `reply` is called for other messengers but this PR doesn't touch them, right?

roschaefer (approved):
I'm giving this an approval even though I'm a bit worried that https://github.com/tactilenews/100eyes/pull/2089/files/efe9875a5b9c12179b73c97a9995c3aff5ce7a18..7e23fb19ff55d2ef7dead832c4df51c4569c8803#diff-5147cd4a4f7646dbcf1c6e9ba4df962eb805e7fab7b9dbe6171e28ef6ac35c11L72 has a side effect on other messengers that we didn't anticipate. :thinking:

app/models/contributor.rb
mattwr18: This change was an intermittent change that was reverted to favor setting the request in the adapter itself.
If you run a git diff between the HEAD of this PR and the main branch for the contributor model, you will see that there are no changes.

app/models/contributor.rb
mattwr18: I like this change, I also implemented it this way when I was adding the reply to logic to this method here https://github.com/tactilenews/100eyes/pull/2089/commits/5da7eb560a391990a25b51b1bd6b85b010862dec#diff-5147cd4a4f7646dbcf1c6e9ba4df962eb805e7fab7b9dbe6171e28ef6ac35c11R72-R74, but since I've reverted all changes to this file, I won't update it. My idea is to refactor so that the contributor class does not have the responsibility of assigning the correct request to a reply or saving the message and then this code will be removed.

app/jobs/whats_app_adapter/three_sixty_dialog/process_webhook_job.rb
mattwr18: yes, and the `@message.save!`

app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
mattwr18: Correct, see 8c7ea1c0

app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
mattwr18: the `{ |message| message.contributor.reply(adapter) }` block was passed in, but it's been removed.

app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb
mattwr18: For sure.

PR #2103, opened 2025-01-23
Allow messages to be saved without a request

### LINKED ISSUE #1649: [Feature] Allow messages without requests to be saved
```
As a user of 100eyes
I would like all messages received from clients to be saved
So that I don't lose potentially important information
```

### DEV Notes:

* Track how often we receive messages before sending out a request
 * Save them so that they are available in the admin view
 * Lay the groundwork for extending the move message feature to **no request**
 * Being able to view the messages somewhere (maybe a view with all messages to/from a contributor)

### LINKED ISSUE #1869: [BUG] Incoming messages before the first request has been broadcast are lost
### Steps to reproduce

1. Start with a clean instance
2. Onboard with a channel, of your choice.
3. After receiving the onboarding message, respond to it.
4. See that the message is completely ignored/lost.

### Dev Notes

This occurs because of the logic in how we save messages. At the moment, we require there to be an `active_request`. This is determined by either the latest request the contributor has received, or the latest request that has been broadcast. This causes multiple issues:

* If there are requests that have been broadcast, but none of them have targeted the contributor, by their tag list, and that contributor responds, it can cause confusion since the message can be assigned to the request that they did not receive.
*  It also means that if there are no requests, the messages are simply not saved at all.

There is a proposal to remove the hard requirement that a message belong to a request, see https://github.com/tactilenews/100eyes/issues/1649. While this would solve both issues :point_up: , it requires bigger changes.

It has been proposed that we get past the issue where the message is not saved at all by always ensuring that a new instance has a request created with the content of the welcome message. This way, if a contributor replies before the campaign starts it will be assigned to this request.

mattwr18:
Closes #1649
Fixes #1869 

After deploying, one must:

- [ ] Run the rake task `rails messages:add_organization_id_to_messages` to add the organization id to the message since it can no longer be derived through the organization
- [ ] Run the `WhatsAppAdapter::ThreeSixtyDialog::CreateTemplatesJob` to create the `new_direct_message` template, which is used when a direct message goes out that does not belong to a specific request.
- [ ] Update the 360dialog template text and buttons to use the Öffnen button
- [ ] Update the WDR/ZVW answer_request to Öffnen


app/components/requests_feed/requests_feed.rb
```diff
@@ -14,7 +14,7 @@ def initialize(organization:, contributor:, **)
     attr_reader :organization, :contributor
 
     def replies
-      @replies ||= contributor.recent_replies
+      @replies ||= contributor.most_recent_replies_to_some_request
```
roschaefer, line 17:
```suggestion
      @replies ||= contributor.recent_replies_to_a_request
```

app/models/contributor.rb
```diff
@@ -72,10 +72,9 @@ def self.with_lowercased_email(email)
   end
 
   def reply(message_decorator)
-    request = active_request or return nil
     ActiveRecord::Base.transaction do
       message = message_decorator.message
-      message.request = request
+      message.request = active_request
```
roschaefer, line 77: It can be `nil` and it's fine now?

app/models/contributor.rb
```diff
@@ -109,9 +108,7 @@ def channels
   end
 
   def active_request
```
roschaefer, line 110: Maybe we can do away with this method altogether? `received_messages.first&.request` sounds a lot better than `active_request`.

app/models/organization.rb
```diff
@@ -14,7 +14,7 @@ class Organization < ApplicationRecord
   has_many :contributors, dependent: :destroy
   has_many :requests, dependent: :destroy
   has_many :notifications_as_mentioned, class_name: 'ActivityNotification', dependent: :destroy
-  has_many :messages, through: :requests
+  has_many :messages, dependent: :destroy
```
roschaefer, line 17: Why didn't we do this earlier?


Is the connection between request<->organization now given through messages? :thinking:

config/locales/de.yml
```diff
@@ -144,6 +144,9 @@ de:
     move_message_form:
       submit: Verschieben
       cancel: Abbrechen
+      no_request:
+        label: Keine Frage
```
roschaefer, line 148:
```suggestion
        label: Gehört zu keiner Frage
```

config/locales/de.yml
```diff
@@ -144,6 +144,9 @@ de:
...
+        help: Wenn sie nicht zu einer Frage gehört, wähle diese.
```
roschaefer, line 149:
```suggestion
        help: Wähle diese Option, wenn die Nachricht zu keiner Frage gehört.
```

spec/requests/messages_request_spec.rb
```diff
@@ -109,26 +109,22 @@
     end
   end
 
-  describe 'POST /:organization_id/messages/:id/request' do
+  describe 'PATCH /:organization_id/messages/:id/request' do
```
roschaefer, line 112: :clap:

roschaefer (approved):
Wow, quite a large PR. Well done. Can't see any issues from looking at the code. Yet, it's quite dangerous so my suggestion is to remove the request from the `:message` factory and check which test cases fail. If there's nothing we forgot, we can fix them with a trait `create(:message, :with_request)`.

impressive

app/components/requests_feed/requests_feed.rb
mattwr18: I updated it to `most_recent_reply_to_each_request` to accurately describe what it returns.
I also moved the method to the request feed class since it is the only place it is used.

app/models/organization.rb
mattwr18: I think it wasn't there before because the dependent option is set on the request since it went through the request. It would be redundant to have it in both places.

Quite possibly, but I don't know if this PR is the place to drill all these down. We still could benefit from doing this with the contributor.organization after we decided that a contributor can only ever belong to one organization.

config/locales/de.yml
```diff
@@ -477,6 +480,8 @@ de:
         new_request_night_1: "Hallo %{first_name}, heute Nacht ist mir etwas zum Thema „%{request_title}“ eingefallen. Wenn du antworten möchtest, klicke auf 'Antworten'."
         new_request_night_2: "Hallo %{first_name}, ich weiß, es ist schon spät. Aber ich möchte gern mehr zum Thema „%{request_title}“ erfahren. Wenn du antworten möchtest, klicke auf 'Antworten'."
         new_request_night_3: "Hallo %{first_name}, zum Thema „%{request_title}“ möchte ich heute noch gern mehr erfahren. Wenn du antworten möchtest, klicke auf 'Antworten'."
+      direct_message_template:
+        new_direct_message: "Hallo %{first_name}, wir haben eine neue Nachricht für dich. Wenn du antworten möchtest, klicke auf 'Antworten'."
```
mattwr18, line 484: According to @FrauCsu 
Hallo %{first_name}, wir haben eine neue Nachricht für dich. Wenn du sie öffnen möchtest, klicke 'Öffnen'.

PR #2116, opened 2025-02-11
Add basic functionality for quote reply to WhatsApp

### LINKED ISSUE #2115: [feature] Add quote reply to 360dialog WhatsApp
Related to https://github.com/tactilenews/100eyes/issues/1478

mattwr18:
Closes #2115

### DIFF
--- a/app/adapters/whats_app_adapter/three_sixty_dialog_outbound/text.rb
+++ b/app/adapters/whats_app_adapter/three_sixty_dialog_outbound/text.rb
@@ -49,12 +49,14 @@ module WhatsAppAdapter
       end
 
       def text_payload
-        base_payload.merge({
-                             type: 'text',
-                             text: {
-                               body: text || message.text
-                             }
-                           })
+        payload = base_payload.merge({
+                                       type: 'text',
+                                       text: {
+                                         body: text || message.text
+                                       }
+                                     })
+        payload.merge!({ context: { message_id: message.reply_to_external_id } }) if message&.reply_to_external_id.present?
+        payload
       end
 
       def welcome_message_payload
--- a/app/components/chat_form/chat_form.html.erb
+++ b/app/components/chat_form/chat_form.html.erb
@@ -1,8 +1,7 @@
 <%= form_with(scope: :contributor, url: message_organization_contributor_path(organization, contributor), class: 'ChatForm', id: 'chat-form', local: true) do %>
   <% if reply_to.present? %>
-    <h3><%= "Zur Frage: #{reply_to.request.title}" %></h3>
     <%= hidden_field_tag 'message[reply_to_id]', reply_to&.id %>
   <% end %>
-  <%= c 'textarea', id: 'message[text]', value: prefilled_value, required: true, placeholder: t('components.chat_form.placeholder') %>
+  <%= c 'textarea', id: 'message[text]', required: true, placeholder: t('components.chat_form.placeholder') %>
   <%= c 'button', label: t('components.chat_form.submit'), type: :submit %>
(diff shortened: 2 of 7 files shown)


soey:
It looks all fine on the implementation side when considering only the change code. But a few points to consider:
- Shouldn't there be some tests covering the changed behavior of the adapter?
- I'm a bit concerned with possible inconsistency on the model level of Message and would suggest to think about it. We have the reply? method there deriving the boolean from the sender being a contributor or not. Now with the new field there are two values that are interdependent. Currently I think there is no validation that the the reply_to_external_id can only be set if the sender is a contributor. So there can be an inconsistent data set. I'm always concerned when things like this can happen..
- Can we receive inline answers? What happens then?

mattwr18:
> It looks all fine on the implementation side when considering only the change code. But a few points to consider:
> 
>     * Shouldn't there be some tests covering the changed behavior of the adapter?
> 
>     * I'm a bit concerned with possible inconsistency on the model level of Message and would suggest to think about it. We have the reply? method there deriving the boolean from the sender being a contributor or not. Now with the new field there are two values that are interdependent. Currently I think there is no validation that the the reply_to_external_id can only be set if the sender is a contributor. So there can be an inconsistent data set. I'm always concerned when things like this can happen..
> 
>     * Can we receive inline answers? What happens then?

Thanks for getting back to me @soey 

* Yes, I will add tests, I just wanted to get your quick feedback on the implementation cause I wasn't 100% on it. I love its simplicity, but I also had some questions about the need to save this `reply_to_external_id` to the db. I thought about passing it in to the job, but this would require updating the api for all channels, this might be desirable :shrug: 

It is useful to save it for the case where the user replies to a contributor after the 24 hour period for freeform messages has closed. In this case, it will send a `Message::WhatsAppTemplate` message and if they respond, it will still quote reply the message since it has its reference.

* Actually, I think the new `reply_to_external_id` would be useful, potentially, for incoming and outgoing messages so there would be no inconsistency. At the moment, when someone quote replies or answers a template message, we receive the same `context` object with the external id of the template they are responding to. See https://github.com/tactilenews/100eyes/blob/main/app/adapters/whats_app_adapter/three_sixty_dialog_inbound.rb#L19. This is how we can send them a request that is not the latest they have received. Inexplicably, meta does not provide the external message id if the contributor simply quote replies a freeform message. I assume this is not set in stone though and could potentially change. I have not looked into what we receive in the payload for the other messengers, but if we do receive any indication that the contributor is quote replying, my intention would be to save it in the same attribute and potentially, we could use it to display it in the UI in 100eyes.

We do already have some inconsistencies along these lines that you might have noticed. `recipient_id`, `broadcasted`, and `blocked` are only ever valid for outbound messages. `unknown_content` and `telegram_media_group_id` are only ever valid for inbound messages. `highlighted` could be for both, but I guess in practice there might be little reason for them to highlight their own direct messages. `creator_id` is used only for `reply?` though they are not technical inbound, they are a way for the user to create a message manually that they might have received from the contributor in some way that circumvents our app.

I agree that we should have the conversation about when the two models divert sufficiently to warrant separating them and maybe have some shared attributes, but extending them to keep the api cleaner; however, I don't think this attribute actually does belong only to outbound messages.

* What do you mean by inline answers? Do you mean quote replies from the contributor? I kind of already answered this. Unfortunately, we do not receive any indication if the contributor quote replies a freeform message. I also confirmed this with the 360 dialog AI. We do receive this information for template messages, we could potentially add this attribute to the `message_whats_app_template` records and be able to generate some kinds of statistics on how often someone responds to our templates. For other messengers, I will have to look into it.

